PR #5, opened 2015-02-11
Add support for browserify with polyfill generated from github/fetch.

tracker1:
This will allow this module to work with browserify, including the [fetch polyfill](https://github.com/github/fetch).

* bumped version to `1.1.0`
* renamed `index.js` to `node-fetch.js`
* added `whatwg-fetch` (github/fetch) to package.json
* added `install.js` script to generate `browser-fetch.js` on install
* added `browser-fetch.js` to `.gitignore` (generated at install)
* created new `index.js` with browser/node logic for browserify

Removes any need for [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch)


tracker1:
Not sure about the travis errors for node 0.10 and iojs 1.0... they seem really odd in this case.

bitinn:
Actually I would prefer ppl with isomorphic needs go use isomorphic-fetch instead, node-fetch and github/fetch are using different tech underneath. And [I started this project](https://github.com/bitinn/node-fetch#motivation) with a server-side only idea in mind.

Let @matthew-andrews knows if there is anything isomorphic-fetch missed?

matthew-andrews:
Yep @tracker1 isomorphic-fetch should be exactly what you need. It pulls in both this repo and the browser one.  Please let me know if something about it doesn't work quite right for you.

PR #11, opened 2015-04-16
Allow body option to be a buffer

dmikis:
Subj.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ default values are shown, note that only `method`, `headers` and `body` are allo
 	, timeout: 0      // request timeout in ms, 0 to disable, timeout reset on redirect
 	, compress: true  // support gzip/deflate content encoding, false to disable
 	, size: 0         // maximum response body size in bytes, 0 to disable
-	, body: empty     // request body, can be a string or readable stream
+	, body: empty     // request body, can be a string, a buffer  or a readable stream
 	, agent: null     // custom http.Agent instance
 }
 ```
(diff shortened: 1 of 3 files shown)


bitinn:
@dmikis thx for the pull request, but I don't see why this can't be handled in application code?

could you provide a use-case?

dmikis:
@bitinn, what about binary data?

bitinn:
@dmikis you should always consider stream first.

dmikis:
To be honest, after your question I've realised I don't really need that in my project.

But, never the less, I believe that forbidding `Buffer` as a body you introduce an unnecessary limitation to the API. Especially taking into account the fact that that node's own `http.request` allow that. Necessity to send binary data doesn't actually mean that you have a file (or even data from other http request). You may as well generate the data in your application.

bitinn:
Sure, though I don't think we ever use Buffer instead of Stream.

My idea is to expose minimal API unless there is really a good reason for it (much like why we only support `res.text()` and `res.json()` at the moment)

dmikis:
OK, let's close this request for now.

mbrevda:
As for a use case: to be able to send data directly to a method that expects a Buffer:

```js
var p = fetch('http://some/' + image.location)
    .then(function(res){
        var hash = crypto.createHash('sha1').update(res.body).digest('hex')
        fs.createWriteStream(hash).write(res.body)
    })
```

Or is there some way to do this currently?

bitinn:
@mbrevda this PR deals with `req.body`, not `res.body`. see #51 for further discussion, but on node.js you should always be able to `pipe` a stream.

ref: http://stackoverflow.com/questions/18658612/obtaining-the-hash-of-a-file-using-the-stream-capabilities-of-crypto-module-ie

PR #12, opened 2015-04-16
Clear request timeout after response

dmikis:
Otherwise node.js waits it. In other words, if you make a request with, say, 10 seconds timeout, your script will run for about 10 seconds doing nothing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,10 +102,11 @@ function Fetch(url, opts) {
 
 		// send request
 		var req = request(options);
+		var reqTimeout;
 
 		if (options.timeout) {
 			req.once('socket', function(socket) {
-				setTimeout(function() {
+				reqTimeout = setTimeout(function() {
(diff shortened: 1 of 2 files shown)


bitinn:
Please show me an example of waiting `node-fetch` request, what you are saying could be by design.

dmikis:
```
84:~/.../Coding/node-fetch[master]$ time node -e "require('./')('http://ya.ru', {timeout: 1e4}).then(function () { console.log('ok'); })"
ok

real	0m10.100s
user	0m0.096s
sys	0m0.019s
84:~/.../Coding/node-fetch[master]$
```

bitinn:
This makes sense now, I am ok to merge it, care to add a test to verify it's behaving as expected? Might be a tricky one though.

bitinn:
I will do some double checking before publishing a new version.

PR #20, opened 2015-05-03
Throw instead of ignore, if headers are invalid

emmenko:
After spending on hour trying to debug the problem, I found out that I was passing a number as header value, thus the header was ignored. 

I think it would be much more useful throwing an error if something is not valid, instead of silently ignore it.

Thanks

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -32,11 +32,9 @@ function Headers(headers) {
 		if (typeof headers[prop] === 'string') {
 			this.set(prop, headers[prop]);
 		} else if (headers[prop] instanceof Array) {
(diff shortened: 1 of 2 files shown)


bitinn:
I would like to see how browsers and github polyfill handles it before making changes.

I would like to avoid throwing errors if possible, especially on value like `null` and `undefined`

bitinn:
I have pushed a version 1.2.1 to normalize `Number` type instead (only for `new Headers` at the moment, you need to take care using  `.set` or `.append`)

emmenko:
@bitinn perfect, that would do it as well! Thanks :+1:

PR #22, opened 2015-05-05
Use iconv directly instead of encoding library

emmenko:
**tl;dr** Drops `encoding` and favours `iconv-lite` to be able to bundle it with (e.g.: `webpack`, `browserify`) without problems.

Currently, if I bundle `node-fetch` with `webpack` I get a warning that `encoding` is using dynamic dependencies. Unfortunately the guy from the `encoding` library explained that there is no way around it.
He then suggested though to actually use directly `iconv`. Tests are still working of course.

You can follow the conversations about this problem here:
https://github.com/andris9/encoding/pull/13
https://github.com/webpack/webpack/issues/918#issuecomment-98640907

Hope it's ok for you :)

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -6,7 +6,7 @@
  */
 
 var http = require('http');
-var convert = require('encoding').convert;
+var iconvLite = require('iconv-lite');
 
 module.exports = Response;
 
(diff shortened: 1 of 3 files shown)


bitinn:
I think what you want is: https://github.com/matthew-andrews/isomorphic-fetch

emmenko:
@bitinn didn't know of that library. Yes, it looks like what I need, thanks! :)

Issue #8, opened 2015-03-20
tasks for considering (2.0 release)


bitinn:
just collecting some common complain about 1.x release
- [x] provide some guideline on how to handle rejected promise (failed fetch), this can be caused by many reasons: url, headers.location, network timeout, request config etc; currently the only way to distinguish them is to check error message.
- [x] provide some alternatives to extract redirect headers, it's not possible with current fetch spec, only the final url will be expose to user when following redirects.
- [ ] some way for cookies to persist when you do a POST request that redirects (which should be followed by a GET request, often with cookie as credential). Though if we implement extraction for redirect headers we can make this possible (keep the cookie headers for your next GET request).
- [x] expose Request/Headers/Response interfaces, we don't currently have a Request interface due to the fact that no one are using them.
- [x] normalize Headers `.set` and `.append` to prevent non-string input, it's recommended to already cast your input to `String` type first.
- [x] Headers/Response interface should support ES6 features, such as `keys()` and `entries()`

tracker1:
* handling rejected promises, does the public specification provide any guidelines on this... ideally this module should match the error signatures used by the public version.  Perhaps a complimentary module that can interrogate the error to give a more useful sub-type for the given error message?

* Redirect headers, perhaps a `history` on the response, containing any response objects through a history chain the first entry being the first response then the redirect chain, until the final result.
  * each entry in history is a response object
  * entries in history do not have a `history` property, only the final response
  * the final response is not included in `history`
  * history will be `undefined` unless there are values (will never be a 0 array)
    * this will allow a simpler `if (res.history)` kind of check instead of `if (res.history && res.history.length)` ... not really required, but an idea

Note: neither of the above suggestions would require a 2.0 bump as they would be additive... for the history, once implemented, would put forward an issue to chromium/safari/ie/firefox to add the feature and to whatwg to extend the standard.

bitinn:
@tracker1 Thx for the suggestions.

Just to clarity, item 2 isn't possible with Fetch spec because of security issue, so I don't expect spec to change. see https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

For item 1 there isn't a standard, if we are to tackle it, we will have to come up with ours. see https://github.com/whatwg/fetch/issues/25

Good news is there are many alternatives to workaround these Fetch limits, so I am not jumping to solve them now.

bitinn:
[...]

PR #23, opened 2015-05-11
Implement request redirect modes

rricard:
Hi !

I implemented one missing piece of the spec that I need to see working about redirects.

Related spec part: https://fetch.spec.whatwg.org/\#concept-request-redirect-mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,13 +139,14 @@ default values are shown, note that only `method`, `headers` and `body` are allo
 ```
 {
 	method: 'GET'
-	, headers: {}     // request header, format {a:1} or {b:[1,2,3]}
-	, follow: 20      // maximum redirect count, 0 to not follow redirect
-	, timeout: 0      // req/res timeout in ms, 0 to disable, timeout reset on redirect
-	, compress: true  // support gzip/deflate content encoding, false to disable
-	, size: 0         // maximum response body size in bytes, 0 to disable
-	, body: empty     // request body, can be a string or readable stream
-	, agent: null     // custom http.Agent instance
+	, headers: {}        // request header, format {a:1} or {b:[1,2,3]}
+	, follow: 20         // maximum redirect count, 0 to not follow redirect
+	, redirect: 'follow' // request redirect mode: follow, manual or error
+	, timeout: 0         // req/res timeout in ms, 0 to disable, timeout reset on redirect
+	, compress: true     // support gzip/deflate content encoding, false to disable
+	, size: 0            // maximum response body size in bytes, 0 to disable
+	, body: empty        // request body, can be a string or readable stream
+	, agent: null        // custom http.Agent instance
 }
 ```
 
--- a/index.js
+++ b/index.js
@@ -71,6 +71,7 @@ function Fetch(url, opts) {
 			, method: opts.method || 'GET'
 			, headers: opts.headers || {}
 			, follow: opts.follow !== undefined ? opts.follow : 20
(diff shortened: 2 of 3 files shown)


rricard:
Added tests, waiting for the CI...

bitinn:
Thx for the PR, care to show me why this is important for your case?

I don't really see a good use case for this when we can already set `follow: 0` to disable following redirect?

rricard:
I need to use it in tests in order to make assertions on the `location` or `set-cookie` headers without following it. If I use `follow: 0` I'll trigger an error without being able to access the response object (in which there are things I want to assert on). Moreover, that's in the spec. If you need me to refactor stuff or anything, I would be happy to do it!

rricard:
Just read #8 and saw the security issues related to this https://fetch.spec.whatwg.org/#atomic-http-redirect-handling. That's true that in the spec https://fetch.spec.whatwg.org/#requestredirect, `manual` cannot be set but we're on node and that's something we could bypass ... And that's really important for those who were using something like [supertest](https://www.npmjs.com/package/supertest) before for testing API endpoints.

bitinn:
Hmm, that's quite an edge case for using Fetch API, maybe if we have a better Error standard, we can expose headers/body of the failed response.

But for now, I would consider promisify a low-level API like `http.request` to achieve this particular need, because what you want is probably a request that don't follow by itself and give you full response? like doing a `curl -I http://some.redirect.url/` ?

By adding this optional flag, we are essentially saying: *don't reject when max redirect limit is reached, just get back the final response*. While I am ok not implementing some aspect of Fetch Spec security, accept a response when it should have rejected them, is really what hold me from moving in this direction.

rricard:
The manual mode just stops there even if there's a redirection. It's not an error case, it's what we expect. A 302 is not an error, it's just a valid HTTP response and we should be able to inspect this response, at least in a server environment. In a client environment, sure, there are some security issues (you don't want to see a malicious script inspect a set-cookie header). In my opinion, fetch is a low level api (with promises), it should be able to handle this kind of thing (make an http request and stop there, don't follow anything).

I already promisified things like supertest. It works but it's not the fetch API: fetch is way nicer to use and way more flexible ...

bitinn:
In both Fetch and XMLHttpRequest's view, redirect is explicitly transparent, even for `HEAD` request, so either you got to a page that isn't redirecting (Fulfilled, even for 4xx and 5xx), or you have reached the redirect limit or timeout or network error (which are all Rejected).

Allowing Fetch to stop at redirection would be an extension to that mental model (3xx can be Fulfilled or Rejected based your input). And we will have to answer questions like:

- Can I extract headers and body for each redirection if we follow more than once?
- Can we change the request headers and body upon redirection?

To me both answers would be No, and that's why I don't believe Fetch API is low-level in our context: because a truly low-level tool don't care about Promise or Status Code or JSON Parsing.

If you prefer a higher-level tool then `http(s).request`, maybe module like https://github.com/leukhin/co-request can help.

bitinn:
After giving this more thought...

I would prefer an approach where rejected Error object contains optional `err.headers` and `err.body` of last response received (for cases like: `req.follow` is 0, but server response is a redirection). So user may do a `.catch` to inspect headers if so desired.

Having a standardized Error might also help debugging in general, we will collect feedbacks on #8 

Closing this PR, but thx for your input.

rricard:
Ok, that's a perfectly reasonable approach. I'll work on it then. Thanks for your insightful feedback :)

PR #25, opened 2015-05-18
Factoring out Request class

dbo:
A first step of factoring out the Request class. Options etc. stay the same as they are.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,194 +1,39 @@
-
 /**
  * index.js
  *
  * a request API compatible with window.fetch
  */
 
-var parse_url = require('url').parse;
-var resolve_url = require('url').resolve;
-var http = require('http');
-var https = require('https');
-var zlib = require('zlib');
-var stream = require('stream');
-
+var Request = require('./lib/request');
 var Response = require('./lib/response');
 var Headers = require('./lib/headers');
 
-module.exports = Fetch;
+module.exports = fetch;
 
 /**
- * Fetch class
+ * fetch function
  *
- * @param   String   url   Absolute url
- * @param   Object   opts  Fetch options
+ * @param   String|Request   input
+ * @param   Object   		 init
  * @return  Promise
  */
-function Fetch(url, opts) {
-
-	// allow call as function
-	if (!(this instanceof Fetch))
-		return new Fetch(url, opts);
-
+function fetch(input, init) {
 	// allow custom promise
-	if (!Fetch.Promise) {
-		throw new Error('native promise missing, set Fetch.Promise to your favorite alternative');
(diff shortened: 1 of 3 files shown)


bitinn:
I agree with the general direction of refactoring Request class, but a couple of questions:

1. Do you happen to use the `Request` class? I know the spec has it, but we recommend using object for Request and Headers whenever possible (because they are easier).

2. To me the better approach is only refactor the options parts into Request, but leave the actual request in index.js, see github polyfill's approach: https://github.com/github/fetch/blob/master/fetch.js#L206-L220, save us from doing `this.execute` at least.

dbo:
First of all thanks for answering on PRs.

1. Yes, I use the Request ctor, to allow it as a drop-in replacement for isomorphic code.
2. (Assuming I got your point correctly) you look for structural equality of both (Request and RequestInit options). I think this has the drawback that you need to check defaults in two places. I also don't think the ctor call matters performance-wise.

bitinn:
1. Plain object is supported by both native fetch and polyfill, and plain object is isomorphic by default, so why use `Request` or `Headers` interface directly? It adds more lines to your codebase and provide little value on abstraction, plus it defeats the point of using a simplified `fetch` API. I am not here to judge, but it always seem weird to me to use `new Request/Headers` directly.

2. I prefer smaller and incremental change in this case. If I were to do it, I probably will:

- create an empty Request interface first
- try to move these attributes into it https://github.com/bitinn/node-fetch/blob/master/index.js#L66-L80
- add checks for `instanceof Request` in index.js, so that we can support passing in a Request object
- expose it to fetch.Request

So in my case, `new Request` will only contain a few attributes (as close to a plain object as possible), and none of the `http.request` or `promise`. Those code remain in the index.js

basically, I will start with something like https://github.com/bitinn/node-fetch/blob/master/lib/response.js#L20-L32, and add `.clone` when there is any actual use-cases.

dbo:
I need to support the official fetch API, and I don't have influence on all the client code, so even though you're right about using a plain object, it doesn't help me.
Please also mind the double checking of attributes + defaulting in your suggestion: You will need to check these in the ctor (constructing Request out of url + RequestInit) as well when just reading the plain object.

bitinn:
Noted, I would like to work on this when available, probably end of May or early June.

I will keep this PR open as a reminder, my suggestion for now is only expose `fetch` to client code (not always possible, I know, but when I started this project the whole point is to provide a `fetch` API, we only recently exposed `Headers` and `Response` interface due to popular request for mocking these interfaces)

bitinn:
close via #26, released as v1.3.0

PR #28, opened 2015-07-11
Allow a custom host to be set

matthew-andrews:
Some of the requests we make from the server need to overwrite the host header (e.g. for ‘apis’ which were originally written with a CDN in mind and we're just making requests that act like a CDN) and node-fetch currently chokes if you try to set a custom host header, check out e98f22b3ba8350a37c6f3c8917df56655e5c4fc7, run `npm test` and you'll get this output:-

```
  1) node-fetch should allow setting of custom host:
     TypeError: undefined is not a function
      at Agent.createSocket (_http_agent.js:186:39)
      at Agent.addRequest (_http_agent.js:166:23)
      at new ClientRequest (_http_client.js:154:16)
      at exports.request (http.js:49:10)
      at /Users/mandrews/sandboxes/node-fetch/index.js:83:13
      at new Fetch (/Users/mandrews/sandboxes/node-fetch/index.js:44:9)
      at Fetch (/Users/mandrews/sandboxes/node-fetch/index.js:32:10)
      at Context.<anonymous> (/Users/mandrews/sandboxes/node-fetch/test/test.js:763:10)
      at callFn (/Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runnable.js:251:21)
      at Test.Runnable.run (/Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runnable.js:244:7)
      at Runner.runTest (/Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runner.js:374:10)
      at /Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runner.js:452:12
      at next (/Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runner.js:299:14)
      at /Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runner.js:309:7
      at next (/Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runner.js:248:23)
      at Immediate._onImmediate (/Users/mandrews/sandboxes/node-fetch/node_modules/mocha/lib/runner.js:276:5)
      at processImmediate [as _immediateCallback] (timers.js:358:17)
```

I'm not entirely happy with my hack but it seemed the least intrusive way to make this work :disappointed:.  Very happy to tweak the implementation if you can think of a better way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -79,6 +79,11 @@ function Fetch(url, opts) {
 
 		options.headers = headers.raw();
 
+		// HACK: headers.host must be a string, cannot be an array otherwise get error ‘undefined is not a function’
+		if (options.headers.host) {
+			options.headers.host = options.headers.host[0];
+		}
+
 		// send request
 		var req = send(options);
 		var reqTimeout;
--- a/test/server.js
+++ b/test/server.js
@@ -244,4 +244,10 @@ TestServer.prototype.router = function(req, res) {
 		});
 	}
 
+	if (p === '/host') {
+		res.statusCode = 200;
+		res.setHeader('Fetch-Sent-Host', req.headers.host);
+		res.end();
+	}
+
 }
--- a/test/test.js
+++ b/test/test.js
@@ -751,4 +751,18 @@ describe('node-fetch', function() {
 			expect(res.ok).to.be.true;
 		});
 	});
+
+	it('should allow setting of custom host', function() {
+		url = base + '/host';
+		opts = {
+			method: 'HEAD',
+			headers: {
+				host: 'bitinn.net'
+			}
+		};
+		return fetch(url, opts).then(function(res) {
+			expect(res.headers.get('fetch-sent-host')).to.equal('bitinn.net');
+		});
+	});
+
 });


matthew-andrews:
I don't think the travis failure is a real failure :confused:

bitinn:
I am trying to understand this issue:

1. So node.js http/https module can't handle a host header that's an array (fair enough), which our Headers interface are using by default when you set a header.
2. And you need to send test request to domain A while using a host header of domain B.
3. And this is a server-side only feature, because fetch on client-side shouldn't allow users to set custom host header (it's a forbidden header).

All True?

An alternative solution would be to use String instead of Array when storing headers internally (unless the input is an Array), so headers.raw() return host header as a string by default.

Any other case that might warrant such a change? If not, perhaps this hack has a lower risk of breaking things...

matthew-andrews:
Yes all true…

1. Yes, it appears so — at least on the versions of Node I'm using (`v0.12.x`).  Actually I couldn't find any docs suggesting that **array** was supported (although it seems to be in most cases).
2. Yes.
3. Yes that's correct, but this is equivalent to `node-fetch`'s ability to allow a custom user agent to be set so I don't think it would be wrong for `node-fetch` to support this.

> An alternative solution would be to use String instead of Array when storing headers internally (unless the input is an Array), so headers.raw() return host header as a string by default.
> Any other case that might warrant such a change? If not, perhaps this hack has a lower risk of breaking things...

I couldn't think of a use case where you might want to set multiple values for the same header, but as you say, since it's supported already fixing it for the host case only lowers the risk.  I'm really not that familiar with http……

A quick glance at the [spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) suggests the following:-

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

I wonder if that's what node's http(s) request module does internally:-

```
headers: {
  'my-header': [ 'a', 'b' ]
}
```
=>

```
curl … -H 'my-header: a, b'
```

matthew-andrews:
^^ That does seem to be the case…

webserver.js
```js
require('http').createServer(function(request, response){
        console.log("my-header: " + request.headers['my-header']);
        console.log("my-other-header: " + request.headers['my-other-header']);
}).listen(8080);
```

test.js
```js
var req = require('http').request({
        host: 'localhost',
        port: 8080,
        method: 'POST',
        path: '/',
        headers: {
                'my-header': ['a','b'],
                'my-other-header': ['a, b']
        }
});
req.write('');
req.end();
```

Output of webserver:-
```
my-header: a, b
my-other-header: a, b
```

So I think it would be safe If I changed my implementation to something like:-

```js
options.headers = Object.keys(options.headers).map(function(header) {
    return options.headers[header].join(', ');
  });
```

Does that look ok to you?

matthew-andrews:
ok that code doesn't work hold on……

matthew-andrews:
```js
                options.headers = headers.raw();

                options.headers = Object.keys(options.headers).reduce(function(previousHeaders, header) {
                                previousHeaders[header] = options.headers[header].join(', ');
                                return previousHeaders;
                        }, {});
```

bitinn:
Not sure how to feel about your last snippet, to me your original PR is better, as it's easier to figure out why we make such change :)

I have searched around on github and found a few users already using the [internal `_headers` store and expect Array type on each header](https://github.com/avoidwork/haro/blob/812824eba0371e2ec732c875c685138750052711/lib/haro.js#L582-L589), so I think your hack is the best way forward.

matthew-andrews:
okay let's go with that :)

matthew-andrews:
Although my last snippet doesn't change anything about the internal _headers store …… I'm proposing putting that into `index.js` just before the `options` object gets passed into `send`.

bitinn:
@matthew-andrews got it, i was refering to my alternative solution to change Headers. Also looping around the all headers just to workaround host header seem like a bad deal.

matthew-andrews:
ahh ok :)

> looping around the all headers just to workaround host header seem like a bad deal

:+1: to this

Thanks for your help!

matthew-andrews:
Definitely won't want to change the headers interface — I think it currently matches the spec perfectly:-
https://developer.mozilla.org/en-US/docs/Web/API/Headers/get
https://developer.mozilla.org/en-US/docs/Web/API/Headers/getAll

bitinn:
v1.3.1 release include this workaround.

matthew-andrews:
thanks!

PR #30, opened 2015-07-13
Expose ‘keepAlive’ option

matthew-andrews:
Hi @bitinn, I wonder if you'd consider exposing the `keepAlive` option to `http.request`/`https.request`?

https://nodejs.org/api/http.html#http_http_request_options_callback

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ default values are shown, note that only `method`, `headers` and `body` are allo
 	, size: 0         // maximum response body size in bytes, 0 to disable
 	, body: empty     // request body, can be a string or readable stream
 	, agent: null     // custom http.Agent instance
+	, keepAlive: null // keep sockets around in a pool to be used by other requests in the future
 }
 ```
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -57,6 +57,7 @@ function Request(input, init) {
 		input.compress : true;
 	this.size = init.size || input.size || 0;
 	this.agent = init.agent || input.agent;
+	this.keepAlive = init.keepAlive || input.keepAlive;
 
 	// server request options
 	this.protocol = url_parsed.protocol;


bitinn:
I actually prefer using a custom agent to achieve the same thing in this scenario, I don't think it's necessary to expose this options directly. What do you think, and do you actually see a huge improve in performance?

https://nodejs.org/api/http.html#http_new_agent_options

matthew-andrews:
Hmm, fair point.  I guess it's for the convenience of setting a flag rather than a few extra lines.

I'll run some tests and get back to you……

milahu:
here is a keepAlive wrapper for node-fetch, based on [this](https://stackoverflow.com/questions/62500011/reuse-tcp-connection-with-node-fetch-in-node-js) stackoverflow post

```js
// fetch: add option keepAlive with default true
const fetch = (function getFetchWithKeepAlive() {
  const node_fetch = require('node-fetch');
  const http = require('http');
  return async function (url, userOptions) {
    const options = { keepAlive: true };
    Object.assign(options, userOptions);
    if (options.keepAlive == true)
      options.agent = (parsedUrl => parsedUrl.protocol == 'http:' ? httpAgent : httpsAgent);
    delete options.keepAlive;
    return await node_fetch(url, options);
  }
})();

const response = await fetch('https://github.com/');
const response = await fetch('https://github.com/', { keepAlive: false });
```

jimmywarting:
I would probably have gone with [Ilan Frumer](https://stackoverflow.com/a/62500224/1008999) solution instead than having to reconstruct agents for each request

milahu:
> I would probably have gone with [Ilan Frumer](https://stackoverflow.com/a/62500224/1008999) solution

same thing : ) the two agents are initialized in the `const fetch = ...` IIFE and stay constant
ive added the `fetch` function name to make it more clear

PR #31, opened 2015-07-22
Automatically pull headers from node-form-data.

DylanPiercey:
This is convenient for isomorphic apps that rely on form-data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ fetch('http://httpbin.org/post', { method: 'POST', body: stream })
 var FormData = require('form-data');
 var form = new FormData();
 form.append('a', 1);
-fetch('http://httpbin.org/post', { method: 'POST', body: form, headers: form.getHeaders() })
+fetch('http://httpbin.org/post', { method: 'POST', body: form })
 	.then(function(res) {
 		return res.json();
 	}).then(function(json) {
--- a/index.js
+++ b/index.js
@@ -77,6 +77,10 @@ function Fetch(url, opts) {
 			headers.set('accept', '*/*');
 		}
 
+		if (!headers.has('content-type') && options.body && typeof options.body.getBoundary === "function") {
+			headers.set('content-type', 'multipart/form-data; boundary=' + options.body.getBoundary());
+		}
+
 		options.headers = headers.raw();
 
 		// http.request only support string as host header, this hack make custom host header possible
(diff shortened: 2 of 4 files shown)


bitinn:
Hi, thx for the PR, but I don't think `node-fetch` should have knowledge about `body.getBoundary`.

So really we are making a special case for `node-form-data`, while my example is just to demonstrate that it works with `node-form-data`.

Any reason you have to do this instead of passing headers separately?

DylanPiercey:
I understand that it is a bit specific to `node-form-data` however I think it would be a nice add on as both libraries are attempting to achieve similar goals (isomorphic ajax requests).

Currently I have to write code something like this to make it work isomorphically.

```javascript
form = new FormData();

form.append("file", file);

// Check if I am in the browser :(
if (!process.browser) {
  headers = form.getHeaders();
  for (field in headers) {
    options.headers[field] = headers[field];
  }
}

// Use node-fetch with options
```

It would be awesome if this small exception could be added (similar to supporting streams) to allow for a nice serverside (form-data) experience.

bitinn:
Alright I get your use-case here, here is my understanding:

- It looks like you would like `FormData` support on `node-fetch`, which is fair enough.
- The main problem: unlike streams, no spec for FormData on node.js
- The main difference between client-side FormData api and `node-form-data`: on client-side Fetch API generate the body payload and boundary, on node.js, `node-form-data` handles it.
- To keep `node-fetch` lightweight and free from dependency, we were only supporting the [alternative method](https://github.com/felixge/node-form-data#alternative-submission-methods) mentioned on `node-form-data`.
- So unfortunately this use case is not as isomorphic as it seems.

I will take the PR, as I see no better solution to detect FormData reliably on node.js

bitinn:
PS: to clarify my personal stand on isomorphic javascript, I think it can be misleading to developers who thinks the code just works both client-side and server-side, when the truth is far from it (what we are doing is a good example of such problem).

bitinn:
released as v1.3.2, thx for your work.

DylanPiercey:
Thanks a lot for your work to get this merged and released so quickly. I understand your viewpoint (that devs should realize that the code on the server is often different from the client). Personally I think that this is explained well in both node-fetch and node-form-data and that users of both understand that they work with native node modules under the hood and expose some api differences  (Streams and such).

I am glad to see this merged as both libraries together act as a great (complete), abstraction for isomorphic fetch.

Issue #33, opened 2015-07-31
Response must parse headers from options


kirill-konshin:
According to spec if `options.headers` is not a `Headers` object, then it must be converted.

bitinn:
Not quite sure what you mean by that, we do run `new Headers` when you call `fetch` with an object as options. But I won't expect the input itself be turned into `Header` instance.

Please provide us an example on where you run into this problem?

kirill-konshin:
Test case:

```js
(new Response()).headers instanceof Headers // true
```

kirill-konshin:
Same applies to `Request` class as well. In current implementation it is a simple assignment:

```js
this.headers = init.headers || input.headers || {};
```

How it should be done:

```
this.headers = new Headers(options.headers)
```

kirill-konshin:
Also `Response` may receive it's body as a string, in this case it should be either converted to a stream or stored as string, but in this case `.__decode()` method should be aware of that.

bitinn:
You are right. We didn't follow the spec in these aspects as we don't see any good use-case for it.

kirill-konshin:
Unit tests -- to pass around certain responses and requests since the rest of the code is fetch-based.

bitinn:
You might need to provide a real-world example for it.

The reason we expose Header/Response/Request classes are not because we want this module to be isomorphic as `window.fetch`. But rather there are requests to have them exposed for testing purpose.

I hope there is no confusion here.

kirill-konshin:
It's not about isomorphism, it's about not following the spec. Since you already expose `Header/Response/Request` classes they should be compatible with the latest spec.

It's good that they are exposed for testing purposes because this is the main use case: to have an ability to build a custom `Response` and pass it as a mock w/o calling `fetch`.

bitinn:
Could you show me some code? What were you trying to do that's affected by this issue? I am a bit busy early august so this might not be on my priority list.

kirill-konshin:
Very short example:

```js
function multipartMixedParserToBeTested(response){
    if (response.headers.get('content-type').indexOf('multipart/mixed') !== -1) {
        // ...
    }
}

var responseMock = new Response(body, {{headers:{'content-type': 'multipart/mixed'}});

multipartMixedParserToBeTested(responseMock).then(...);
```

bitinn:
Fair enough, I will look into fixing it next week (parse headers for Response and Request when input is not already a Headers)

kirill-konshin:
[...]

PR #37, opened 2015-08-07
Fix for #33

kirill-konshin:
Added parsing of Headers in Response & Request constructors
Added support of string bodies in Response

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -6,6 +6,7 @@
  */
 
 var parse_url = require('url').parse;
+var Headers = require('./headers');
 
 module.exports = Request;
 
@@ -42,7 +43,7 @@ function Request(input, init) {
 
 	// fetch spec options
 	this.method = init.method || input.method || 'GET';
-	this.headers = init.headers || input.headers || {};
+	this.headers = new Headers(init.headers || input.headers || {});
 	this.body = init.body || input.body;
 	this.url = url;
 
--- a/lib/response.js
+++ b/lib/response.js
@@ -7,6 +7,7 @@
 
 var http = require('http');
 var convert = require('encoding').convert;
+var Headers = require('./headers');
 
 module.exports = Response;
 
@@ -19,10 +20,12 @@ module.exports = Response;
  */
 function Response(body, opts) {
 
+	opts = opts || {};
+
 	this.url = opts.url;
 	this.status = opts.status;
 	this.statusText = http.STATUS_CODES[this.status];
-	this.headers = opts.headers;
+	this.headers = new Headers(opts.headers);
 	this.body = body;
 	this.bodyUsed = false;
 	this.size = opts.size;
--- a/test/test.js
+++ b/test/test.js
@@ -808,4 +808,18 @@ describe('node-fetch', function() {
 		});
 	});
 
+	it('should support parsing headers in Response constructor', function(){
+
+		var r = new Response(null, {headers: {'foo': 'bar'}});
+		expect(r.headers.get('foo')).to.equal('bar');
+
+	});
+
+	it('should support parsing headers in Request constructor', function(){
+
+		var r = new Request('http://foo', {headers: {'foo': 'bar'}});
+		expect(r.headers.get('foo')).to.equal('bar');
+
+	});
+
 });


bitinn:
2 comments on this:

- The headers part I generally agree with, now that Request returns an instance that's Headers instance, we can probably simplify this as well https://github.com/bitinn/node-fetch/blob/master/index.js#L61-L62

- The string and buffer input to Response, isn't it just easier and more foolproof to do it in your unit test by creating a Buffer? Why rely on Response to parse input string?

kirill-konshin:
Spec allows to pass strings and also Blobs'n'stuff, but we don't cover that. In NodeJS it is useful to allow to pass Buffers.

bitinn:
@kirill-konshin I understand the spec says it. But what I am asking is why you decides to rely on `node-fetch` to do such a thing.

If you are doing unit test that's unrelated to Fetch Spec, ie. you are just using `node-fetch` as a polyfill or a dependency, not testing fetch in general, then why not rely on its existing features? That way you will know it works.

Looking at your code I don't see why you rely on Response accepting String body. Instead of using a helper module like `resumer`.

https://github.com/bitinn/node-fetch/issues/33#issuecomment-126767289

kirill-konshin:
Let's reiterate. Spec says, that strings are allowed. Isomorphic code that uses both native and your implementation may rely on that.

If you don't want buffers -- I'm OK with that. I've added them only because string will be converted to buffer to comply with the rest of your implementation, I wanted to touch as less things as possible.

bitinn:
@kirill-konshin I am a jerk for saying this: but I do think `node-fetch` features relies on developers showing me what they are doing, and me trying to figure out if that hurdle is huge enough to cover in `node-fetch` code.

I don't hate following fetch spec (or add String/Buffer support), but I do avoid features that few developers use, because it's bound to be a maintenance issue. Could you show me how you would use such features? Like what sort of unit testing you are doing that rely your test cases to be isomorphic?

Same argument on `rely on code being used as isomorphic` [has been made before](https://github.com/bitinn/node-fetch/pull/25), so you are not alone, but I need more examples than ppl may rely on it...

kirill-konshin:
Regarding buffers here is a non-intrusive example of `clone()` method in our other conversation https://github.com/bitinn/node-fetch/issues/34#issuecomment-128913428, I'm sure you'll read it there, but I'll leave it here for history.

I can't copy code because it's not yet ready and it will be hard to copy only the minimum essence, sorry for that. 

But here are short examples.

Part of the functionality of my lib is parsing `multipart/mixed` responses into a bunch of actual `Response` objects, with bodies, headers and all. The implementation is tricky because I have to first get the `text()` of original response, then parse it to an array of `Responses`, but I also have to save original response body (this is why it needs to be cloned).

Second, I rely on mocking layer to get rid of network requests during tests (for example to test first example), so if certain flags are on I substitute a call to `fetch()` with a custom one, which creates a new `Response` from text, that was provided by mock.

bitinn:
> The implementation is tricky because I have to first get the text() of original response, then parse it to an array of Responses, but I also have to save original response body (this is why it needs to be cloned).

if that's the case I think you are trying to read the body stream twice (1st to text and then pipe to somewhere), which means all sorts of problems (things like [back pressure](http://stackoverflow.com/questions/19553837/node-js-piping-the-same-stream-into-multiple-writable-targets#comment38541924_19561718) mentioned here).

I saw your suggested code for `.clone()`, while it works, it also means you are buffering all response body, so the cloned `Response` is not really identical and you are no longer streaming... To be honest I don't know how browser does it without adding some sort of limits.

For multipart or other binary data, I will feel a lot more at ease to just get the body stream and save/decode/parse it separately (as I do with my own use-case of `node-fetch` + [sharp](https://github.com/lovell/sharp), where I tried to fetch/save the raw image AND create a thumbnail of it using 1 stream, and failed; so I cache the raw image locally instead).

> so if certain flags are on I substitute a call to fetch() with a custom one, which creates a new Response from text, that was provided by mock

why not create the mock response using a stream that contains your text? I assume you just want to test the result is correct.

kirill-konshin:
> why not create the mock response using a stream that contains your text

Because it's an isomorphic library and I don't what it to know about stream details of `node-fetch`, when I test it in browser, there are no streams... This is why text...

The implementation of `clone()` simply takes whatever has been buffered already and does the same thing as you do in `_decode` - concatenates buffers into one.

bitinn:
I am happy to accept your Headers related commits, but not the String/Buffer support at the moment.

To me, Fetch isn't fully isomorphic without low-level support like Streams and Blobs. You may have much better chance by refactoring your library so that it's more decoupled from `node-fetch`, in a way that allow you to test without needing to mock fetch api (but instead just stub it by returning your test data).

When I got around to refactor a Body class for Request/Response, I will look into supporting better body input when using Request/Response directly.

kirill-konshin:
In the lib I have to return responses to the end user, that's the purpose of the lib :)

I still don't quite understand why

```js
	if (typeof (this.body) == 'string') {
		this.body = new Buffer(this.body);
	}

	if (this.body instanceof Buffer) {
		this._raw = [this.body];
		return Response.Promise.resolve(this._convert());
	}
```

Is not OK... Checks are solid, implementation reuses `_raw` property and `_convert()` method, that are already there, with minimum impact. It's just a shortcut in case body type is known. I think it's safe enough.

I can refactor Req/Res class to have common body implementation. It's straightforward.

kirill-konshin:
Speaking about isomorphism, I think it's not a big deal to have FormData, Blobs and strings at front end and their node equivalents - `form-data` module, Buffers, strings and streams, users should understand limitations and use appropriate objects.

By saying "isomorphic approach to fetch" we must understand, that most likely the actual applications will not be isomorphic, but, for example, my library currently is isomorphic, so with certain limitations mentioned above and in your docs, it can be easily used anywhere.

bitinn:
> I still don't quite understand why ... Is not OK

I don't want to encourage use of Request/Response directly until I am done with Body refactor and extensive test cases. I can then decide on exactly what we are supporting and not supporting in these scenarios. As described here: https://github.com/bitinn/node-fetch/blob/master/LIMITS.md

> my library currently is isomorphic

Which is why I highly recommend you decouple it from `node-fetch`, so that you are can test it as an isomorphic library. Same goes for github's `fetch`, you use as a polyfill, not a dependency.

kirill-konshin:
I see your point about direct usage, but since these classes are already exposed, I think a small fix with appropriate mention in `LIMITS.md` is quite OK. We are talking about such a small change that does not worth the conversation...

kirill-konshin:
Anyway, I've reverted strings and buffers, I will put them as separate PR.

kirill-konshin:
Btw, how about to use `resumer` internally to wrap all body types if they're not streams? Do you insist that this should be done externally?

Since your implementation of Request allows to pass string/buffer bodies (primarily because http(s) allows to do that), Response should have it too for the sake of consistency.

bitinn:
@kirill-konshin I intend to support String and Buffer input when refactoring Request/Response, but no current roadmap just yet, end of august maybe.

I know it sucks not being able to use Request and Response directly in test cases, but it also means your code depends on response having a Headers interface and a Body interface. I think that's strong dependency on fetch api.

You could instead write a helper function that use Headers and Body interface (does nothing else), but return a plain object for your main lib to consume. That way, when you test, just stub the helper function and return the test result. No fetch feature needed. And you only use what's available in both client-side `fetch` and `node-fetch`.

kirill-konshin:
Library relies on WhatWG specification as an interface, plus it's main usage is a wrapper around fetch-related stuff, so this dependency is strong. Roughly, all the library does is creating Requests with certain pre-configured headers for authentication etc. and parses responses in a correct way.

What do you think about adding at least string support for your Responses? It will cure like 90% of my headache )))

kirill-konshin:
Helper function sounds good, but it means that I'll have to implement `text()` & `json()` methods in the mock version. Which I don't really want to do since string support for your lib is just one line of code...

kirill-konshin:
Check out the follow up of this PR and conversation -- new PR https://github.com/bitinn/node-fetch/pull/39

Issue #38, opened 2015-08-10
Support string/Buffer bodies in Response constructor


kirill-konshin:
According to spec

```js
var response = new Response('{"foo":"bar"}');
```

is a valid code.

This is needed for testing purposes because my code relies on WhatWG spec as on interface.

simplyianm:
:+1: please merge one of the below fixes if possible

bitinn:
released as v1.4.0

Issue #36, opened 2015-08-07
Request should have Body methods like text(), json(), etc.


kirill-konshin:
`Request` implements `Body`, so it also has the following methods available to it:

- `Body.arrayBuffer()`
- `Body.blob()`
- `Body.formData()`
- `Body.json()`
- `Body.text()`

https://developer.mozilla.org/en-US/docs/Web/API/Request

bitinn:
I am not going to implement this, as I don't see a real need for it. PR is welcomed but I will need to review the use case first.

kirill-konshin:
It's part of the standard... :)

Use case: library gets `Request` object from user, detects that it's content type is `application/json`, get the body, stringifies, returns a new `Request` object with serialized body.

bitinn:
I will quote [my intention](https://github.com/bitinn/node-fetch#features) for this module.

> Make conscious trade-off when following whatwg fetch spec and stream spec implementation details, document known difference.

I don't want to falsely claim `node-fetch` is providing a full API that covers all client-side use case. It would take considerable more code and tests (or dependencies) to avoid messing up, as a lot of client-side API isn't present on node.js

To achieve what you said we at least need to refactor into a `Body` class, and handle things like Blob and formData which doesn't make much sense in a server use-case context.

kirill-konshin:
I'd say, that `Blob` or `FormData` is an obvious overkill. Simple `json()`, `text()` and `buffer()` methods would be more than enough for both `Request` and `Response`.

kirill-konshin:
Especially for given use-case.

bitinn:
Please do show me your code, I feel there are more solutions to your problem.

kirill-konshin:
So far I have a solution -- I simply access `request.body` property. But for truly isomorphic code I'd like to use `text()` and `json()` methods...

Basically, we're talking about sharing part of the code of `Request` for `Response` class as well, as-is, nothing more. Not a big deal, I think.

tschaub:
It looks like `text()` and `json()` were added in af21ae6c1c964cd40e48e974af56ea2e1361304f.

bitinn:
@tschaub nope, he meant Request and Response instances should all inherit Body's methods like `text` and `json`. To me they are an overkill as of now, if you are calling Request ctor directly, you are on the very edge of Fetch API (even less likely in a server-side context, where you have so many other options, and no service worker to deal with)

kirill-konshin:
[...]

PR #39, opened 2015-08-10
Fix for #38 and #36

kirill-konshin:
Support string/buffer bodies in Response constructor
Request should have Body methods like text(), json()

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -1,4 +1,3 @@
-
 /**
  * response.js
  *
@@ -8,6 +7,7 @@
 var http = require('http');
 var convert = require('encoding').convert;
 var Headers = require('./headers');
+var Body = require('./body');
 
 module.exports = Response;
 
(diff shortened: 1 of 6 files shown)


lib/response.js
```diff
@@ -26,156 +26,13 @@ function Response(body, opts) {
...
- * Decode response as json
- *
- * @return  Promise
- */
-Response.prototype.json = function() {
 
-	return this._decode().then(function(text) {
-		return JSON.parse(text);
-	});
+	Body.call(this, body, opts);
```
TimothyGu: You need at least a `Body.Promise = Response.Promise`

kirill-konshin:
```js
// expose Promise
Response.Promise = global.Promise;
// expose Promise
Body.Promise = global.Promise;
```

It is exposed this way.

TimothyGu: That's only for the default. The whole system is designed in which the module passes down its own `Promise` to the modules it calls.

alubbe:
Any news on this? It'd be really helpful to have this functionality.

kirill-konshin:
I have rebased the PR on top of most recent changes. But CI fails due to:

```
/home/travis/build/bitinn/node-fetch/node_modules/coveralls/bin/coveralls.js:18
        throw err;
              ^
Bad response: 500 {"message":"Build processing error.","error":true,"url":""}
```

I have nothing to do with it...

bitinn:
I have no issue with it, I will fix the coverage when available.

bitinn:
It's been a long road for this PR, thanks for sticking with it :)

kirill-konshin:
No problem )

PR #40, opened 2015-08-19
supports `rejectUnauthorized` option

khtdr:
I want to be able to connect to my own "bad" SSL

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -808,6 +808,28 @@ describe('node-fetch', function() {
 		});
 	});
 
+	it('should not support bad https request', function() {
+		this.timeout(5000);
+		url = 'https://192.30.252.128/';
+		opts = {
+			method: 'HEAD'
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -808,6 +808,28 @@ describe('node-fetch', function() {
 		});
 	});
 
+	it('should not support bad https request', function() {
+		this.timeout(5000);
+		url = 'https://192.30.252.128/';
```
khtdr: This is the github IP. Couldn't think of any other way to test a bad/missing cert.

bitinn:
Actually, passing it a `http.Agent` with the same option works:

https://github.com/bitinn/node-fetch/issues/15#issuecomment-97366559

khtdr:
ah, I missed that. Disregard this then, I will do it that way :)

Thanks

PR #42, opened 2015-09-05
add handle error of response body

dsuket:


bitinn:
Thx for the PR, could you explain to me how do you run into a scenario where the body stream throws error, and what would happen at that point?

Better yet, can you think of a test case for this error handler?

dsuket:
In some pages, despite Content-Encoding is specified to deflate, in fact also be transferred without being compressed.
At that time, an error occurres in the pipe processing of body, it can not handle the error.
There is a possibility that an error occurs when you expand the compression, so I think it is better that it was possible to handle the error of body .
It will show the example below.

```javascript
var fetch = require('node-fetch');
fetch('http://goo.gl/fb/').then(function(res) {
  return res.text();
}).then(function(text) {
  console.log(text);
}).catch(function(err) {
  console.log('err:', err);
});
```

bitinn:
Thx, I will look into it soon.

bitinn:
This is fixed in v1.3.3 release, which would reject body with incorrect content-encoding with a custom error message for easier debugging.

PR #50, opened 2015-10-08
Detect encoding from meta charset at large header

dsuket:
If the meta tag is in the back of the head, charset,  it is not possible to correctly determine the charset.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -140,7 +140,7 @@ Response.prototype._convert = function(encoding) {
 
 	// no charset in content type, peek at response body
 	if (!res && this._raw.length > 0) {
-		str = this._raw[0].toString().substr(0, 1024);
+		str = this._raw.slice(0,4).join('').substr(0, 1024*2);
 	}
 
 	// html5
--- a/lib/response.js
+++ b/lib/response.js
@@ -140,7 +140,17 @@ Response.prototype._convert = function(encoding) {
 
 	// no charset in content type, peek at response body
 	if (!res && this._raw.length > 0) {
-		str = this._raw.slice(0,4).join('').substr(0, 1024*2);
+		if (this.headers.get('transfer-encoding') === 'chunked') {
+			var buf, index = 0, length = 0;
+			while(length < 1024 && (buf = this._raw[index])) {
+				length += buf.length;
+				index++;
+			}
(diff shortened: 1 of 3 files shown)


bitinn:
Please see latest spec on details about encoding sniffing: http://www.w3.org/html/wg/drafts/html/master/syntax.html#determining-the-character-encoding

At some point you are going to run into pages that are badly structured and cause decoding issues (in node-fetch or in browser). Your PR is a workaround to support those pages, which is fine, but I think eventually you will run into worse pages than those.

The end result is having to go through the whole page, instead of the first `n * 1024` bytes.

As it is with all sniffing algorithms, we can only be so accurate before getting false positives. So please let me know what page it is and maybe we can work towards a resolution.

PS: I know the pain of dealing with japanese/chinese websites, so you have my full empathy, though I don't think increasing the `n` in sniffing algorithm is going to solve your longterm headaches :)

dsuket:
thanks for your review and understanding our pain.
ok. i got it.
I had misunderstanding my problem.
The problem is not large head. chunked data is it.
For example, at the website(http://pc.watch.impress.co.jp/docs/news/event/20151007_724488.html) ,
charset definisiton is in first 1024bytes. but that is provided with 'Transfer-Encoding: chunked'.

bitinn:
Interesting, so you mean the first chunk we take has much less than 1024 bytes, I get it now.

Just wondering, does http://pc.watch.impress.co.jp/docs/news/event/20151007_724488.html always give a small chunk (much less than 1024 bytes), or is it random?

dsuket:
Um.. sorry, I am not owner of that site, so i dont know.
I tried to get the url few times, first chunk size was always 32byte.
```
.-----------------------------------------------------.
| chunk size | byte  | content                        |
+------------+-------+--------------------------------+
|         20 |    32 | <!DOCTYPE HTML>%0A<htm         |
|         6f |   111 | xmlns="http://www.w3           |
|        9a2 |  2466 | >%0A<head>%0A <meta http       |
|        2f3 |   755 | <script charset="utf           |
...
```
generated by [chunkview](https://github.com/kazeburo/chunkview)

bitinn:
@dsuket sorry for taking so long, but this is now in v1.4.0, also see changelog :)

dsuket:
@bitinn thanks!

Issue #53, opened 2015-10-26
Missing forEach method in Headers object


tricoder42:
[whatwg-fetch](https://github.com/github/fetch/blob/master/fetch.js#L72) implements `forEach` method in Headers object prototype. I've tried to use `node-fetch` as a backend for [restful.js](https://github.com/marmelab/restful.js/blob/master/src/http/fetch.js#L21), because I'm using `isomorphic-fetch` package. Unfortunately, `restful.js` uses `forEach` method, so it doesn't work on backend.

The question is: Should `Headers` object provide `forEach` method? Is it in specification? I couldn't find anything.

I'm happy to provide a PR with patch unless there's a reason why `Headers` doesn't have `forEach`.

bitinn:
You raised a good point, I read https://github.com/github/fetch/issues/149 and `forEach` it's added due to need for an iterable interface (implied to support `forEach`).

Though I do wonder about your use-case, restful.js is using `request` on server-side, is this truly what blocks them from using `node-fetch`?

tricoder42:
I know it’s using `request` in README. However, the `fetch`/`request` library is wrapped in `fetchBackend`/`requestBackend`, so I was hoping that I could use `node-fetch` instead of `request`. It’s much easier to import `isomorphic-fetch` which imports the right library for the frontend and backend.

I’ll try to implement this method and see if it works :)

bitinn:
sorry for taking this long, but v1.4.0 release has this built-in.

PR #54, opened 2015-10-27
Add Headers.forEach method.

tricoder42:
As discussed in #53


bitinn:
Thx, will publish a new version when I got a chance to do so.

tricoder42:
Nice, thank you :+1: :shipit:

Issue #15, opened 2015-04-28
Allow to pass all available https options


emmenko:
Currently only a subset of all the available options are supported https://github.com/bitinn/node-fetch/blob/master/index.js#L66-L80

For `https` important options are missing (e.g.: `rejectUnauthorized`, `secureProtocol`, `key`, `cert`, ...)
https://nodejs.org/api/https.html#https_https_request_options_callback

I should be able to pass all of those options.

bitinn:
As this is out of scope for Fetch, I am against exposing `https` options in Fetch API.

You will be much better served by using a module that fully support `http` and `https` options (say [request](https://github.com/request/request), you can find `co-` or thenable version of it); alternatively, use [denodeify](https://github.com/matthew-andrews/denodeify) or [bluebird](https://github.com/petkaantonov/bluebird) to promisify `https.request`.

Our intention to keep `node-fetch` API minimal and close to spec, unless there are very strong use-cases. You can find similar discussion on #11

emmenko:
I see your point. Thanks for clarifying.

So I guess I would have to provide a valid SSL certificate even for local development.

bitinn:
I am not exactly sure what you are trying to achieve, but trying passing a `https.Agent` with your CA configuration using node-fetch's `agent` options might help.

http://stackoverflow.com/questions/21004645/where-is-nodes-certificate-store

emmenko:
Nothing special, just I have a https server for local development with a self-signed certificate and I need of course to allow the request with `rejectUnauthorized: false`.

But yes, I saw I could come around with it by configuring the `globalAgent` https://nodejs.org/api/https.html#https_https_globalagent. I will give it a try.

bitinn:
Or you can just create a https.Agent instance and pass it to node-fetch, instead of changing globalAgent.

see `agent` in https://github.com/bitinn/node-fetch#options

emmenko:
Ok, I'll try that as well. Thanks :+1:

emmenko:
Yep, I did it like this then

```js
const agent = new https.Agent({
  key: fs.readFileSync(`${CERT_PATH}.key`),
  cert: fs.readFileSync(`${CERT_PATH}.crt`),
  rejectUnauthorized: false
})
```

mmcgahan:
the main downside of this is that you can't use use an isomorphic script with `fetch` this way - when it runs in the browser there's no `require('https')`.

[...]

PR #65, opened 2015-12-10
added support for rejectUnauthorized option

wouterverweirder:
Was running into SSL certificate issues for certain requests (unable to verify the first certificate).

The standard https library has an option to ignore this (setting the rejectUnauthorized option to false).

This pull request enables you to pass that rejectUnauthorized option.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -47,6 +47,8 @@ function Request(input, init) {
 	this.body = init.body || input.body;
 	this.url = url;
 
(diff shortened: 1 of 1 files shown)


bitinn:
These options are passed as a `http.Agent`, through `options.agent`, see #15 and #40

wouterverweirder:
ok, missed out on that option, sorry about that!

PR #67, opened 2015-12-12
Support being able to extract the ‘location’ header from redirected HTTP requests

matthew-andrews:
What do you think about this?

At the moment we're currently doing this:- 

```js
return shellpromise(`curl -s http://www.ft.com/fastft/?p=${id} -I -H 'X-FT-Access-Metadata: remote_headers' | grep -i location`)
	.then(loc => {
		return loc
			.replace(/location\:/i, '')
			.trim();
	});
```

It would be nice if `node-fetch` supported this use case.

Apologies that the API is so unwieldy.  Very happy to change `throwOnMaximumRedirect` to something more succinct.  Also I will update the docs before this should get merged.

I also added a test that explicitly checked the current behaviour.

cc @wheresrhys

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ language: node_js
 node_js:
   - "0.10"
   - "0.12"
-  - "iojs"
   - "node"
+  - "4"
+  - "5"
 before_install: npm install -g npm
 script: npm run coverage
--- a/index.js
+++ b/index.js
@@ -118,33 +118,37 @@ function Fetch(url, opts) {
 		});
 
 		req.on('response', function(res) {
+			var throwOnMaximumRedirect = opts && opts.throwOnMaximumRedirect === false ? false : true;
 			clearTimeout(reqTimeout);
 
 			// handle redirect
 			if (self.isRedirect(res.statusCode)) {
-				if (options.counter >= options.follow) {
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -2,7 +2,8 @@ language: node_js
 node_js:
   - "0.10"
   - "0.12"
-  - "iojs"
```
matthew-andrews: iojs seemed unavailable on nvm

bitinn:
I need a bit of time to think about this, it's been an improvement I would like to make, per #8

One thing I am unsure about is whether to add such a flag, or just populate rejected promise `err` object with the information one would need (say headers, 2 top requests are location and cookie headers).

wheresrhys:
I would favour a flag over rejecting with info. One or two other libraries
(request is one I think) reject with info, and it's unwieldy to have to
inspect the error for properties.

How about redirectLimit: 'hard'/'soft'

matthew-andrews:
Rejecting with info would work but I agree with @wheresrhys, whilst it'd enable access to the properties that we need it's a bit confusing…  Redirects shouldn't really be treated _like_ errors in my opinion…

bitinn:
v1.5.0 added support for the new redirect mode, so this is no longer needed:

https://fetch.spec.whatwg.org/#concept-request-redirect-mode

PR #69, opened 2016-01-15
fix: support string and Buffer bodies in Response constructor. Fixes #38.

### LINKED ISSUE #38: Support string/Buffer bodies in Response constructor

simplyianm:


bitinn:
Thanks, if this includes #72 and required less change to existing code, I may pick this PR instead.

bitinn:
this is now supported in v1.4.0 :)

PR #70, opened 2016-01-19
Add buffer method to Response. Resolves #51

### LINKED ISSUE #51: Use case for res.arrayBuffer
I'm currently working on a `.docx` file parser that should work on both browsers and NodeJS environments.

https://github.com/MrOrz/hacktabl-parser/blob/fcfc7626c533b0434d11215e4c3c1d305990db26/src/fetchDoc.js

`.docx` files are a bunch of XML files archived in a zip format, thus I am using JSZip to extract XML files from `.docx`. Since `node-fetch` does not support `res.arrayBuffer()`, I cannot use `isomorphic-fetch` to grab `.docx` files on NodeJS.

If `node-fetch` were to support `res.arrayBuffer()`, the large if-statement that differentiates server & browser behavior could be gone. `http` and `https` library don't have to be required in the code anymore, leading us to a smaller bundle when the code is compiled using webpack for use in browsers.

axfree:
This patch add support for accessing the response body stream as a Buffer.
This is similar to arrayBuffer, but in nodejs style.

Use cases:

```javascript
fetch('http://.../Some.TV.Show.S01E01.FeTCH.smi')
    .then(function(res) {
        return res.buffer();
    }).then(function(buffer) {
        var smi = convert(buffer, 'utf-8', 'CP949').toString();
        ...
    });


fetch('http://.../Some.TV.Show.S01E01.FeTCH.sub')
    .then(function(res) {
        return res.buffer();
    }).then(function(buffer) {
        var charset;
        var subType;
        
        if (buffer[0] == 0xef && buffer[1] == 0xbb && buffer[2] == 0xbf) {
            charset = 'utf8';
            buffer = buffer.slice(3);
        }
        if (buffer[0] == '1')
            subType = 'srt';
        else if (buffer[0] == '<' && buffer[1] == 'S')
            subType = 'smi';
        else
            ...
            
        var sub = convert(buffer, 'utf-8', charset).toString();
        ...
    });
```

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -41,7 +41,7 @@ function Response(body, opts) {
  */
 Response.prototype.json = function() {
 
-	return this._decode().then(function(text) {
+	return this.text().then(function(text) {
 		return JSON.parse(text);
 	});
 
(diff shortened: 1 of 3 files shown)


bitinn:
Sorry for the delay in reply, but would just getting the body as a stream and converting it into Buffer better left for user? See #51

axfree:
Because converting stream to buffer is not that trivial. 
This is by far the simplest code I came up with

```javascript
var toArray = require(‘stream-to-array’);

fetch('https://github.com/')
    .then(function(res) {
      return toArray(res.body);
    }).then(function (parts) {
      var buffers = []
      for (var i = 0, l = parts.length; i < l ; ++i) {
        var part = parts[i]
        buffers.push((part instanceof Buffer) ? part : new Buffer(part))
      }
      var buffer = Buffer.concat(buffers)
      console.dir(buffer);
      ...
    });

```
`fetch` does have similar code internally (_decode). Why not expose it for convenience.
This looks much better :)

```javascript
fetch('https://github.com/')
    .then(function(res) {
      return res.buffer();
    }).then(function (buffer) {
      console.dir(buffer);
      ...
    });

```
This patch seems to fix #50 as well.

jimmywarting:
> # Features

> - Stay consistent with `window.fetch` API.
> - ...

> > but would just getting the body as a stream and converting it into Buffer better left for user?

Fail to see your point there @bitinn. If it's going to be consistent with the fetch API then we should have a `body.buffer()` and/or `body.arraybuffer()` even if that means a degration of not using pipes.

I can also see some use cases for having that for some small response data where you don't want pipe it to a file/transformer or anything like that.<br>Checking for magic number (filetype) and for usage with [protobuf.js](https://www.npmjs.com/package/protobuf.js) (aka [protocol buffer](https://developers.google.com/protocol-buffers/)) just to mention a few

bitinn:
@jimmywarting because I don't really see a point for `arrayBuffer` on server-side (as discussed in #51). I think when people need to buffer the full body, they are often missing the better solution to their problem.

For your specific use case, ei. file type detection, you probably only need the first a few chunks, I believe `buffer` is an overkill (it's a great time for a micro module, `first-chunk` and `first-n-chunk` are still available on npm!)

PS: `node-fetch` exists because people have argued for months about porting github's fetch polyfill to node.js: https://github.com/github/fetch/issues/9

jimmywarting:
Please add this

bitinn:
It's a bit late but good news nonetheless: buffer() has landed in v1.6.0

PR #72, opened 2016-01-28
Add string body support to Response

BlakeWilliams:
This adds support to pass in a string value to `Response` instead of
always expecting a pipe. This makes `Response` more compatible with the
spec as well as making it easier to construct a manual `Response`
object.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -88,33 +88,37 @@ Response.prototype._decode = function() {
 		}
 
 		// handle stream error, such as incorrect content-encoding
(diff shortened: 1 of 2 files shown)


bitinn:
+1 for this, but I haven't had the time to review this yet :)

kirill-konshin:
This PR can now be closed since the subject has been covered in https://github.com/bitinn/node-fetch/pull/39

BlakeWilliams:
Awesome, glad that it now supports strings. :smiley:

PR #77, opened 2016-02-10
Allow rejectUnauthorized request flag.

rektide:
Hiya. I'd like to add Node's rejectUnauthorized flag. I've tried to add code with a similar form and function to existing code, and this PR preserves current Node default behavior of rejecting communications with unauthenticated endpoints.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -58,6 +58,9 @@ function Request(input, init) {
 		input.compress : true;
 	this.size = init.size || input.size || 0;
 	this.agent = init.agent || input.agent;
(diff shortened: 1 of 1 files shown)


bitinn:
Hi there, you can pass in a http.Agent, which is designed to handle this type of things.

see
- https://github.com/bitinn/node-fetch/issues/15#issuecomment-97361898
- https://github.com/bitinn/node-fetch#options

bitinn:
closing this PR as feature is already support, see comments

PR #82, opened 2016-02-24
Reject fetch promise with actual request error

jnu:
The real request error has valuable fields such as `code` and `stack`, but these are being swallowed. The `message` that's currently given is only marginally helpful.

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,7 +115,7 @@ function Fetch(url, opts) {
 
 		req.on('error', function(err) {
 			clearTimeout(reqTimeout);
-			reject(new Error('request to ' + options.url + ' failed, reason: ' + err.message));
+			reject(err);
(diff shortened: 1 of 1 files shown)


bitinn:
Fetch Spec actually decide against showing specific errors for security reasons. We are providing error message just for debugging purpose.

Do let us know the use-case and we will consider it in #8

jnu:
oh ok, thanks. In my case the message itself was vague ("parse error"), but the `code` from the original error told me specifically what the issue was (`HPE_UNEXPECTED_CONTENT_LENGTH`). Is there any reason the code can't be included with the message?

bitinn:
@jnu that's a good point, one of our goal in #8 would be to provide a more meaningful error for debugging,

Issue #46, opened 2015-09-20
Support "redirect" attribute of Request


cpetrov:
See https://fetch.spec.whatwg.org/#dom-request-redirect.

The *manual* `RequestRedirect` will allow preventing HTTP 3xx responses from being automatically followed.

bitinn:
This has been raised before #6 #23, but I don't think `manual` is exposed, ie. you can't actually stop fetch from following redirect due to https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

So the `manual` description is likely for browser vendors (internal), not users (API). Also see https://github.com/github/fetch/issues/137

cpetrov:
Thank you for the information! Closing the issue.

PR #84, opened 2016-03-07
Don't error on redirect if follow is zero

quarterto:
I interpret the semantics of `follow: 0` to be "just give me this response whether it's a redirect or not". I don't consider "no redirects" to be "too many" and thus an error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,7 +122,7 @@ function Fetch(url, opts) {
 			clearTimeout(reqTimeout);
 
 			// handle redirect
-			if (self.isRedirect(res.statusCode)) {
+			if (self.isRedirect(res.statusCode) && options.follow > 0) {
 				if (options.counter >= options.follow) {
 					reject(new Error('maximum redirect reached at: ' + options.url));
 					return;


Ghirro:
Completely agree with the semantics of this PR and have just come across the same problem myself.

pekeler:
I also would prefer to get the 3xx response instead of an error if the actual number of redirects is larger than the `follow` option. Assuming this is compliant with the standard.

bitinn:
Sorry but `follow: 0` doesn't suppress redirect, this is according to Fetch spec.

Many has raised this issue before (see #6 #23 #46) and I hope to provide an alternative (see #8) without breaking the mental model where a request can only be (a) fulfilled or (b) rejected.

Adding a `nofollow` flag breaks this mental model because 3xx can either be rejected and fulfilled based on your input; while in both XHR and Fetch spec, 3xx are explicitly transparent (if you stop at 3xx it must be rejected, mostly due to network timeout or follow limit)

pekeler:
Ah, OK. I didn't know. Then I guess you can close this PR.

bitinn:
In release v1.5.0 we added support for redirect mode (see Fetch Spec), now you can do `redirect: manual` to extract full response without following redirects :)

PR #86, opened 2016-03-08
added example in README.md capturing error output

coderofsalvation:


bitinn:
Thx for the PR, but I would go against using `_raw`. Please see my comment at https://github.com/bitinn/node-fetch/issues/85#issuecomment-197224658

PR #87, opened 2016-03-12
Align with es6 modules

KnisterPeter:
This adds a default export so it could be used
with es6 modules default export as well without
breaking api.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ var Headers = require('./lib/headers');
 var Request = require('./lib/request');
 
 module.exports = Fetch;
+module.exports.default = module.exports;
 
 /**
  * Fetch class


bitinn:
@KnisterPeter @kirill-konshin 

I am not used to writing node module for es6 (I mean write in node style but to be compatible with es6), would you point to a brief primer on why this is a good idea? what would be missing without default export?

`module.exports.default = module.exports;`

kirill-konshin:
```js
```

accesses default property of exports. Some loaders are smart enough to see that there is no default and use exports object as default, but it's better to explicitly provide it.

bitinn:
Thx @kirill-konshin, but I still need a bit more explanation: now that we explicitly set `module .exports.default` to `module.exports`, and that `module.exports = Fetch`, are we doing a named export or default export? Is it some kinds of convention for tools that load node modules as es6 modules?

KnisterPeter:
The main point is that from commonjs perspective a default export ist just a named export with the name default. :smile: 
This is kind of over simplified, but it holds. And in commonjs you could export a function directly on the module.exports whereas you could no import this on es6-modules side.

dbo:
Excuse my ignorance, but why is this needed?

Babel already handles this for CommonJS modules, and loaders like SystemJS do this internally when loading CommonJS modules, turning them into ES modules, see https://github.com/systemjs/systemjs/blob/master/dist/system.src.js#L1290

KnisterPeter:
But what about webpack? Or typescript? or &lt;you name it>?

dbo:
Can you elaborate on this problem regarding webpack or typescript? I don't see it.

IMO it's the responsibility of either the used loader and/or transpiler of the *client code* to properly handle imported CommonJS modules.

KnisterPeter:
You are right with you statement. But in the real world not every tool is able to do so. So it is just nice to create exports which are compatible to es6 and commonjs.

PR #88, opened 2016-03-16
Clone method

### LINKED ISSUE #34: Support cloning of Responses
According to MDN https://developer.mozilla.org/en-US/docs/Web/API/Response/clone and the spec https://fetch.spec.whatwg.org/#dom-response-clone Response class should have a clone() method, which seems to be missing now.

### LINKED ISSUE #85: res.text() doesn't work after res.json() fail
Hi, I would like to get the response raw text when res.json() throws an error.

Example scenario: expected json response is not json but an html error. Therefore, res.json() throws an error. 

    var response;
    fetch(url, {
      headers: headers,                                                                                                                                                                           
      body: JSON.stringify(payload),
      method: method.toUpperCase()
    }).then(function(res) {
      response = res;
      return res.json();
    }).then(function(json) {
         // do stuff
      }
    }).catch(function(err) {
       response.text().then(function(html) {                               // <---- doesnt work
        console.log("instead of json, I got this response:");
        return console.log(html);
      });
      return console.error(err);
    });

What's the best way of printing 'corrupt json' responses?

> Note #1 : Im not asking how to check the header (http status codes etc)
> Note #2 : This is a pretty straightforward scenario imho, so an example of this would be very handy in README.md

kirill-konshin:
Closes #34
Closes #85
Added clone() method

Tests for all 3 body types were added, incl. asynchronous test for stream body.

### DIFF
--- a/lib/body.js
+++ b/lib/body.js
@@ -25,6 +25,7 @@ function Body(body, opts) {
     this.timeout = opts.timeout || 0;
     this._raw = [];
     this._abort = false;
+    this._bytes = 0;
 
 }
 
@@ -66,9 +67,7 @@ Body.prototype._decode = function() {
     }
 
     this.bodyUsed = true;
-    this._bytes = 0;
     this._abort = false;
-    this._raw = [];
 
     return new Body.Promise(function(resolve, reject) {
         var resTimeout;
@@ -185,6 +184,17 @@ Body.prototype._convert = function(encoding) {
         , encoding
         , charset
     ).toString();
+
+};
+
+Body._clone = function(source, destination) {
+
+    // reassign the body
+    destination.body = source.body;
+
+    // reassign whatever has been decoded by this moment
+    destination._raw = [].concat(source._raw);
+    destination._bytes = source._bytes;
 
 };
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -70,3 +70,14 @@ function Request(input, init) {
 }
 
 Request.prototype = Object.create(Body.prototype);
+
+/**
+ * @return {Request}
+ */
+Request.prototype.clone = function(){
+
+	var clone = new Request(this);
+	Body._clone(this, clone);
+	return clone;
+
+};
--- a/lib/response.js
+++ b/lib/response.js
@@ -27,12 +27,24 @@ function Response(body, opts) {
 	this.statusText = http.STATUS_CODES[this.status];
 	this.headers = new Headers(opts.headers);
 	this.ok = this.status >= 200 && this.status < 300;
+	this.opts = opts; // save for future
 
 	Body.call(this, body, opts);
 
 }
 
 Response.prototype = Object.create(Body.prototype);
+
+/**
+ * @return {Response}
+ */
+Response.prototype.clone = function(){
+
+    var clone = new Response(null, this.opts);
+    Body._clone(this, clone);
+    return clone;
+
+};
 
 // expose Promise
 Response.Promise = global.Promise;
--- a/test/test.js
+++ b/test/test.js
@@ -986,6 +986,43 @@ describe('node-fetch', function() {
 
     });
 
+	it('should support cloning of Responses with stream body and direct use of .body', function(done) {
+
+		var str = resumer();
+		var body = str.pipe(new stream.PassThrough());
+		var r = new Response(body);
+
+		function readBody(body) {
+			return new fetch.Promise(function(resolve, reject) {
+				var chunks = [];
+				body.on('data', function(chunk) {
+					chunks.push(chunk);
+				});
+				body.on('end', function() {
+					resolve(Buffer.concat(chunks));
+				});
+			});
+		}
+
+		str.queue('foo'); // write first part before clone
+
+		// make sure actual data will arrive with delay
+		setTimeout(function() {
+			str.queue('bar'); // write second part after clone, both should be captured
+			str.end();
+		}, 100);
+
+		// attempt to read with delay as well
+		setTimeout(function() {
+			fetch.Promise.all([readBody(r.body), readBody(r.clone().body)]).then(function(texts) {
+				expect(texts[0].toString()).to.equal('foobar');
+				expect(texts[1].toString()).to.equal('foobar');
+				done();
+			}).catch(done);
+		}, 50);
+
+	});
+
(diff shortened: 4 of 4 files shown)


kirill-konshin:
@bitinn Please review and approve.

bitinn:
Thx for the PR, a few questions:

1. Does Fetch Spec even allow `clone()` to be called after the first `text()`? Body can only be used once, and cloning it after you called `text()` appears to create a race-condition that may throw TypeError. (But current codebase will always allow it because of how it's implemented.)

2. In addition, on node.js platform stream can only be read once, so I think `destination.body = source.body` will still give you that 1 stream, likely already being read. see ref1 link.

3. I think adding a few simpler testcases might also help (we have a simple test server script to help avoid heavy lifting in test script, see ref2 link at the end).

4. currently response clone method is tested, but I think it missed the harder testcases, for instances: can you fetch a json response, but clone and call text/json respectively (as describe in #85); can you fetch a text response, but clone and use both body without issues, etc.

5. Although trivial, maybe request clone should also be tested.

ref1: http://stackoverflow.com/questions/19553837/node-js-piping-the-same-stream-into-multiple-writable-targets
ref2: https://github.com/bitinn/node-fetch/blob/master/test/server.js#L98-L122

bitinn:
I think currently github/fetch implementation of clone is also flawed. I am not sure, but if browser supports stream, they will have the same body problem... That's why the spec says `teeing body's stream`.

kirill-konshin:
1. `clone()` after `text()` or `json()` is pointless since body has been used already and you can't read it twice
2. copying body from src to dest is needed so that new object will keep reading same stream on top of copied `_raw` array
3. i'll add more server-based tests later, but for this particular case I think it's fine to work with streams directly to mimic needed behavior and sequence of events
4. `json()` is just a wrapped `text()`, also the author of #85 confirmed that it worked for him
5. sure

Let's merge this for now and I'll add more tests in the next few days.

bitinn:
> Let's merge this for now

-1, please demonstrate your clone method allow this:

fetch a slow text response, clone, use both response body as a stream, and you are able to read full response from both stream without problem.

bitinn:
> clone() after text() or json() is pointless since body has been used already and you can't read it twice

Am I missing something, I think your testcases do exactly just that?

kirill-konshin:
My test does this:

1. Create stream-based response
2. Immediately start reading text
3. Write first chunk
4. Clone
5. Start reading clone text
6. When both texts are resolved -- compare

I did not quite understood what you meant by saying "use both response body as stream" -- per spec there are no such methods... And you can't access `.body` directly.

Let me paraphrase, the one should not `clone()` when `text()` was resolved.

bitinn:
> 2. Immediately start reading text
> ...
> Let me paraphrase, the one should not clone() when text() has been resolved.

This is my question 1, you shouldn't need this, you should always call `clone` before the 1st `text`, because no one guarantees response won't be consumed at the time you call `clone`, hence race-condition.

> use both response body as stream ... per spec there are no such methods

On node.js you have streams, it's simple really, `res.body.pipe(...)`. So I made it a a feature:

https://github.com/bitinn/node-fetch#features

Fetch Spec will eventually have this as well, once Stream Spec is settled... just probably not `.pipe()` like node.js, see Fetch Spec's `progressive` example.

https://fetch.spec.whatwg.org/#fetch-api

(like any live spec, that section is added after `node-fetch` implementation...)

I am sorry if my further response appears slow...

kirill-konshin:
Technically yes, you should never call `clone()` after `text()`. In test case I wanted to make sure that even with reversed order of calls whatever has been read already won't be lost...

`res.body.pipe` -- that's against spec, body is should not be accessible. I propose to add two non-spec methods: `.stream()` and `.buffer()`, but no direct access to property.

If the stream has been read, or it had emitted a chunk at least one, if you clone, you'll never see it, because `data` event will be missed, that's according to current implementation. So if you want to get stream of clone, we should re-stream whatever has been read and pipe the initial stream into that second one. Tricky...

I think we should go at least with what I've added to this PR and potentially add something on top.

kirill-konshin:
Wow, seems that the spec has been upgraded, formerly body was not accessible...

kirill-konshin:
Anyway, we're not obligated to support everything at once, let's make enhancements step by step.

kirill-konshin:
I've added a direct body read test.

bitinn:
I think what I will do at this point:

- find an afternoon in the weekend
- fix the stream piping issue
- improve code coverage and code formatting due to recent PR
- publish a new version

Please note that your current test failed on node v0.10, see travis

As to why I would want to clone and pipe body: imagine you write an image proxy like the one github currently use for external image, you might want to save both the original file and pipe the file somewhere for image resizing. It's a real problem I experienced without `clone` method, but that's an inconvenience at best.

So far people come to `clone` to workaround API problems, we shouldn't push to release a partially implemented feature just so ppl can workaround their API issues (when there are alternatives to `clone`).

kirill-konshin:
Clone is an officially recommended by MDN way to solve the issue that you've see in #85.

```
clone() throws a TypeError if the response Body has already been used. In fact, the main reason clone() exists is to allow multiple uses of Body objects (when they are one-use only.)
```
https://developer.mozilla.org/en-US/docs/Web/API/Response/clone

I will add a condition for used bodies. Partially implemented feature is better than nothing, especially if implementation is following spec.

bitinn:
> Partially implemented feature is better than nothing

I don't agree here, and that implementation isn't following spec because your current implementation is not teeing the body properly, you just copy it like a primitive value.

bitinn:
Furthermore, I take issue with this code:

```
// reassign whatever has been decoded by this moment
destination._raw = [].concat(source._raw);
destination._bytes = source._bytes;
```

you shouldn't need this because you shouldn't have started to read the body, you should always clone the un-read body.

kirill-konshin:
I said, I will adjust implementation to throw an exception if body has been used already. So that code you mentioned won't be needed anymore.

bitinn:
> I will adjust implementation to throw an exception if body has been used already

ok that's relieving to hear, I didn't see it in previous comments

bitinn:
v1.4.0 now has the `clone` method, with proper body stream support. So I am closing this. Your credits has been noted in [changelog](https://github.com/bitinn/node-fetch/blob/master/CHANGELOG.md) :)

PR #89, opened 2016-03-17
Export window.fetch in browser bundles

dsblv:
There's a conventional `browser` field in `package.json` which different bundlers (browserify, webpack, rollup) look into to find a browser-specific version of the module.

Since `node-fetch` is already an implementation of a browser feature, users (me included) may want to *not* incude it in their bundles.

Even though this is easily achievable through bundle configs, I thought it'd be a nice move towards universal node/browsers modules.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,0 +1,8 @@
+/**
+ * browser.js
+ *
+ * exports window.fetch when bundled by browserify/webpack
+ */
+
+module.exports = window.fetch;
+module.exports.default = module.exports;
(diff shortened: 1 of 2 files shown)


kirill-konshin:
Looks good to me

bitinn:
I think that's what isomorphic-fetch does https://github.com/matthew-andrews/isomorphic-fetch

bitinn:
I will close this given isomorphic-fetch exists :)

PR #92, opened 2016-03-23
Fix formdata support when you are trying to wrap the request

### LINKED ISSUE #91: Integration with FormData is broken in 1.4.0
```js
if (bodyStream(body)) {
    pass = new PassThrough();
    body.pipe(pass);
    body = pass;
}
```

This corrupts the FormData object and the logic in `fetch` function:

```js
		// detect form data input from form-data module, this hack avoid the need to pass multipart header manually
		if (!headers.has('content-type') && options.body && typeof options.body.getBoundary === 'function') {
			headers.set('content-type', 'multipart/form-data; boundary=' + options.body.getBoundary());
		}

		// bring node-fetch closer to browser behavior by setting content-length automatically for POST, PUT, PATCH requests when body is empty or string
		if (!headers.has('content-length') && options.method.substr(0, 1).toUpperCase() === 'P') {
			if (typeof options.body === 'string') {
				headers.set('content-length', Buffer.byteLength(options.body));
			// detect form data input from form-data module, this hack avoid the need to add content-length header manually
			} else if (options.body && typeof options.body.getLengthSync === 'function') {
				headers.set('content-length', options.body.getLengthSync().toString());
			// this is only necessary for older nodejs releases (before iojs merge)
			} else if (options.body === undefined || options.body === null) {
				headers.set('content-length', '0');
			}
		}
```

becomes useless.

That was the reason why in my implementation I did not re-stream body and used it as is.

bitinn:
closes #91 

Note that we can't really properly clone a FormData object without having it as a dependency of node-fetch, so I choose to just pass the reference ahead. Hopefully no one is going to clone Request instance with FormData object and somehow expect it to be a complete new clone.

### DIFF
--- a/lib/body.js
+++ b/lib/body.js
@@ -206,11 +206,14 @@ Body.prototype._clone = function(instance) {
 	var pass;
 	var body = instance.body;
 
+	// don't allow cloning a used body
 	if (instance.bodyUsed) {
 		throw new Error('cannot clone body after it is used');
 	}
 
-	if (bodyStream(body)) {
+	// check that body is a stream and not form-data object
+	// note: we can't clone the form-data object without having it as a dependency
+	if (bodyStream(body) && typeof body.getBoundary !== 'function') {
 		pass = new PassThrough();
 		body.pipe(pass);
 		body = pass;
--- a/test/test.js
+++ b/test/test.js
@@ -800,13 +800,26 @@ describe('node-fetch', function() {
 		});
 	});
 
-	it('should allow cloning a json response, and log it as text response', function() {
+	it('should allow cloning a json response and log it as text response', function() {
 		url = base + '/json';
 		return fetch(url).then(function(res) {
 			var r1 = res.clone();
-			return fetch.Promise.all([r1.text(), res.json()]).then(function(results) {
-				expect(results[0]).to.equal('{"name":"value"}');
-				expect(results[1]).to.deep.equal({name: 'value'});
+			return fetch.Promise.all([res.json(), r1.text()]).then(function(results) {
+				expect(results[0]).to.deep.equal({name: 'value'});
+				expect(results[1]).to.equal('{"name":"value"}');
+			});
+		});
+	});
+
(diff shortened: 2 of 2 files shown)


kirill-konshin:
I am still not convinced that we actually need to re-stream, because even if clone will read the same stream -- nothing will break...

bitinn:
@kirill-konshin I have added an additional testcase to explain this. It will fail without piping the body.

https://github.com/bitinn/node-fetch/pull/92/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R814

kirill-konshin:
I see. It fails because the second read is after the first one is completed. Makes sense. But this could be done a bit differently:

```js
Promise.all([res.json(), res.clone().text()]).then(function(stuff){
    var json = stuff[0], text = stuff[1];
});
```

In this case no re-stream is needed.

bitinn:
@kirill-konshin yep that test is still there if you look a few lines up, but I want to make sure the stream feature didn't cause problem I personally ran into :)

kirill-konshin:
Sure.

I am a bit worried about brute force of the solution :) We should probably create a ticket in FormData project asking for ability to clone FormData... And while they're working on it we can find a way to clone FormData externally. Just in case.

bitinn:
It's not in the FormData spec, the next best things I can think of at the moment is `FormData.entries()` to loop over key/value pairs but you still need to construct the FormData instance somehow.

`FormData.clone()` would be nice but since `form-data` also support Node.js stream like us, they will have to pipe and clone body as well. Somehow I feel end users shouldn't try to push it to the edge.

Fetch Spec probably doesn't allow you to clone a FormData body as-is (I mean with all those FormData methods), the cloned body will probably just be a readable stream.

So that's not something I actively seek to fix. Just don't clone your Request with `form-data` body and use it as a new one. Alternatively, only take the FormData.body and setup headers properly yourself. (note that `form-data` has their reasons on why `content-length` isn't set in their headers...)

PR #98, opened 2016-04-02
Method to not follow redirect

jimmywarting:
Way you would do it in browser

```
let req = new Request('http://localhost:3001', {redirect: 'manual'})
window.fetch(req).then(res => console.log(res))

let req = new Request('http://localhost:3001', {redirect: 'error'})
window.fetch(req).catch(err => console.log(err instanceof TypeError))
```

```
var express = require('express')
var app = express();

app.get('/', function(req, res){
	res.set('Location', 'http://localhost:3001/hoho');
	res.status(301).send('moved temporarily')
});

app.get('/hoho', function(req, res){
	res.send('hej')
});

app.listen(3001);
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,8 +127,15 @@ function Fetch(url, opts) {
 		req.on('response', function(res) {
 			clearTimeout(reqTimeout);
 
+			
 			// handle redirect
-			if (self.isRedirect(res.statusCode)) {
+			if (options.redirect !== 'manual' && self.isRedirect(res.statusCode)) {
+				
+				if (options.redirect === 'error'){
+					reject(new TypeError("Failed to fetch"))
+					return
+				}
+				
 				if (options.counter >= options.follow) {
 					reject(new Error('maximum redirect reached at: ' + options.url));
 					return;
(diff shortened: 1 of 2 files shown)


bitinn:
please see discussion #6 #23 #46 on why browser doesn't actually expose `manual` to user.

bitinn:
Hmm wait, now the spec has changed to make this possible to end-user? I will have to look into this. I will keep this PR open.

Note: Github's fetch polyfill can't polyfill this behaviour because XHR treats 3xx as transparent.

jimmywarting:
Yupp, it's possible for the end user. not just possible for the polyfill.

jimmywarting:
Yup, it's possible for end user if you create a `new Request('', {redirect: 'manual'})` but making `fetch('', {redirect: 'manual'})` dosen't. wonder why... bit strange to me. maybe not just implemented yet...

bitinn:
This is now in v1.5.0, I added a bit more code to handle location normalization and credited you in CHANGELOG :)

PR #99, opened 2016-04-03
Error codes

pekeler:
Following up on #7 

test.js's formatting looks bad unless the tab size is 2. Suggestions?

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,7 @@ var Body = require('./lib/body');
 var Response = require('./lib/response');
 var Headers = require('./lib/headers');
 var Request = require('./lib/request');
+var error = require('./lib/error');
 
 // commonjs
 module.exports = Fetch;
@@ -114,14 +115,14 @@ function Fetch(url, opts) {
 			req.once('socket', function(socket) {
 				reqTimeout = setTimeout(function() {
 					req.abort();
-					reject(new Error('network timeout at: ' + options.url));
+					reject(error('network timeout at: ' + options.url, 'ETIMEDOUT'));
 				}, options.timeout);
 			});
 		}
 
 		req.on('error', function(err) {
 			clearTimeout(reqTimeout);
-			reject(new Error('request to ' + options.url + ' failed, reason: ' + err.message));
+			reject(error('request to ' + options.url + ' failed, reason: ' + err.message, err.code));
 		});
 
 		req.on('response', function(res) {
@@ -130,12 +131,12 @@ function Fetch(url, opts) {
 			// handle redirect
 			if (self.isRedirect(res.statusCode)) {
 				if (options.counter >= options.follow) {
-					reject(new Error('maximum redirect reached at: ' + options.url));
+					reject(error('maximum redirect reached at: ' + options.url, 'ENREDIRECT'));
 					return;
 				}
 
 				if (!res.headers.location) {
-					reject(new Error('redirect location header missing at: ' + options.url));
+					reject(error('redirect location header missing at: ' + options.url, 'ENOLOCATIONHEADER'));
 					return;
 				}
 
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ var Body = require('./lib/body');
 var Response = require('./lib/response');
 var Headers = require('./lib/headers');
 var Request = require('./lib/request');
-var error = require('./lib/error');
+var FetchError = require('./lib/error');
 
 // commonjs
 module.exports = Fetch;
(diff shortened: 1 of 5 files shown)


bitinn:
You need to have a tab size of 4 space (use tab instead of space).

I think this is a step in the right direction, though I would prefer to define error as constant in the Error interface.

Use `TypeError` if you can as Fetch Spec appears to throw TypeError on most (if not all) occasions.

pekeler:
> I think this is a step in the right direction, though I would prefer to define error as constant in the Error interface.

@bitinn I'm not seeing any other use of `const` anywhere. Or do you mean something else?

pekeler:
@bitinn as for the Travis CI failure, when I run `npm run-script coverage` locally against trunk of this project (not just my fork), I'm getting the same error. Know what this is about?

bitinn:
Now that I have given it more thought:

1. I prefer a custom error class, aka `new FetchError` instead of a `error()` function
2. It should inherit prototype from `Error`, giving it `.name` and `.message` property
3. Set `.name` to `FetchError` so we know this is a custom error from `node-fetch`
4. `FetchError` ctor should accept 2 string argument: `message` and `type`
5. Set `.message` to `message`, it should be human readable
6. Set `.type` to type, it should be machine readable and preferably simple. I am not sure if http://man7.org/linux/man-pages/man3/errno.3.html covers our error code need, if not, I prefer a custom type like `MissingLocation` to `ENOLOCATION` because user knows immediately it's not from underlying system (to me, things like `ERANGE` implies error from system, see https://nodejs.org/api/errors.html#errors_system_errors)
7. if you want `.code` then it should be a number, and that number should actually be a mapping from `type` to `code`, where you define within FetchError (so when a user try to handle the error code, he only has to look at FetchError file, not hunting for code number.)

bitinn:
coverage is provided by coveralls.io, their api is not the most stable, so something request fails

https://github.com/lemurheavy/coveralls-public/issues/752

pekeler:
Not disagreeing with anything you're suggesting. However, using `.code` with string values on top-level Error seems to be common in node. Also, inventing new E* codes is also common. For example, ENOTFOUND is not an actual POSIX error but something invented by node.

Quote from the system_errors link you provided (with my own highlights), 

> error.code: Returns a **string** representing the error code, which is **always** E followed by a sequence of capital letters, and **may** be referenced in man 2 intro.

So, code is not a number, but a string that always starts with "E" and that does not have to be a system error.

If you still want to divert from node for this, I'll implement your suggestions. ... tomorrow

bitinn:
```
error.code#

error.errno#

Returns a string representing the error code
```

Did you read it wrong?

bitinn:
Taking back what I said, I guess `.code` is string afterall. But I still think `ENOTFOUND` is better left for node.js core (I personally have never need to handle a `E` related error by checking it's name)

bitinn:
So IMO:

If we use `.type`, then it should be a custom name like `MissingLocation`;
If we use `.code`, then it should be like node.js style `ENOLOCATION`;

I guess I can live with either, but personally more comfortable with former due to having doubts on whether userland code should invent system error name. I will go ask about it...

Also note that there are other places where Error is thrown, such as in Request.js

pekeler:
One interesting error case is the timeout. Timeouts can happen on the system level (ETIMEDOUT) and from within node-fetch. Both are configurable. So depending on which timeout value is shorter, the exact same scenario can either cause an ETIMEDOUT or a timeout by node-fetch.

I'm using node-fetch in my program. Now I'd claim that to the user of my program it makes no difference from which layer in the stack the timeout originated. They just care that something was too slow and therefore timed out. It doesn't even make a difference to my program. I assume that same thinking led the Request project to make their timeouts look like a system level timeout (https://github.com/request/request/blob/a7ef602e90fae6c1c715c461d11cd78802ff4c2d/request.js#L768).

bitinn:
@pekeler I am currently leaning towards `.type` solution (or `.code` with custom error code instead of `E`-style), but it would be interesting to see other's input.

I want to clearly differentiate from native system error. We can't really cover the linux timeout limit without user changing their system, so I would rather user know which is which, so they can fix it accordingly (without me troubleshooting them with github issues)

pekeler:
> Also note that there are other places where Error is thrown, such as in Request.js

The errors in Request.js are thrown due to programmer errors (because node-fetch documents what an acceptable URL is). I'd prefer this PR to focus on operational errors. See https://www.joyent.com/developers/node/design/errors for more details about the difference.

pekeler:
OK, we now have a FetchError for operational errors. FetchError always has a type property. And if the underlying problem is a system error, the code property of that system error is copied into the FetchError. Invented E-style error codes are not used anymore.

The types I've picked are:

- socket-timeout
- body-timeout
- max-size
- max-redirect
- invalid-redirect
- system

bitinn:
@pekeler Thx! I will need to find time to review this PR :)

bitinn:
released in v1.5.0, I should note that I changed `socket-timeout` to `request-timeout` because it's not actually socket timeout but custom timeout limit set by user.

Issue #101, opened 2016-04-08
Content-Length header not sent with DELETE request with a body


jpierson:
After some troubleshooting with an API integration test I where node-fetch was being used I found that the body of my request in Express was showing up as `Object` without any of the properties that I has specified originally on the JSON body with the request. Looking into a bit it appears that DELETE requests are generally allowed to have a request body although it's debatable whether it is correct to use it as seen by the [behavior in Angular](http://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request). After digging into the problem by looking into any [reported issues for the body-parser package](https://github.com/expressjs/body-parser/issues/69) typically used with Express it was pointed out that `Content-Type` and `Content-Length` should be specified. Looking into my request with the debugger I could see that the Content-Length field was not being filled in for DELETEs so I added it manually to the headers of the request and now body-parser seems happy and everything appears to work as expected.

Should `Content-Length` be set automatically with DELETEs as it seems to be with other request verbs?

bitinn:
Depends on how you pass the body, `content-length` may or may not be set. But the only instance where we actively set `content-length`, is when you pass it a `form-data` body. Anything else is managed by native `http` api.

For example, if you pass `node-fetch` a stream as body, the `content-length` is never set, regardless of methods.

bitinn:
I got more time to look at this today and do find we set `content-length` less consistently than I would hope to.

https://github.com/bitinn/node-fetch/blob/master/index.js#L91-L101

I think this could be added for a future release, doesn't seem necessary to exclude `DELETE`.

mcfedr:
It seems this would be fixed by removing the check `options.method.substr(0, 1).toUpperCase() === 'P'` before automatically setting the content-length

bitinn:
DELETE request now sets `content-length` IF your body is a string (you need to set content-length manually if you use stream as body, as it is with other request type)

released as v1.5.3

PR #106, opened 2016-04-14
Handle error form-data cannot calculate content-length

mcfedr:
This happens when trying to send a stream of an unknown length

Fixes https://github.com/bitinn/node-fetch/issues/102

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,7 +93,9 @@ function Fetch(url, opts) {
 				headers.set('content-length', Buffer.byteLength(options.body));
 			// detect form data input from form-data module, this hack avoid the need to add content-length header manually
 			} else if (options.body && typeof options.body.getLengthSync === 'function') {
-				headers.set('content-length', options.body.getLengthSync().toString());
+				try {
+					headers.set('content-length', options.body.getLengthSync().toString());
+				} catch (e) {}
 			// this is only necessary for older nodejs releases (before iojs merge)
 			} else if (options.body === undefined || options.body === null) {
 				headers.set('content-length', '0');


bitinn:
Sorry, nope, programming error shouldn't be caught, it should be exposed, so you know you shouldn't pass it a `form-data` with a stream without setting `content-length` explicitly.

mcfedr:
Its not a programming error, there are cases when you wont know the content-length of a stream you are trying to send.
As far as I am aware, can can test, its fine to send without content-length in this case.
I could be including a stream of a file i am downloading from another server, I dont know the length.

bitinn:
> its fine to send without content-length in this case.

The whole point of add content-length automatically is because some servers won't allow post without content-length.

However I do understand your pain, so I am asking `form-data` about that too. But try/catch just isn't a viable solution.

ref:
- https://github.com/bitinn/node-fetch/commit/d60e9acbca37429a2479955fa504b9c0d8d176d3
- https://github.com/form-data/form-data/issues/196

mcfedr:
I see. I had tested in node and with a apache+php server and didn't see problems not setting content-length.

I have added a test for node-fetch that attempts to send a file using `form-data` https://github.com/mcfedr/node-fetch/commit/706d2fd32005353ef7d2980eac6debba4d4286a7 - it fails at the moment

I can understand why the empty `catch` isn't great, maybe add something to check for the specific error that is thrown and if its anything else `throw` it again?

There are plenty of cases when `node-fetch` doesn't set a `content-length` header, if i just set a stream as the body for example

bitinn:
As for when to add `content-length`, I will work on it in #101, I will add better support for non-stream first.

The suggestion for now is: you should properly figure out `content-length` yourself, most of the time you can (sometimes it means you have to count the stream first, but `node-fetch` isn't magic, all we can do is to count the stream for you)

mcfedr:
For the moment I have done

    data.getLengthSync = null;

The prevents `node-fetch` from trying to find the length, and everything works.

mcfedr:
The problem with counting the stream is that the whole steam ends up buffered in local memory, thats no good if I need to send a large file that I am downloading.

For example, I have a large file on S3, I have opened a stream to that file, and I want to POST it somewhere else.

bitinn:
> The problem with counting the stream is that the whole steam ends up buffered in local memory, thats no good if I need to send a large file that I am downloading.

Depends the size of the file, you should pipe it to a local temp folder for now. If you are POST ing something in the size of TB/PB, `node-fetch` isn't the best tool for that.

mcfedr:
What this seems to come down to, it that you are convinced that 'content-length' needs to be set, and I think its fine not to set it.

bitinn:
You have misunderstood my intention. My intention is to fix this issue of *having to call getLengthSync when you clearly don't need it*. But not by catching error or setting `getLengthSync  = null`.

Please see my reply on https://github.com/form-data/form-data/issues/196

Issue #102, opened 2016-04-12
Cannot calculate proper length in synchronous way. (node-fetch + form-data)


mahnunchik:
```
Error: Cannot calculate proper length in synchronous way.
      at FormData.getLengthSync (node_modules/form-data/lib/form_data.js:245:17)
      at node_modules/node-fetch/index.js:96:48
      at new Fetch (node_modules/node-fetch/index.js:49:9)
      at Fetch (node_modules/node-fetch/index.js:37:10)
```

This line causes the error: https://github.com/bitinn/node-fetch/blob/master/index.js#L96

bitinn:
This is a hack to support `form-data` module, so you don't have to set `content-length` header manually. You might want to figure out why `form-data` throws this error in the first place (probably related to your body data type).

mahnunchik:
@bitinn 

I've tried to send following body:
```js
form.append('my_file', fs.createReadStream('/foo/bar.jpg'));
```

bitinn:
Due to the nature of callback, we don't currently use `getLength` to determine `form-data`'s `content-length`, but you can calculate the `content-length` yourself (either with `FormData.getLength` or count the stream byte itself) then pass it onto `node-fetch`'s `headers` option manually, this will skip the hack as `content-length` is provided already.

mcfedr:
I have worked around this bug for the moment by removing the `getLengthSync` method before passing it to `node-fetch`

    form.getLengthSync = null;

bitinn:
@mcfedr please don't this, why can't you call `getLength` to figure out `content-length`?

bitinn:
Alternative, don't use `form-data`, because you are just sending files right? Only use `form-data` to manage your custom headers, let `node-fetch` handle the body as native stream instead of `form-data` instance.

mcfedr:
For local files its possible to use `fs.stat` and set `knownLength` when adding to `form-data`

I am sending another stream, and I dont know the length of the stream.

https://gist.github.com/mcfedr/0a6fbaeaf0cfaddbeadfe145c002c8e5

bitinn:
Quick summary: unfortunately `form-data` doesn't have an API for checking whether we can call `getLengthSync` yet, so we have 3 options:

1. to do it correctly and manually, end user can do the calculation with `form.getLength` callback and overwrite our default behavior with custom `content-length`
[...]

PR #107, opened 2016-04-14
Add a test for posting files

mcfedr:
This fails at the moment, but should help with fixing the issue


bitinn:
#102 is closed with v1.5.3, close this PR as well.

PR #117, opened 2016-05-18
Fixes json parsing of an empty response body

joernroeder:
ran into this issue while i was fetching a express `res.status(200).end();` aka. _ok, but nothing here_ response.

### DIFF
--- a/lib/body.js
+++ b/lib/body.js
@@ -40,7 +40,7 @@ function Body(body, opts) {
 Body.prototype.json = function() {
 
 	return this._decode().then(function(text) {
-		return JSON.parse(text);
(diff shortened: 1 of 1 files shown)


TimothyGu:
The [Fetch spec](https://fetch.spec.whatwg.org/#concept-body-consume-body) disagrees with this. There are no special provisions for empty responses. Plus, returning an empty string as JSON certainly makes no sense.

A better choice is to send `{}` on the server.

bitinn:
Yep empty response is invalid json, if it's a third-party api and developer is unwilling to fix it, you have to do a `text()` first.

joernroeder:
@TimothyGu @bitinn you're right. thx

PR #119, opened 2016-05-19
Adds polyfill support

tarikjn:
This PR adds Polyfill support for `node-fetch`, this is required to use 3rd party packages which depend on fetch.
See the `README` changes for more info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,23 @@ Hence `node-fetch`, minimal code for a `window.fetch` compatible API on Node.js 
 
 `npm install node-fetch --save`
 
+# Polyfill
+
+External packages that depend on `fetch` will generally expect `fetch` to be available on the global object, which means that it need to be used as a polyfill to work with these. To do so, simply add `require('node-fetch/polyfill')` at the top of your application entry point.
+
+If your use Webpack to package your Node code, you can simply add `node-fetch/polyfill` in the `entry` configuration option before your application entry point e.g.:
+
+```javascript
+module.exports = {
+   entry: ['node-fetch/polyfill', './app/js']
+};
+```
+
 
 # Usage
 
 ```javascript
+// if not using the polyfill
 var fetch = require('node-fetch');
 
 // If you are not on node v0.12, set a Promise library first, eg.
--- a/polyfill.js
+++ b/polyfill.js
@@ -1,1 +1,26 @@
-global.fetch = require('./index.js');
+// commonjs
+module.exports = polyfill;
+// es6 default export compatibility
+module.exports.default = module.exports;
+
+// this is exported as a function so it can be ran multiple times for testing
+// while going around Nodejs module caching
+function polyfill() {
+
+	if (global.fetch !== undefined ||
+		global.Request !== undefined ||
+		global.Headers !== undefined ||
+		global.Response !== undefined) {
+
+		throw new Error('a fetch (or Request, Headers, Response) object is already defined globally');
+	}
+
+	var Fetch = require('./index.js');
+	global.fetch = Fetch;
+	global.Request = Fetch.Request;
+	global.Headers = Fetch.Headers;
+	global.Response = Fetch.Response;
+}
+
+// polyfill automatically
+polyfill();
--- a/polyfill.js
+++ b/polyfill.js
@@ -1,0 +1,1 @@
+global.fetch = require('./index.js');
--- a/test/test.js
+++ b/test/test.js
@@ -27,7 +27,7 @@ fetch.Promise = fetch.Promise || bluebird;
 var url, opts, local, base, polyfill;
 
 // dry-run polyfill to go around Nodejs module caching
-isolateTest(() => {
+isolateTest(function () {
 	polyfill = require('../polyfill.js')
 })
 
@@ -1286,14 +1286,14 @@ describe('node-fetch', function() {
 	});
 
 	it('should provide a polyfill', function() {
-		isolateTest(() => {
+		isolateTest(function() {
 			polyfill();
 			expect(global.fetch).to.be.an.instanceof(Function);
 		})
 	})
 
 	it('should expose other fetch globals', function() {
-		isolateTest(() => {
+		isolateTest(function() {
 			polyfill();
 			expect(global.Request).to.be.an.instanceof(Function);
 			expect(global.Headers).to.be.an.instanceof(Function);
@@ -1302,7 +1302,7 @@ describe('node-fetch', function() {
 	})
 
 	it('should throw an error if globals are polluted', function() {
-		isolateTest(() => {
+		isolateTest(function() {
 			global.fetch = 1;
 			expect(function() {
 				polyfill();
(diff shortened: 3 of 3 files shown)


jimmywarting:
I beleve this was explicitly built for node. For browser version there is the https://github.com/github/fetch

tarikjn:
This is for Node exclusively, NOT for the browser, NOT for isomorphic environments. You cannot use 3rd party modules that depend on or extend fetch, e.g. [fetchival](https://github.com/typicode/fetchival), [frisbee](https://github.com/glazedio/frisbee) etc. if you don't polyfill `fetch` in Node.

Same with the documentation for Webpack, it is exclusively for demonstrating how to use with the Webpack node target mode and will not work on the browser.

TimothyGu:
> You cannot use 3rd party modules that depend on or extend fetch, e.g. [fetchival](https://github.com/typicode/fetchival), [frisbee](https://github.com/glazedio/frisbee) etc. if you don't polyfill fetch in Node.

https://github.com/typicode/fetchival#usage-examples has a Node.js example that doesn't use global `fetch`.

frisbee recommends isomorphic-fetch and uses global `fetch`, since it is designed to be isomorphic, not "Node exclusively."

All in all, this patch just goes against the Node.js way. If a Node.js-only module requires `fetch` being in the global scope, the module is doing it wrong.

tarikjn:
@TimothyGu I don't agree with your assessment, in particular because `fetch`

TimothyGu:
@tarikjn, you raised some good points, but I believe that is slightly misguided.

`babel-polyfill` deals with a different issue: it polyfills standard JavaScript/ECMAScript features, while Node.js is designed as a full-featured JavaScript interpreter. In other words, `babel-polyfill` helps Node.js better fulfill its original mission.

`node-fetch` is a different case. Fetch is not part of standard JavaScript, but rather available as an open Web API. Node.js is _not_ designed as a browser implementing Web APIs, and globalizing Fetch is actually making Node.js something it is not.

For a more radical example, one could look at how `Promise` is handled in `node-fetch`. Even though Promise is part of standard ECMAScript, instead of relying on a global `Promise` being always available as a browser module would, `node-fetch` allows using a custom `Promise`.

For the point about identical APIs, I believe your argument can be reduced to the following statement: because `node-fetch` emulates the browser API, it should emulate the browser behavior to the fullest extent – including a global reference, despite that being a big no-no in Node.js land. This is just like an Englishman who is learning French, insisting on speaking French in England, despite English being the common language in England.

----

That being said, I will not be extremely upset if this pull request gets merged, since the polyfill is not imported by default – what I _am_ against is thinking of this module as if it is designed to be a polyfill in the sense that `babel-polyfill` is, which it is not.

----

Even if this patch is to be merged, a check for `global.fetch` is necessary so we aren't causing any conflicts (a problem only possible to occur if globals are polluted). We also might want to make global `Response`, `Request`, and other related classes, since they are marked as `Exposed` in the spec.

bitinn:
Here is my view on this:

1. A module designed for node.js should not depend on a global fetch, because node.js core doesn't have it.

2. I don't see why `node-fetch` need to provide this, because it's a one-line module and you can already write yourself, and better yet, publish on npm (there is probably existing ones).

3. I don't see the reason to avoid `isomorphic-fetch`, which does add a global fetch to your code.

jimmywarting:
I'm using node-fetch only on the server. So this is not something I would really want to be merged.
`node-fetch` is also not fully poly filled if a third party lib would think that it exist then it could fail cuz of it when it trying to do `res.blob()` which `node-fetch` or the node environment don't have - if you add node-fetch to the global window then it could break other applications because of this

If it's on the web then i use something else.
Fetch polyfill that use XmlHttpRequest and are more built for the browser, others frameworks ajax function like jQuery's ajax, angulars $http and so on, or just plain old XmlHttpRequest that has support for progress event (unlike most framework) 

Adding node-fetch to browser with webpack is a big overhead, you would need to bring in the stream and http modules as well when using it with webpack something that the browser don't have.
`node-fetch` would also need to bring more to the table if we are going to think about supporting browsers as well it would halt the development of going forward cuz we have to build browser specific hacks for IE...

tarikjn:
@TimothyGu thanks for your feedback, it's very helpful. The comparison would be more like an Englishman in England who is learning French and inviting a bunch of Frenchmen at his house for dinner who don't speak English (the `fetch` userland).

For Promise, there is actually a check if it is available as a global and then it is used if so. We can't and shouldn't always expect libraries that extend or depend on `fetch` to provide an option for the user to point to where the `fetch` lib is. Besides, doing this would be a leaking configuration abstraction.

And I agree with you on how to see it, default `notde-fetch` usage is a CommonJS style, but for users who want an emulation closer to the native fetch API and more library compatibility having that option would be useful. This is a pragmatic approach.

@bitinn 

1.  That's precisely the problem, we don't know if a module is designed for Node or the browser, most modules will simply depend on `fetch` and may actually just work on Node out of the box. e.g. `Frisbee`. Most of the time these modules should not even be concerned about their environment, it should be the concern of the user where to run them.
2.  I've thought about this but it seem it would be more user-friendly to provide/document this in the package. If a package on npm extend this, it will either have to follow all the releases or be a peer dependency. Either way it doesn't feel like a separate feature or an extension but more like a different loading option. It might make sense if it was an isomorphic option, but it's not.
3.  This doesn't have much to to with isomorphicity but more about how to correctly abstract and interface `fetch`

@jimmywarting you're confused, this is not for the browser. As for cases where libraries depending on fetch may not work, this is a different issue and is documented there in [LIMITS](https://github.com/bitinn/node-fetch/blob/master/LIMITS.md).

bitinn:
@tarikjn I am still on the *no* side, simply because `frisbee` decides to depend on a global fetch is not a reason for `node-fetch` to encourage a global fetch polyfill.

If you don't care about environment you are talking about *isomorphism*, whether you use universal javascript or other names doesn't really change this fact. 

`isomorphic-fetch` will always export the correct fetch module into global space, even if you don't use it on client-side, it's still a good option for Node environment.

tarikjn:
One other way to look at it is that defining on the global is a more flexible way of defining a peer dependency. It allows the user to bring-their-own-fetch. It doesn't mean an isomorphic approach, but it means there is a concern for portability, which is one of the main point of node.js.

Not only it's more flexible, I think this is the best practice for anything that want to work with fetch i.e. a library that extends `fetch` SHOULD assume it is on the global. If, say, it provided users with a way to pass in a reference to `fetch`, it would be a worse integration as it would have to be configured at every `require` -- unless if using some configuration injection, anyway those configuration points end up located in entry points that will not be located in the 'right' places. It's outside their concern because they are building on a standardized API, and eventual portability happens in the user environment.

You're basically trading off familiarity with the CommonJS approach of Node for a divergent API abstraction to fetch that will make it worse for integrations over time.

I do think your approach to using the underlying `node` APIs vs emulating `XHR` is very good approach thought, but that's on the 'back-end', the `fetch` integration is part of the implementation front-end (not referring to the browser folks!) where it should not necessarily follow the idiomatic Node approach.

bitinn:
Closing this as I have updated document to point users towards `isomorphic-fetch`, and if they don't want side-effect, `fetch-ponyfill`.

Issue #93, opened 2016-03-24
Error: incorrect header check


maticrivo:
I'm using node-fetch to fetch a URL that returns a JSON and I'm getting an error when parsing the response with `response.json()`.
The error I'm getting is: `Error: incorrect header check`
i curled the URL and there are no redirects or anything, this is the response header from curl:
```
HTTP/1.1 200 OK
Cache-Control: private
Content-Length: 57338
Content-Type: application/json; charset=utf-8
Server: Microsoft-IIS/8.5
X-AspNetMvc-Version: 5.2
X-AspNet-Version: 4.0.30319
Set-Cookie: WebApplication.Context=Culture=he-IL&UICulture=he-IL&AddressId=0&SearchCityId=0&SearchStreetId=0&SearchStreetAddress=&SearchStreetName=&SearchCityName=&SearchHouseNumber=&SearchCountryCode=&SearchFullAddress=&ShoppingCartGuid=00000000-0000-0000-0000-000000000000&WebsiteId=10bis&DomainId=10bis&BrowsingRestaurantId=0&SearchPhrase=&NavigationStartPoint=&HideAllPoups=False&IsKosher=False&DesiredDateAndTime=01/01/0001 00:00:00&CouponCode=&CompanyPreferredSite=&CompanyAdminPermissionLevel=&PreferredResList=List&PreferredResListSitting=Map; expires=Thu, 24-Mar-2016 13:11:08 GMT; path=/
X-Powered-By: ASP.NET
Date: Thu, 24 Mar 2016 09:11:08 GMT
```

using latest node-fetch version `node-fetch@1.4.1`

any idea how to fix this?

bitinn:
This error is coming from node.js zlib, similar to #45 (also a IIS server, if I recalled correctly)

But it has nothing to with http headers, instead, it's complaining about the header of response body (which would be in `gzip` or `deflate` compression in order to trigger this error).

If you must handle this page, option `compress: false` should cause server to send back a non-compressed version, which `node-fetch` should be able to read.

Note that `curl` might not send content-encoding header by default, so to properly test if that server is returning valid gzip/deflate. See this:

http://stackoverflow.com/questions/9140178/how-can-i-tell-if-my-server-is-serving-gzipped-content

bitinn:
closing as issue is from lower level

though with v1.5.0 you might be able to handle this error more gracefully (still, you need to correct the response to be able to extract with built-in zlib)

ankit-1729:
[...]

Issue #45, opened 2015-09-18
Unhandled error in deflate compression handling


pocketjoso:
Okay, I have to admit that I have no idea what's going on here, sorry for the vague issue.

Node `v0.12.7`
```
var fetch = require('node-fetch')
fetch('http://time-attendance.co.uk')
```

gives
```
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: incorrect header check
    at Zlib._handle.onerror (zlib.js:366:17)
```

...
The fetch results in a `ok` response... The promise is not rejected. But I still get this error thrown,  somewhere.. Confused. :/

pocketjoso:
Ah. Just curled that url. A _redirect_? Not supported in this library?

pocketjoso:
How do I handle this error? And is it possible to follow the redirect using `fetch-node`?

bitinn:
Yep, `node-fetch` follows redirect by default, so the problem is not the redirect itself, but somehow node.js zlib is not able to handle the `deflate` response this server is returning.

https://github.com/bitinn/node-fetch/blob/master/index.js#L141

As we do test for deflate response support, my guess would be that particular server is returning something weird.

https://github.com/bitinn/node-fetch/blob/master/test/server.js#L67-L74
https://github.com/bitinn/node-fetch/blob/master/test/test.js#L329-L338

As a workaround for now, setting options `compress: false` will cause the server to return decompressed content, allowing your fetch to go through.

Neverthless, I think in this case we need to handle this error so that `fetch` reject the result.

pocketjoso:
Thanks for your reply.

>  I think in this case we need to handle this error so that fetch reject the result

Yeah I think so too. As it is now I seem to have no way to handle the error, and I would prefer to throw here than having to change the default compression (I'm making many requests).

> to truly fix the problem require checking what went wrong during zlib inflate, maybe open a ticket on node.js core?

Not very read up on these things. How would I be able to tell what wen't wrong during the `inflate`? Would I not need to get onto the server serving that url?

bitinn:
@pocketjoso I plan to work on this module next week, will update when the fix has landed.

pocketjoso:
Awesome. :)

bitinn:
Fixed in v1.3.3, just note that we can't reject the fetch request (because the error is on body, which is a stream) but we do reject it when you call `res.text()` or `res.json()`.

Otherwise you can use `body.on('error', handler)` to it independently.

[...]

PR #123, opened 2016-05-30
Switch to inflateRaw when inflate header check fails

gzzhanghao:
See [chromium/filter/gzip_filter.cc:131](https://chromium.googlesource.com/chromium/src/net/+/master/filter/gzip_filter.cc#131)

> As noted in Mozilla implementation, some servers such as Apache with mod_deflate don't generate zlib headers.
> See 677409 for instances where this work around is needed.
> Insert a dummy zlib header and try again.

Possible fix for #93 and #45

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ var https = require('https');
 var zlib = require('zlib');
 var stream = require('stream');
 
+var Inflate = require('./lib/inflate');
 var Body = require('./lib/body');
 var Response = require('./lib/response');
 var Headers = require('./lib/headers');
@@ -174,7 +175,7 @@ function Fetch(url, opts) {
 					if (name == 'gzip' || name == 'x-gzip') {
 						body = body.pipe(zlib.createGunzip());
 					} else if (name == 'deflate' || name == 'x-deflate') {
-						body = body.pipe(zlib.createInflate());
+						body = body.pipe(new Inflate());
 					}
 				}
 			}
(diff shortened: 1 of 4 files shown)


bitinn:
Thx for the PR and an explanation to the mystery!

But this is my opinion:

- I really don't feel well workaround server problem.
- IF we must do so, it should be supported through a well-tested small npm module (is there already one?), we certainly don't want to maintain this ourselves.

bitinn:
On a second thought, why can't we detect zlib deflate from zlib deflateRaw? if we can, we only need to choose zlib inflate and inflateRaw and not add many lines of code.

bitinn:
Actually, if we want to be rough about this, try `zlib.inflate` and catch any error with `zlib.inflateRaw`, if none works, throw a custom error to inform users.

gzzhanghao:
@bitinn Any update on this?
I've pushed a commit to detect if we should use InflateRaw according to the first byte (not aligned to browsers, hopefully it works :P).
Also, I think we should align to browsers' behaviour AMAP, so I think this workaround is necessary for us.

bitinn:
@gzzhanghao I will keep this open for now, I would like to find a better solution then maintaining our own inflate lib (which is entirely a workaround).

bitinn:
This is now fixed in v1.6.0, I chose a less verbose method, but still thank you very much for pointing me in the right direction.

PR #132, opened 2016-07-12
default to 200 response status

wheresrhys:
As documented in the fetch spec https://fetch.spec.whatwg.org/#dom-response, the status should default to 200

```
dictionary ResponseInit {
  unsigned short status = 200;
  ByteString statusText = "OK";
  HeadersInit headers;
};
```

While the aim of node-fetch isn't to 100% match the spec, this seems like a reasonable bit of compliance to add.

Note - this PR implicitly sets status text to 'OK' too

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -23,7 +23,7 @@ function Response(body, opts) {
 	opts = opts || {};
 
 	this.url = opts.url;
-	this.status = opts.status;
+	this.status = opts.status || 200;
 	this.statusText = opts.statusText || http.STATUS_CODES[this.status];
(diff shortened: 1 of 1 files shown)


bitinn:
Yeah lgtm. Though it's interesting 200 is the default, I guess most requests are expected to be successful by default.

wheresrhys:
Thanks for merging. Is it reasonable to expect this to be published to npm soon?

bitinn:
Might have to wait for weekends, there are a few small fixes I would like to roll up with this release.

PR #135, opened 2016-07-21
add expectations for numeric status codes

wheresrhys:
This implements another bit of conformance with the `Response` spec. https://fetch.spec.whatwg.org/#dom-response.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -23,6 +23,11 @@ function Response(body, opts) {
 	opts = opts || {};
 
 	this.url = opts.url;
+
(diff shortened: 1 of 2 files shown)


bitinn:
I decide to close this PR, I know you are doing the right thing, but also note that `status` is not completely under developer control, as underlying Node.js `http.request` might not normalize the status code between 200-599...

It makes sense for us to throw RangeError on `new Response()` (programming error), but not quite right to reject with RangeError when response is outside of 200-599 range (not a programming error, which should use custom FetchError)...

Anyway. thx for pointing this out :)

PR #138, opened 2016-07-29
Update README.md form.getHeaders note

RusAlex:
getHeaders exists only in node environment. (if you run it server side). Browser FormData does not support this API.


bitinn:
I have updated the document to make sure developers notice the environment difference.

PR #140, opened 2016-08-03
Web streams (ReadableStream) support, blob & arrayBuffer accessors, ES6

gwicke:
- Expose `body` as WhatWG `ReadableStream`, in line with the spec & browser implementations.
- Remove charset detection, in line with the spec & browser implementations.
- Add missing blob & arrayBuffer accessors in the response.
- Migrate to a ES6 subset compatible with Node 4.
  - Use ES6 accessors to lazily construct a `ReadableStream` only when needed.
- Drop Node 0.10 & 0.12 support, and test Node 4 (LTS) & 6.


bitinn:
Thx for the PR, but as great as it might be, I won't be merging it due to multiple breaking change. I will however keep it open for future reference.

jimmywarting:
I think this would be a good starting point of v 2.0

lib/body.js
```diff
@@ -15,37 +15,37 @@ const webStreams = require('node-web-streams');
  */
 class Body {
     constructor(body, opts) {
-		opts = opts || {};
+        opts = opts || {};
```
dandv: [:+1:](https://medium.com/@hoffa/400-000-github-repositories-1-billion-files-14-terabytes-of-code-spaces-or-tabs-7cfe0b5dd7fd)  <-- click

lib/body.js
```diff
@@ -19,215 +13,190 @@ module.exports = Body;
...
+            // FormData instance. Make an exception, for now.
+            return rawBody;
+        } else {
+            throw new TypeError(`Unsupported Response body type: ${typeof body}`);
+        }
+    }
+
+    set body(val) {
+        this._rawBody = val;
+    }
```
TimothyGu: This goes against the WHATWG spec, which specifies that

```webidl
readonly attribute ReadableStream? body;
```

The `readonly` annotation means that the setter should be left blank.

lib/body.js
```diff
@@ -19,215 +13,190 @@ module.exports = Body;
...
+        return this._consumeBody()
+            // Convert to ArrayBuffer
+            .then(body => body.buffer.slice(body.byteOffset,
+                        body.byteOffset + body.byteLength));
+    }
+
+    /**
+     * Accumulate the body & return a Buffer.
+     *
+     * @return  Promise
```
TimothyGu: Make clear what the type of the encapsulated data in the Promise is as well.

lib/body.js
```diff
@@ -19,215 +13,190 @@ module.exports = Body;
...
-		instance.body = p1;
-		body = p2;
-	}
-
-	return body;
+class Body {
+    constructor(body, opts) {
+        opts = opts || {};
+        this._rawBody = body;
+        this.bodyUsed = false;
```
TimothyGu: Since the Fetch spec declares `bodyUsed` as `readonly` you might want to do the getter/setter dance similar to `rawBody` and `body`.

lib/body.js
```diff
@@ -19,215 +13,190 @@ module.exports = Body;
...
-		// set instance body to teed body and return the other teed body
-		instance.body = p1;
-		body = p2;
-	}
-
-	return body;
+class Body {
+    constructor(body, opts) {
+        opts = opts || {};
+        this._rawBody = body;
```
TimothyGu: You could make `rawBody` a Symbol

gwicke:
This PR is pretty old btw. Current work is in the master branch of https://github.com/gwicke/node-fetch-polyfill.

Issue #139, opened 2016-08-02
Unused clone causes stream back-pressure


ThaNarie:
**issue**
Cloning a stream, which you end up not using, causes back-pressure on the original response, preventing it from draining the source.

```js
fetch(url).then(function(res) {

	// unused clone causing back-pressure in the stream
	var clone = res.clone();

	// this will start reading internally, but just 'halts' mid-way and never completes
	res.json().then(function(result) {
		console.log(result);
	});
})
```
re 419597fe131cec0552d5d907f145032df485a2d6

**background**
The reason why this happens is perfectly explained here:
http://www.bennadel.com/blog/2817-the-affect-of-back-pressure-when-piping-data-into-multiple-writable-streams-in-node-js.htm

**use case**
In my project I have some checking in the `then` (status code, etc), and when something 'fails' (or the `res.json()` throws an error because the response is not json for some reason), I want to the error to end up in the chained `catch` handler.
In the `catch` handler I wanted to read the response as `.text()`. When the `.json()` was already used I couldn't do that, so I needed a clone.

So what I ended up doing, was to create a clone before reading the json (scoping it to the 'parent' function, and using that clone to read the `.text()` response in the `catch`.
This means that the clone was only used when an error occurred.

**internal logging**
I ended up adding logs in the `"data"` (https://github.com/bitinn/node-fetch/blob/952aa2a13546532ff612bfa98ec8216dda8a3871/lib/body.js#L105) and `"end"` handlers of the body stream, and if the size was longer than `32768` (in my test case that was 2 chunks), it stopped there, without calling `"end"`.

When, after a timeout, you start reading from the clone, both the clone and original continued until all data was drained.

There seems to be a difference between normal and gzipped requests, normal requests can be larger before the issue occurs.

**workarounds**
Maybe first ask the question if we want to solve this?
Cloning things without using them seems something you don't want to do, but if you do need it later, you can only clone it before you've read it the first time. And I don't think the code should silently 'fail' if for some reason you won't end up using the clone.

There are workarounds, what I've done now is to read the response as `.text()`, and parse that to json manually, so I can use the text response later on _if_ I need it.
But that would become a bit cumbersome when having to do that for blobs, arrayBuffers and formData as well.

**how to solve?**
I'm new to streams, so I'm not sure how to correctly solve this.

[...]

PR #142, opened 2016-08-03
Fix back-pressure issues when cloning responses

ThaNarie:
Add tests showing that cloned responses will cause back-pressure

When forking streams, if one of the streams doesn't read, and its
internal buffer is full, it will stop all streams to pause.

The added tests will now fail because of the 2 second timeout, but
without the timeout they will pause indefinitely.
They are copies of the test above them, but with a larger response.

re #139

(Note: this are only the tests, not the actual fixes)


bitinn:
I will take some times to think about how to solve it, probably with a slightly more clever PassThrough stream (one that simply drain itself when there is no downstream pipes)...

bitinn:
Since there isn't a good solution to this problem, we decide to put it in the [KNOWN LIMIT document](https://github.com/bitinn/node-fetch/blob/master/LIMITS.md).

See also #151 on discussion about buffer size.

Issue #112, opened 2016-05-01
headers not getting sent


omnidan:
I have the following call:

```js
fetch('https://url:port/status',
  {
    headers: {
      'X-Token': 'XXX'
    }
  }
)
```

Which results in `401 Unauthorized`. However, the same call with `curl` works fine:

```
curl -H "X-Token: XXX" https://url:port/status
```

It seems like the headers are not getting sent with `node-fetch`.

bitinn:
We don't limit headers at all, do a simple test with http://httpbin.org/ or make a failing test cases would help.

omnidan:
@bitinn thanks for the suggestions - httpbin.org is very useful for debugging :grin:

This is very weird though - I got the requests to look exactly the same on httpbin, but it's still not working with `node-fetch` while it works fine with `curl`. Also, I'm trying to access the plex api (e.g. `/status/sessions`).

Any idea how I could further debug this?

bitinn:
hmmm, I can only imagine they demand a nicer user agent? check their robots.txt for example?

you might want to figure out if they provide more detailed error message somewhere? Seem more like a Plex issue to me than Node.js issue.

omnidan:
@bitinn actually I did set the `User-Agent` header to `curl/7.43.0` too so the requests are exactly the same.

Yeah I thought so too, but if the requests are exactly the same, why does `curl` work and `node-fetch` doesn't - very weird.

bitinn:
You might want to raise it to Plex people (if your requests are exactly the same, then either your network or plex is acting up), I will keep this open.

mmocny:
FYI it looks like node-fetch lowercases all headers instead of respecting case provides.  According to HTTP spec, they are supposed to be case-insensitive, so this should be fine, but some servers do this incorrectly.

We also just had an issue with a server which was doing case-sensitive lookup and we were at a loss why the headers were not being passed down, until we did a raw dump and realized the casing issue.

bitinn:
hmmm it could be the cause to OP's problem, httpbin.org might not output headers according to their original casing.

PS: not lowercasing within `node-fetch` might lead to user setting duplicate headers unknowingly, if we do change this behavior to allow uppercase, it would probably be a breaking change of some sort.

bitinn:
Ah Fetch Spec did suggest lower-casing header names https://fetch.spec.whatwg.org/#terminology-headers

bitinn:
there isn't much we can do for this issue, closing for now.

[...]

PR #144, opened 2016-08-09
Change output header key to be same as input string

horiuchi:
In now implementation, all output header key convert to lowercase, when request server.
I'd prefer to output the original string for request header.

And I'm sure that the header key is case insensitive.

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -5,6 +5,8 @@
  * Headers class offers convenient helpers
  */
 
+var caseless = require('caseless');
+
 module.exports = Headers;
 
 /**
@@ -15,8 +17,8 @@ module.exports = Headers;
  */
 function Headers(headers) {
 
-	var self = this;
 	this._headers = {};
+	this._dict = caseless(this._headers);
 
 	// Headers
 	if (headers instanceof Headers) {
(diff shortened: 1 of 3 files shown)


jimmywarting:
From [RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1"][1], [Section 4.2, "Message Headers"][2]:

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

The updating  RFC 7230 does not list any [changes from RFC 2616](https://tools.ietf.org/html/rfc7230#appendix-A.2) at this part.


  [1]: http://www.w3.org/Protocols/rfc2616/rfc2616.html
  [2]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

---

If the browser version don't change the header key, then we shouldn't do that either.

bitinn:
Sorry but the Fetch Spec requires us to use lowercase naming: https://fetch.spec.whatwg.org/#terminology-headers

See similar discussion at #112

> To set a name/value (name/value) pair in a header list (list), run these steps:
> 
> - **Byte-lowercase** name.
> 
> - If there are any headers in list whose name is name, set the value of the first such header to value and remove the others.
> 
> - Otherwise, append a new header whose name is name and value is value, to list.

horiuchi:
@bitinn OK, I see. It's by the spec.

PR #149, opened 2016-08-16
[WIP] piping gunzip to a passthrough requires a call to resume?

Peleg:
This is very much a work in progress. I'm only half way sure of what's going on here and would love to hear others thoughts.

The issue is this:

Calling `clone` on a response with gzip encoding and multiple data chunks seems to miss some of `data` events and the `end` event which deems the response useless to read.

Trying to read from the original body immediately after the clone seems to solve the issue but I'm confused why as it's just another instance of PassThrough.

From what I can tell, this seems to be an issue only when the stream is Gunzip.

PS: maybe this is related? https://nodejs.org/api/stream.html#stream_additional_notes

### DIFF
--- a/lib/body.js
+++ b/lib/body.js
@@ -8,6 +8,7 @@
 var convert = require('encoding').convert;
 var bodyStream = require('is-stream');
 var PassThrough = require('stream').PassThrough;
+var Gunzip = require('zlib').Gunzip;
 var FetchError = require('./fetch-error');
 
 module.exports = Body;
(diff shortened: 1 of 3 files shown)


bitinn:
The question being, does the stream end itself? It's not back-pressure like #142?

I see no reason for chunks to go missing, but it's possible your stream never ends due to one of the pass through stream internal buffer being full. A safe workaround at the moment is to avoid doing that to large response (ie. make sure you pipe both stream, not one after another).

Peleg:
@bitinn you are right. it was back-pressure related. hopefully #154 fixes this issue.

Issue #151, opened 2016-08-18
Usage of clone stream and backpressure (aka default highWaterMark value)


bitinn:
It seems browsers (at least Chrome) are smart enough to _duplicate_ the response stream if users:
1. `clone` the response.
2. resolve the original response completely.
3. and then resolve the cloned response.

Per Fetch Spec's [clone definition](https://fetch.spec.whatwg.org/#concept-body-clone), we should `tee` the stream at step 1, and `tee` is defined in [Stream Spec](https://streams.spec.whatwg.org/#rs-tee). `node-fetch` does this already.

Problem: **we don't know the response body size beforehand**. So when user `clone`, they still risk running into backpressure issue; as on step 2, when data starts flowing for both the original and the cloned stream, and because we are not reading from cloned stream at the time, its internal buffer will fill up, eventually, causing both stream to pause and hence what we called _backpressure_.

It seems Chrome has a higher default buffer size than Node.js, so when users approach this problem with isomorphism, they run into issues where request works on client-side, but failed on server-side.

There are on-going discussion at https://github.com/whatwg/streams/issues/401 and I am asking this again just to be sure https://github.com/whatwg/streams/issues/506

thebigredgeek:
Is this fixed?  Ran into an issue in production where request promises aren't resolving or rejecting on the server due to the use of `clone` in an isomorphic frontend using `react` and `apollo-client`.  Would love to see a fix for this problem

bitinn:
The last time I checked, there is no easy fix for this as it's by design. We can keep increasing the internal stream bucket size but eventually backpressure will kicks in. Users need to be aware of backpressure when working with streams, unfortunately.

(From my phone)

DevBrent:
> Users need to be aware of backpressure when working with streams, unfortunately.

@bitinn I don't know if this is an answer really. I'm fully aware of backpressure due to my use of node-fetch but response.clone()'s implementation makes it physically impossible to fetch large bodies with no known resolution but not to use clone().

Is this an issue where every response copy must be fully read in order for any to read beyond the buffer or is it an issue where it becomes impossible to read beyond the buffer once clone is used at all?

bitinn:
@DevBrent the real problem here is: there is no standard on just how much a consumer of a stream should be willing to hold before it causes back-pressure.

Nodejs has a very small cap, browsers have a large and inconsistent cap. The only safe way is to make sure you don't cause the stream to back-pressure: by giving it some direction to flow to.

[...]

PR #154, opened 2016-08-19
Defer decompression to avoid back-pressure when cloning a response

Peleg:
This PR builds on top of #142 and hopefully fixes the issue where cloning a response and not reading one of the streams causes back-pressure and pauses flowing.

The issue seemed to occur only when dealing with responses which are large Transform streams (Gunzip/Inflate) which would pipe down more than they read after they transform their chunks (see: https://github.com/nodejs/node/blob/v5.5.0/lib/_stream_transform.js#L28-L37 for more info).

By deferring the decompression step until the response is being decoded, we can avoid having this issue and maintain "predictable" chunk sizes when cloning a response.

This, of course, adds the overhead of having to decompress the response each time it's being cloned.


Peleg:
not sure about the test failure. they pass when run locally, even when I set the highWaterMark on the clones to be super low. is it possible it needs a longer timeout?

bitinn:
I will have to look into it later, but I don't think the backpressure problem can be fully resolved given #151 (see linked discussion as well)

Peleg:
Yup. I finally managed to reproduce this when cloning a (much) larger response that isnt gzipped. This PR still seems to help when the response is compressed as it pipes a much smaller stream (gzipped vs raw).

Back to the drawing board 😒

bitinn:
Do let me know if you have come up with anything, due to lack of a good solution, we have added it to [KNOWN LIMIT document](https://github.com/bitinn/node-fetch/blob/master/LIMITS.md).

Peleg:
thanks @bitinn. in the meantime, should there be an option to increase the PassThroughs high water mark so we dont encounter back pressure when cloning a response with body > 16KB?

bitinn:
I don't plan to expose `highWaterMark`, here are my reasons:

- If you know you are running into this problem, the better solution is `Promise.all()`. Run your stream in parallel, don't resolve it one-after-another.
- Setting `highWaterMark` manually is not a good workaround. A good workaround is to put the stream in memory or on file. Basically, avoid `clone()` when you just need to *copy* your data.

Peleg:
- my use case for having access to the hwm is a library which consumes the response's body but doesn't want to interfere with the enduser's ability to read it as well. my library just clones the response before reading it so it can be read again by the enduser. whether it is actually read or not, is not within my control.

- wouldnt setting a higher hwm essentially mean storing the stream in memory?

bitinn:
One thing you can do is to wrap `res.body` (which is a Transform stream) with a stream of larger buffer. `text()`, `json()` will still happily consume `res.body` as long as they are a valid readable stream. End-user wouldn't have to know.

Setting a higher `highWaterMark` doesn't give you the actual control of the data itself. In most cases where you set `highWaterMark`, you want the full data anyway, so why not just buffer it.

TL;DR: your users will have to learn back pressure at some point and setting `highWaterMark` is just a workaround, never a solution. I am sorry but I have to put it bluntly.

PR #162, opened 2016-09-10
Fix for https://github.com/bitinn/node-fetch/issues/161

GalenWarren:
See https://github.com/bitinn/node-fetch/issues/161


bitinn:
See my response at the issue.

PR #166, opened 2016-09-23
Return empty .json() object on 204. Fix #165.

### LINKED ISSUE #165: Return empty object on 204 No Content...
...instead of throwing a syntax error due to attempting to JSON parse `''`.

```
let fetch = require('node-fetch');
fetch('http://httpbin.org/status/204').then(res => res.json()).then(object => {
    console.log(object);
}).catch(error => console.error(error));
```

Demo: http://runkit.com/dandv/node-fetch-204-no-content

Fixing this would attain compatibility with [request](https://github.com/request/request/issues/2177) and [got](https://github.com/sindresorhus/got/pull/78).

More generally, it would be more tolerant to return an empty object if the response is the empty string (regardless of the status code being 204 or other 2xx).

dandv:
+ some cosmetic whitespace fixes

### DIFF
--- a/lib/body.js
+++ b/lib/body.js
@@ -40,7 +40,8 @@ function Body(body, opts) {
 Body.prototype.json = function() {
 
 	return this._decode().then(function(buffer) {
-		return JSON.parse(buffer.toString());
+		var string = buffer.toString();
+		return string === ''? {} : JSON.parse(string);
 	});
 
 };
--- a/lib/body.js
+++ b/lib/body.js
@@ -39,9 +39,8 @@ function Body(body, opts) {
  */
 Body.prototype.json = function() {
 
-	
 	if (this.status === 204) return Body.Promise.resolve({});
-        return this._decode().then(function(buffer) {
+	return this._decode().then(function(buffer) {
 		return JSON.parse(buffer.toString());
 	});
 
--- a/lib/body.js
+++ b/lib/body.js
@@ -39,9 +39,10 @@ function Body(body, opts) {
  */
 Body.prototype.json = function() {
 
-	return this._decode().then(function(buffer) {
-		var string = buffer.toString();
-		return string === ''? {} : JSON.parse(string);
+	
+	if (this.status === 204) return Body.Promise.resolve({});
+        return this._decode().then(function(buffer) {
+		return JSON.parse(buffer.toString());
 	});
 
 };
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "node-fetch",
-  "version": "1.6.1",
+  "version": "1.6.2",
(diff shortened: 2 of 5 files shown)


bitinn:
I agree 204 empty response should be handled, but I think checking `this.status == 204` is a better solution.

I don't think it's `node-fetch`'s responsibility to handle json() on empty string. Or at least Fetch Spec doesn't seem to give any special treatment for it.

https://fetch.spec.whatwg.org/#dom-body-bodyused

dandv:
> I think checking this.status == 204 is a better solution

I thought so too, but opted to check for the actual empty string, regardless of the response code, because that's more resilient and the actual source of the problem. Is there a use case for failing with a syntax error on a 200 response that's empty because the implementor of the service didn't return 204?

bitinn:
> because that's more resilient and the actual source of the problem

I agree to some degree, for maximum compatibility, just checking empty string makes it work better across different scenarios.

But as mentioned in `got`, returning a 200 empty response when users are requesting JSON, is simply invalid, and we are right to throw an error.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "node-fetch",
-  "version": "1.6.1",
+  "version": "1.6.2",
```
bitinn: Could you kindly revert this change, we actually bump version using `npm version` right before publish.

lib/body.js
```diff
@@ -39,7 +39,9 @@ function Body(body, opts) {
  */
 Body.prototype.json = function() {
 
-	return this._decode().then(function(buffer) {
+	
+	if (this.status === 204) return Body.Promise.resolve({});
+        return this._decode().then(function(buffer) {
```
bitinn: Please format this line as tab.

test/test.js
```diff
@@ -1402,7 +1413,8 @@ describe('node-fetch', function() {
 		return req.buffer().then(function(result) {
 			expect(result.toString()).to.equal('a=1');
 		});
-	}); 
+	});
+
```
bitinn: And I have no idea why these changes are needed?

dandv: Those were trailing whitespace automatically removed by my editor.

bitinn: but it also adds an additional newline.

package.json
dandv: Fixed. I thought GitHub would hide the discussion now due to it being on an outdated diff, but I still see it.

test/test.js
dandv: I might have to file a bug with the Kate editor about that... :-/

lib/body.js
dandv: Done.

bitinn:
released in 1.6.2

PR #173, opened 2016-10-03
Add 'post javascript object' example to readme

chrisveness:
Wasn't sure initially if `JSON.stringify()` for post body would work, then it took me longer than it should have to work out that `Content-Type` was needed for POST JSON to work with koa-body (and perhaps other body parsers?), so this might help others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,19 @@ fetch('http://httpbin.org/post', { method: 'POST', body: stream })
 		console.log(json);
 	});
 
+// post with JavaScript object
+
+var body = { a: 1 };
+fetch('http://httpbin.org/post', { 
+        method: 'POST',
+        body:    JSON.stringify(body),
+        headers: { 'Content-Type': 'application/json' },
+    }).then(function(res) {
+        return res.json();
+    }).then(function(json) {
+        console.log(json);
+    });
+
 // post with form-data (detect multipart)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,6 +125,19 @@ fetch('http://httpbin.org/post', { method: 'POST', body: stream })
...
+var body = { a: 1 };
+fetch('http://httpbin.org/post', { 
+        method: 'POST',
+        body:    JSON.stringify(body),
+        headers: { 'Content-Type': 'application/json' },
+    }).then(function(res) {
+        return res.json();
+    }).then(function(json) {
+        console.log(json);
+    });
```
TimothyGu: I think the rest of the file uses tabs instead of spaces

chrisveness: Hadn't noticed that, I never use tabs! Should I commit a revision which can be squashed when PR is merged, or should I squash commits (or just scratch it and start over)?

bitinn: I will squash it, you can just fix the formatting.

Though I do wonder how often people use this when `form-data` is available.

chrisveness: Thanks for giving us `fetch` on Node, by the way – I've dispensed with jQuery `$.ajax` client-side, and now can use the same api server-side :)

TimothyGu:
Thanks!

PR #174, opened 2016-10-03
Implement iterator on header

soyuka:
Node 6+ obviously

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -139,3 +139,68 @@ Headers.prototype['delete'] = function(name) {
 Headers.prototype.raw = function() {
 	return this._headers;
 };
+
+/**
+ * Get an Iterator on keys
+ * @return Iterator
+ */
+Headers.prototype.keys = function() {
+	return new HeadersIterator(Object.getOwnPropertyNames(this._headers));
+};
+
+/**
+ * Get an Iterator on values
+ * @return Iterator
+ */
+Headers.prototype.values = function() {
+	var values = []
+
+	this.forEach(function(value) {
+		values.push(value);
+	});
+
+	return new HeadersIterator(values)
+};
+
+/**
+ * Get an Iterator on entries, same as iterating over the Header instance
+ * @return Iterator
+ */
+Headers.prototype.entries = function() {
+  var values = []
+
+	this.forEach(function(value, name) {
+		values.push([name, value]);
+	});
+
+	return new HeadersIterator(values)
+}
+
+Headers.prototype[Symbol.iterator] = function() {
+	var values = []
+
+	this.forEach(function(value, name) {
+		values.push([name, value]);
+	});
+
+	return new HeadersIterator(values);
+};
+
+function HeadersIterator(values) {
+	this._values = values;
+	this._index = 0;
+};
+
+HeadersIterator.prototype = {};
+
+HeadersIterator.prototype.next = function() {
+	if (this._index >= this._values.length) {
+		return {
+			value: undefined, done: true
+		};
+	}
+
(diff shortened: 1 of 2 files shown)


TimothyGu:
Would it be easier if you implement `values`, `keys`, and `entries` with generator functions?

Either way, I don't think the current output of `values` and `keys` are quite correct, since they are missing a `next` function. With a generator function, you won't need to worry about such details.

Another question is if we should force the user to use Node.js v6+, or we should transpile the code using Babel.

bitinn:
@TimothyGu would you like to be a contributor on this repo and drive this feature? I don't use ES6 feature that much so I haven't had time to investigate.

bitinn:
@TimothyGu Oh and for the second question, my big concern is Node v4 LTS really. We can safely drop v0.10 and v0.12 by the end of this year (bump to `node-fetch` v2.0 of course).

soyuka:
> Would it be easier if you implement values, keys, and entries with generator functions?

The result is the same IMO, but maybe it's cleaner using generators. I have a need to loop with `for (let v of)` on Headers, so I updated the polyfill without really thinking about the best implementation.

> Either way, I don't think the current output of values and keys are quite correct, since they are missing a next function. With a generator function, you won't need to worry about such details.

Yeah, I see what you say here, but nodejs wanted me to declare those on the `[Symbol.iterator]` key. By the way, tests are valid and pass, so IMO it's working as it should. Do you have another test in mind?

> Another question is if we should force the user to use Node.js v6+, or we should transpile the code using Babel.

Transpiling is a good idea IMO, babel would work for Iterators and Generator functions btw. 

> @TimothyGu Oh and for the second question, my big concern is Node v4 LTS really. We can safely drop v0.10 and v0.12 by the end of this year (bump to node-fetch v2.0 of course).

In pm2 we're waiting to drop the support for `v0.10` too... And yeah, the end of this year would be the period of choice. Anyway, to keep supporting node LTS we have to add transpiling. 

@TimothyGu let me know if you need some work on the es6/babel implementation

TimothyGu:
@bitinn, yes please. I'll help whenever I can.

@soyuka,
>> Would it be easier if you implement values, keys, and entries with generator functions?
>
> The result is the same IMO, but maybe it's cleaner using generators. I have a need to loop with `for (let v of)` on Headers, so I updated the polyfill without really thinking about the best implementation.

Fair.

>> Either way, I don't think the current output of values and keys are quite correct, since they are missing a next function. With a generator function, you won't need to worry about such details.
>
> Yeah, I see what you say here, but nodejs wanted me to declare those on the `[Symbol.iterator]` key. By the way, tests are valid and pass, so IMO it's working as it should. Do you have another test in mind?

I see for `HeadersIterator` you've done it the right way: having a `next()` and a `@@iterator()` that returns `this`. It just should be the same case for `values` and `keys` as well.

soyuka:
Ok, I'll improve this. Should I try including babel in this PR?

TimothyGu:
No, don't worry about it. I have a PR incoming for that.

bitinn:
@TimothyGu great you are now a collaborator.

So it looks like we do need a transpiler to get iterator working on Node v4. So be it :)

soyuka:
Only node `0.10` doesn't pass. I had a `let` in the code which was breaking things, actually `0.12` seems to support iterators.

jimmywarting:
https://www.npmjs.com/package/fetch-headers

TimothyGu:
I've implemented #180, which incorporates elements of this pull request but also converts the entirety of Headers to ES2015. Please review.

Issue #175, opened 2016-10-04
WHATWG-URL support


stevenvachon:
[whatwg-url](https://npmjs.com/whatwg-url) is better than core `url`.
```js
fetch( new URL("http://google.com/") )
```

bitinn:
It certainly looks promising. But does it solve the "file:///URL" problem?

Because it can open up to quite a lot of nasty things, so much that Fetch Spec decides to leave it out of the spec. Which is also why we opt to leave it alone in #75 .

For I am quite fine with default `url` parser. Did you run into any problems?

ref: https://fetch.spec.whatwg.org/#basic-fetch

stevenvachon:
I'm not aware of the local file problem. Please tell me.

Aside from that, however, core `url.parse` will still parse such URLs, so I don't see a difference there.

TimothyGu:
The support for this could be pretty easily added. The spec for the [Request class](https://fetch.spec.whatwg.org/#request-class) states that the first argument to `new Request()` must be either another `Request` or a `USVString`; WHATWG's URL class when coerced to a string returns the URL in string format.

TimothyGu:
Fixed in v2 branch.

PR #177, opened 2016-10-04
Work in progress: node-fetch v2

TimothyGu:
## DO NOT MERGE

This PR will be continuously updated as more updates roll in. Feel free to initiate pull requests against this branch to help with this process.
## Work needed to be done
- [x] Use Babel
- [x] Add iterator interface to Headers #174
- [x] Convert all files to modern JavaScript syntax
- Split and merge #140 
  - ~~WHATWG streams support~~
  - [x] Text encoding conversion #184 #188
  - [ ] Determine behavior of `.blob()` 
- [x] Resolve Headers incompatibility #181 #185
- [ ] Implement a few missing methods such as `Response.error` and `Response.redirect`
- [x] Fix spec compliance for objects as URL #175 #187
## Discussion
- **Code style:** it seems to be preferred that spaces are used instead of tabs for most people (myself included), so should we use that instead? **No.**


soyuka:
> Convert all classes to use the class keyword

What's the advantage of doing this? It's perfectly valid to use prototypes with es6.

> Code style: it seems to be preferred that spaces are used instead of tabs for most people (myself included), so should we use that instead?

Oh gosh +1

bitinn:
> Code style: it seems to be preferred that spaces are used instead of tabs for most people (myself included), so should we use that instead?

Tab is just my preference, if Node.js community like spaces, I say why not.

> Can you add use const/let instead of var + use strict?

Are they worth it for the incompatibility they bring? If you use Babel, you end up with a polyfill anyway?

TimothyGu:
> It's perfectly valid to use prototypes with es6.

Yes, but ES6 classes are a bit easier to read in my opinion.

jimmywarting:
Babel transpile what you want it to transpile. That's the good thing about modular

bitinn:
I have a second thought on code style: it might actually be better to keep tabs just because we can see the actual changes in files (formatting them to spaces and we lost ability to view *blame* in github).

If we are making breaking changes might as well make it easy for others to troubleshoot breaking changes.

TimothyGu:
Sounds good to me.

bitinn:
On item *WHATWG streams support*:

Do we want WHATWG stream on Node.js? Or do we just want similar API so people can write isomorphic code (`res.body.getReader` etc.)?

I still don't know how Node.js Streams and Web Streams are going to consolidate, and then there is adoption, do we start to see modules that accept both type of streams?

I think it's safe to assume as of now, if a developer opt to use Web Streams on Node.js, they probably need to patch/rewrite existing modules from ground up to avoid conversion between streams.

TimothyGu:
@bitinn said:
> Do we want WHATWG stream on Node.js? Or do we just want similar API so people can write isomorphic code (`res.body.getReader` etc.)?

Indeed, I think most people wanting to use WHATWG streams probably do so for the isomorphism provided by them. WHATWG streams can be easier to use than Node.js streams. How about we just use the same approach I took for `textConvert()`? Make `body` a getter for the WHATWG stream, while adding a `rawBody` getter that maintains the current behavior for `body`?

bitinn:
> Make body a getter for the WHATWG stream, while adding a rawBody getter that maintains the current behavior for body?

I think it's worth keeping `res.body.pipe` compatible with Node.js stream, 2 arguments:

- As far as I know, WHATWG stream will avoid `pipe()` method, so there won't be any confusion, if you use `pipe()`, you want Node.js streams.

- It will break many codebases, I consider `text()` with non-UTF encoding support a fairly minor use cases (important only if you use `node-fetch` as a crawler). But removing `pipe()` from body may cause many users to not upgrade.

So I say we keep `res.body.pipe` even though it's not in spec.

> WHATWG streams can be easier to use than Node.js streams

I haven't seen a popular WHATWG streams implementation, so people really use it on Node.js?

LinusU:
I saw that #185 was merged behind a global flag. I think that this is very bad for a number of reasons. First of all, I think that following the spec should be the behavior that we want. Secondly, it totally breaks your expectations when requiring the library. If you depend on two libraries which both uses node-fetch, one doesn't set `FOLLOW_SPEC` and the other one sets it to `true`. Then the behavior will be modified for the one library that didn't set any flags.

What would be the downside to having the Headers follow the spec be the only available behavior?

ping @TimothyGu

LinusU:
So I found #181 and I now understand why one would maybe not want to follow the spec exactly at this point 😞 

My second point still stands though, since that global flag is shared between all users of the library, I really think that it should be solved in another way...

jimmywarting:
Just needed `Response.redirect` today... Almost created a new issue

UPGRADE-GUIDE.md
```diff
@@ -0,0 +1,82 @@
+# Upgrade to node-fetch v2
+
+node-fetch v2 brings about many changes that increase the compliance of
+WHATWG's Fetch Standard. However, many of these changes meant that apps written
+for node-fetch v1 needs to be updated to work with node-fetch v2 and be
+conformant with the Fetch Standard.
+
+## `.text()` no longer tries to detect encoding
+
+Currently, `response.text()` attempts to guess the text encoding of the input
```
bitinn: A slight nitpick, can you replace both instance of *currently* with *in v1.x*

PR #179, opened 2016-10-08
json parse fix

vahpetr:
hotfix. need research how work origin fetch method json if server return text


bitinn:
Thx for the PR. But we have many discussion on this already, the answer remains: Fetch doesn't handle cases where server returns invalid response (if you got text when you are asking for json, it's invalid response).

PR #180, opened 2016-10-08
Convert Headers to ES2015 and implement Iterable interface

### LINKED ISSUE #127: Incompatible API in Headers class
Hi,

I use [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) in my project with [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch). Unfortunately there are some incompatibilities between [Headers API](https://developer.mozilla.org/en-US/docs/Web/API/Headers) and https://github.com/bitinn/node-fetch/blob/master/lib/headers.js.

Missing methods:
- entries
- keys
- values

TimothyGu:
Closes #127, #174.

The approach taken by this pull request versus #174 is that the iterator methods are greatly simplified as well as more correct w.r.t. the behavior of `keys()` in case of multiple values.

### DIFF
--- a/src/headers.js
+++ b/src/headers.js
@@ -179,6 +179,13 @@ export default class Headers {
 	[Symbol.iterator]() {
 		return this.entries();
 	}
+
+	/**
+	 * Tag used by `Object.prototype.toString()`.
+	 */
+	get [Symbol.toStringTag]() {
+		return 'Headers';
+	}
 }
 
 const ITEMS = Symbol('items');
--- a/src/headers.js
+++ b/src/headers.js
@@ -5,137 +5,132 @@
  * Headers class offers convenient helpers
  */
 
-module.exports = Headers;
+export default class Headers {
+	/**
+	 * Headers class
+	 *
+	 * @param   Object  headers  Response headers
+	 * @return  Void
+	 */
+	constructor(headers) {
+		this._headers = {};
 
-/**
- * Headers class
- *
- * @param   Object  headers  Response headers
- * @return  Void
- */
-function Headers(headers) {
-
-	var self = this;
-	this._headers = {};
-
-	// Headers
-	if (headers instanceof Headers) {
-		headers = headers.raw();
-	}
-
-	// plain object
-	for (var prop in headers) {
-		if (!headers.hasOwnProperty(prop)) {
-			continue;
+		// Headers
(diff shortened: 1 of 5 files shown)


jimmywarting:
Another tip would just be to include https://www.npmjs.com/package/fetch-headers as a dependency

TimothyGu:
@jimmywarting, in general I don't see why we need to add a dependency when we could just copy that into this repo.

W.r.t. the code, we still maintain compatibility with Node.js versions as old as 0.10, which your package doesn't seem to do. I'm also not sure about why you used WeakMap.

W.r.t. the additional checks of `normalizeName` and `normalizeValue`, I will add them in a future pull request. There are certain API differences between GitHub's Fetch and the Fetch Standard that we need to discuss, before proceeding with changing it.

W.r.t. `Symbol.toStringTag`, good suggestion. I've incorporated that in this PR as well.

jimmywarting:
I just think it's a good idea to factor out some of the logic so it can be used by others as well.
dose keeping this package smaller and cleaner

WeakMap is 100% private and Symbol('map') isn't since you can always get the reference.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertySymbols

jimmywarting:
node v0.10 don't support class so...?

TimothyGu:
We compile the sources with Babel now.

Okay, so WeakMap is in reality more private than Symbols, but with Symbols you also get the advantage of being able to introspect the underlying data structure. So it's like "whatever you want to do you can't touch the map" versus "we've warned you that this object is internal, but if you _really really want_ to see what's in there you can, just don't overdo it." I think I prefer the latter's flexibility.

Symbols are also the approach used by some other projects like Babel.

jimmywarting:
Sure, if we use babel then i don't see any problem with my lib being included in the build process

TimothyGu:
Also CC @soyuka.

bitinn:
@TimothyGu Just a quick remainder: while code coverage remains 100%, maybe it's worth double-checking if all newly added features are tested.

Also a good practice: keep each test short and concise if possible, I haven't read the test throughly, maybe it can be refactored somehow. It's more verbose for sure, but in testing it maybe better to have multiple small tests for each edge cases/features instead of a large one.

PR #182, opened 2016-10-09
Convert all files to ES2015

TimothyGu:
Elements of this commit come from #140 by @gwicke.


TimothyGu:
Unfortunately the test suite fails on 0.12, but for some reason passes on 0.10. I'm able to reproduce this locally, but unable to discover a reason for it. Maybe someone else could take a look?

PR #183, opened 2016-10-10
Miscellaneous improvements over #182

TimothyGu:
* Simplify `Request` and `Body` classes
* Add support for `body.arrayBuffer()` (using [buffer-to-arraybuffer](https://www.npmjs.com/package/buffer-to-arraybuffer) to gain support on older Node.js compared to #140)
* Fix spec compliance when body is `null`
* Add more tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "babel-runtime": "^6.11.6",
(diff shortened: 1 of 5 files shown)


mythz:
Hi do you have a rough idea for when v2 with `arrayBuffer()` will be available? thx.

PR #188, opened 2016-10-13
Add a new res.textConverted() and always use UTF-8 for res.text()

### LINKED ISSUE #184: Charset detection
Originally reported by @gwicke in gwicke/node-fetch-polyfill#1.

Currently, node-fetch attempts to automatically detect the encoding of the input source, and [convert](https://github.com/bitinn/node-fetch/blob/e926483448c675b7443d2c085c49cee9d4619120/lib/body.js#L147) it to UTF-8. However, the [Fetch Standard](https://fetch.spec.whatwg.org/#concept-body-package-data) says that only a plain UTF-8 decode should be performed, regardless of the actual input encoding. As written in the referenced issue and confirmed by myself, the spec behavior is what's implemented by the browsers.

A few options:

1. Keep the current behavior
2. Remove conversion process
3. Use an option (e.g. `res.text({ convert: true })`)
4. Make a new method (e.g. `res.textConverted()`)

I'm leaning towards 4. 1 makes this module incompatible with the browser (something I don't want to do), 2 makes transition to v2 more difficult for people already using this library, and I'm reluctant to modify the behaviors of a well-defined function such as `.text()`.

Thoughts?

TimothyGu:
Also uses iconv-lite directly instead of using the "encoding" package.

Fixes #184.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -78,6 +78,16 @@ export default class Body {
 	}
 
 	/**
+	 * Decode response as text, while automatically detecting the encoding and
+	 * trying to decode to UTF-8 (non-spec api)
+	 *
+	 * @return  Promise
+	 */
+	textConverted() {
+		return this[CONSUME_BODY]().then(buffer => convertBody(buffer, this.headers));
+	}
+
+	/**
 	 * Decode buffers into utf-8 string
 	 *
 	 * @return  Promise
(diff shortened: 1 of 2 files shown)


bitinn:
I am not sure why we want to switch to `iconv-lite`? It's not like `encoding` adds a lot of complexity.

Note that encoding supports `iconv` which in turns support encoding like `ISO-2022-JP`, `iconv-lite` currently doesn't. So this is likely a breaking change and developers can't fix it.

TimothyGu:
> Note that encoding supports `iconv` which in turns support encoding like `ISO-2022-JP`, which `iconv-lite` currently doesn't. So this is likely a breaking change and developers can't fix it.

Good point -- though `iconv` is not a dependency or optionalDependency of either `node-fetch` or `encoding`. I'll revert that change.

PR #194, opened 2016-10-22
One less dependency

jimmywarting:
I assume this won't make util included when you babelify this


TimothyGu:
Unfortunately as you can see from the Travis failure, Babel's transpilation doesn't play well with `instanceof` checks, which was why I didn't migrate to `class`. In the spirit of the original change, however, I've committed 4a9a3246f633c3ddb358d3e79a38b39bb2a61bb1, which removes the dependency as well.

PR #197, opened 2016-11-08
Support disabling the agent connection pooling. Fixes #196

### LINKED ISSUE #196: No way to disable Agent with options
Due to the shortcutting in line: https://github.com/bitinn/node-fetch/blob/master/lib/request.js#L51 one cannot opt out of the pooling provided by the http/https library. Setting `options.agent = false` causes this to operate and use the other side of the `||` statement, which is undefined, opting into the pooling and using the global agent.

This can cause use cases that aren't as expected and aren't in accordance with the node.js http documentation. https://nodejs.org/api/http.html#http_class_http_agent

voodootikigod:
This will allow you to provide `{agent: false}` as an option to disable Agent pooling per the node.js documentation.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -48,7 +48,7 @@ function Request(input, init) {
 		init.compress : input.compress !== undefined ?
 		input.compress : true;
 	this.counter = init.counter || input.counter || 0;
-	this.agent = init.agent || input.agent;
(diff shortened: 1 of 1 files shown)


TimothyGu:
@bitinn, is this change necessary?

bitinn:
I really don't see a need for it, see my reply #196, closing

PR #200, opened 2016-11-23
Add support for blobs

### LINKED ISSUE #78: .blob() support
Hi

So I have been using node-fetch for a while now and for the first time I needed to request an image and get its size, now in the client I can do.

```
fetch('imageURL').then(function(response) {
    return response.blob();
}).then(function(resp){
    console.log(resp.size)
});
```

But unfortunately blob isn't supported by node-fetch so this won't work

Thanks

Jonathan

TimothyGu:
The `Blob` class isn't currently exposed as it is not really interoperable with any other library (the buffer is opaque from the user), and so far the only use cases for it in Node.js are for isomorphism with the browser. Eventually I might write a full implementation for the W3C File API in Node.js, so until then I don't want to encourage the use of it.

Fixes #78.

### DIFF
--- a/src/blob.js
+++ b/src/blob.js
@@ -1,0 +1,104 @@
+// Based on https://github.com/tmpvar/jsdom/blob/aa85b2abf07766ff7bf5c1f6daafb3726f2f2db5/lib/jsdom/living/blob.js
+// (MIT licensed)
+
+export const BUFFER = Symbol('buffer');
+const TYPE = Symbol('type');
+const CLOSED = Symbol('closed');
+
+export default class Blob {
+	constructor() {
+		Object.defineProperty(this, Symbol.toStringTag, {
+			value: 'Blob',
+			writable: false,
+			enumerable: false,
+			configurable: true
+		});
+
+		this[CLOSED] = false;
+		this[TYPE] = '';
+
+		const blobParts = arguments[0];
+		const options = arguments[1];
+
+		const buffers = [];
+
+		if (blobParts) {
+			const a = blobParts;
+			const length = Number(a.length);
+			for (let i = 0; i < length; i++) {
+				const element = a[i];
+				let buffer;
+				if (element instanceof Buffer) {
+					buffer = element;
+				} else if (ArrayBuffer.isView(element)) {
+					buffer = new Buffer(new Uint8Array(element.buffer, element.byteOffset, element.byteLength));
+				} else if (element instanceof ArrayBuffer) {
+					buffer = new Buffer(new Uint8Array(element));
+				} else if (element instanceof Blob) {
+					buffer = element[BUFFER];
+				} else {
+					buffer = new Buffer(typeof element === 'string' ? element : String(element));
+				}
+				buffers.push(buffer);
+			}
+		}
+
+		this[BUFFER] = Buffer.concat(buffers);
+
+		let type = options && options.type !== undefined && String(options.type).toLowerCase();
+		if (type && !/[^\u0020-\u007E]/.test(type)) {
+			this[TYPE] = type;
+		}
+	}
+	get size() {
+		return this[CLOSED] ? 0 : this[BUFFER].length;
+	}
+	get type() {
+		return this[TYPE];
+	}
+	get isClosed() {
+		return this[CLOSED];
+	}
+	slice() {
+		const size = this.size;
+
+		const start = arguments[0];
+		const end = arguments[1];
+		let relativeStart, relativeEnd;
+		if (start === undefined) {
+			relativeStart = 0;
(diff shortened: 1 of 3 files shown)


jimmywarting:
Not so sure that i'm a fan of the W3C File api in NodeJS. And if we do proceed with a html5 File API then there is something that has already been develop here: https://github.com/node-file-api

I'm mostly against this idea

TimothyGu:
@jimmywarting, as you can see that project 1) is dead, and 2) does not have an implementation for Blob in the first place. Also if you look at e.g. #78, there isn't an easy way to find out the file size in an isomorphic fashion, except through `Blob`.

jimmywarting:
how about doing `fetch(...).then(res => res.arrayBuffer()).then(buffer => console.log(buffer.byteLength))`

TimothyGu:
Okay, granted that'll work. But how about the type? The Blob constructor is obligated to do certain [sanitization](https://w3c.github.io/FileAPI/#ref-for-dfn-BPtype-3) on the Content-Type header which we don't currently do.

See the thing is that it really doesn't take much to support blobs just for compatibility's sake. There is no external dependency involved, and the implementation for `blob()` is fairly simple.

jimmywarting:
I also think you could do that yourself

```js
res.headers.get('content-type') // often this is enough 
```

I believe nobody want's to use blob on server side cuz there is nothing you can do with it if you don't have FileReader also

bitinn:
@TimothyGu Do you think you can conjure up a use case of Blob on Node, I tried but failed.

I am also a bit concerned about how we can test our own implementation of Blob, are the tests enough?

TimothyGu:
I think the primary use is isomorphism, and the question of whether or not to implement it is another case of the question of to what extent do we want to pursue isomorphism. I abandoned the idea of supporting Web Streams, because to support it a huge polyfill would have to be brought in. This is different, in that `Blob` is merely a container for a `Buffer`, and therefore no performance penalty would be placed.

But isomorphism might not necessarily be the only factor. Because a blob stores an MIME type as well, doing round-tripping with blobs can be a lot clearer. Imagine the following:

```js
fetch('https://I.me/image.jpg')
  .then(res => res.blob())
  .then(blob => fetch('https://I.me/submit', {
    method: 'POST',
    body: blob
  }))
```

```js
fetch('https://I.me/image.jpg')
  .then(res => res.buffer().then(buf => fetch('https://I.me/submit', {
    method: 'POST',
    body: buf,
    headers: {
      'content-type': res.headers.get('content-type')
    }
  })))
```

(granted, async functions make it easier, but their performance is still not up to par neither natively nor through Babel, except when you use a nonstandard-compliant implementation [fast-async](https://github.com/MatAtBread/fast-async) is used)

Regarding the tests I don't think the ones in node-fetch are necessarily enough, but because I pulled the `Blob` implementation from JSDOM, which uses the official [Web Platform Tests](https://github.com/w3c/web-platform-tests), I'm fairly sure they are properly implemented.

bitinn:
That's a proper use case, thx a lot for writing it up! We might be the module that finally make Blob easily usable on Nodejs ;)

(From my phone)

jimmywarting:
> That's a proper use case, thx a lot for writing it up! We might be the module that finally make Blob easily usable on Nodejs ;)

Separate module? Don't make it part of node-fetch

jimmywarting:
I suggest this becomes a separate module...

TimothyGu:
@jimmywarting, yes, it is my intention to eventually move it to a separate module, which is why I'm not currently exposing it to the user.

jimmywarting:
Yet you merged this anyway...

Btw take a look at this: https://github.com/jimmywarting/Screw-FileReader
bottom line is they are thinking about adding this new standard methods:

- blob.prototype.arraybuffer
- blob.prototype.text
- blob.prototype.stream (unclear how yet)


I just went ahead and implemented it already :)
Maybe you don't have to implement FileReader then at least :P

TimothyGu:
@jimmywarting, PR welcome :)

jimmywarting:
Do you want to work with me on a new repo? implementing blob/file for node?

TimothyGu:
@jimmywarting, sure. I'd base it on the `blob.js` already in the tree. However, I do still want to implement FileReader to maintain compatibility with the browser.

jimmywarting:
https://github.com/jimmywarting/File-API

gap777:
Any chance this work can be made available in the 1.x versions? I don't think I can wait for it until v2 is stable and released.

TimothyGu:
@gap777 v2 alpha is pretty stable already, and has behaviors up to date with the latest browsers, so I'd just suggest using that.

PR #201, opened 2016-11-23
Revert "Return empty .json() object on 204. Fix #165. (#166)"

### LINKED ISSUE #165: Return empty object on 204 No Content...

TimothyGu:
This reverts commit 95b58936b8673f7fdc21050c3bfeb460e9b00bf1.

Fixes #165.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -51,11 +51,6 @@ export default class Body {
 	 * @return  Promise
 	 */
 	json() {
-		// for 204 No Content response, buffer will be empty, parsing it will throw error
-		if (this.status === 204) {
-			return Body.Promise.resolve({});
-		}
-
 		return this[CONSUME_BODY]().then(buffer => JSON.parse(buffer.toString()));
 	}
 
--- a/test/test.js
+++ b/test/test.js
@@ -471,16 +471,6 @@ describe(`node-fetch with FOLLOW_SPEC = ${defaultFollowSpec}`, () => {
 		});
 	});
 
-	it('should return empty object on no-content response', function() {
-		url = `${base}no-content`;
-		return fetch(url).then(res => {
-			return res.json().then(result => {
-				expect(result).to.be.an('object');
-				expect(result).to.be.empty;
-			});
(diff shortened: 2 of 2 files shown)


hmarr:
Hello! Any chance this could be back-ported to a 1.x.x release? Not throwing the error broke compatibility with browser fetch implementations, so I'd consider this commit a fix that gets us back to the previous desired behaviour.

Thanks!

bitinn:
It's been a while since I looked at 1.x release, if it's indeed affected I can take a PR to fix it.

PS: 2.x should be ready for use if you want to. We don't intend to backport spec related changes to 1.x if possible (as 2.x is intended to address this issue without raising too much upgrade problems).

(From my phone)

hmarr:
Ok, thanks @bitinn. What's the ETA for a 2.0 final release?

I'm using node-fetch via [fetch-mock](https://github.com/wheresrhys/fetch-mock), which [depends on 1.x](https://github.com/wheresrhys/fetch-mock/blob/ecac8c5e8e888a8148b5f84ac07bd349708bed59/package.json#L33), and I imagine they'd want to wait for a non-prerelease of v2 before upgrading.

If it's going to be a while off, I'll put in a PR for 1.x doing the backport.

bitinn:
@hmarr actually I have already fixed it in recent v1.7.0 release :)

hmarr:
@bitinn brilliant, thanks!

Issue #203, opened 2016-11-25
What is the proper way to abort an infinite response?


voxpelli:
When doing a request against a realtime API that returns an infinitely long API response, such as a newline delimited stream of JSON objects representing new activities in a system, what is the proper way to abort it if one no longer wants to listen to it?

Just stopping the consumption of the stream doesn't abort the request, at the moment at least. Something more is needed and if I remember correctly then the proper Node.js way of aborting would be to call `.destroy()` on the response `IncomingMessage`, as described in eg. [the node.js docs](https://nodejs.org/api/http.html#http_message_destroy_error).

Calling `.destroy()` isn't currently doable though as the `IncomingMessage` isn't exposed in the `Response` and thus not reachable for any code outside of node-fetch itself.

The way to cancel the response in the browser would seemingly instead be to call `.cancel()` on the  `ReadableStream`, as specified by [the streams spec](https://streams.spec.whatwg.org/#rs-cancel) and eg. [described by Jake Archibald](https://jakearchibald.com/2015/thats-so-fetch/#streams), but neither that is implemented,

Maybe the proper solution to this would be to implement the `.cancel()` and have that call, among perhaps other things, the `.destroy()` on the `IncomingMessage`?

This is probably similar to #95

bitinn:
I have replied in #204 about the status quo of cancellable fetch (TL;DR: not great).

But if you must use Fetch API for JSON streaming, let's see if there are other workarounds:

- Can you use a [custom agent](https://nodejs.org/api/http.html#http_agent_destroy) to destroy the socket? `node-fetch` allow passing an agent for this kind of purpose.

- Can you set a long `timeout` so that you can consistently re-initiate the request?

(Note that only `node-fetch` supports these, as [client-side development on these are not yet in progress](https://github.com/whatwg/fetch/issues/180).)

bitinn:
And note that the whole Fetch mental model is:

- If your response is complete, promise is fulfilled (even if response is 404).
- If your response is running, promise is pending.
- If your network/remote-server is down, promise is rejected (even if you terminate it manually).

So if you use Fetch with JSON streaming, you should expect promise handling to be a bit weird in your code.

voxpelli:
Answered in #204, but core take away: This is about canceling responses, not requests. The requests are successful and the promises are fulfilled, but as the responses are infinite there's a need to cancel them as well because they will never finish on their own.

TimothyGu:
[...]

Issue #95, opened 2016-04-01
Aborting a fetch


jimmywarting:
Implement `AbortController` so we can abort a request

bitinn:
Not at the moment, you could set a shorter timeout and handle the error, but no abort.

jimmywarting:
:disappointed: Don't want that. I'm doing a long poll request since they don't support sockets...
Would rather have a non standard method until it gets shipped into the spec and deprecate it later

The way AngularJS solved it was with a controller. The timeout could either be a number or a promise
(docs)[https://docs.angularjs.org/api/ng/service/$http#usage]

bitinn:
@jimmywarting I don't think `node-fetch` will suit your use case.

Try https://github.com/sindresorhus/got to a similar API, they should allow you to abort.

jimmywarting:
maybe `res.body.cancel()` cancels the request?

jimmywarting:
Any news about this? I'm pretty certain `res.body.cancel` aborts the download stream doz canceling the response. 

With that you won't be able to abort the hole request/response since you need to fetch the headers first, but i'm fine with that since i only wan't to check against cache headers eg like `if-modified-since`.

I don't know where they stands on cancelable promises but i saw a interesting talk about async/await + cancelable promises (don't remember where - think it was a video of es7 in "weekly javascript")

TimothyGu:
Cancelable Promises Proposal is still at Stage 1, and since we don't use Web Streams there are nothing node-fetch can do to support termination except to point users to Node.js's documentation.

For your specific use case, is there a reason why you are not using `HEAD` requests?

jimmywarting:
You would avoid making 2 request...

Maybe i can tackle this in an other way... is there something equivalent to web streams cancel method that node's stream has? can i maybe do `res.body.end()`?

TimothyGu:
Hmm good point. You _could_ get by with `res.body._dump()`, but ideally we should provide the Node.js `ClientRequest` object as well in `Response` class for its [`abort`](https://nodejs.org/api/http.html#http_request_abort) method.

TimothyGu:
The cancelable promises proposal was recently withdrawn (see tc39/proposal-cancelable-promises#70 and whatwg/fetch#27). 

New tracking issue: whatwg/fetch#447.

jimmywarting:
[...]

PR #204, opened 2016-11-25
Add cancel() method to streams

voxpelli:
I'm unfortunately somewhat rusty in this regard, but I think this adds the possibility of canceling an infinite response as requested by me in #203 (and maybe as requested in #95).

This would add a `cancel()` that at least somewhat mimics the browser's `ReadableStream.cancel()`. To fully mimic it would probably have to [return a Promise](https://streams.spec.whatwg.org/#rs-cancel), but as there's no callback to the `message.destroy()` in node.js it would have to just be a `Promise.resolve()` then.

Thoughts on this approach? Good or bad?

Quickly looking at v2 it seems like this could be easily forward ported to that version.

### DIFF
--- a/index.js
+++ b/index.js
@@ -177,9 +177,7 @@ function Fetch(url, opts) {
 			// prepare response
 			var body = res.pipe(new stream.PassThrough());
 
-			body.cancel = function () {
-				res.destroy();
-			};
+			body.clientResponse = res;
 
 			var response_options = {
 				url: options.url
--- a/index.js
+++ b/index.js
@@ -176,6 +176,11 @@ function Fetch(url, opts) {
 
 			// prepare response
 			var body = res.pipe(new stream.PassThrough());
+
+			body.cancel = function () {
+				res.destroy();
+			};
+
 			var response_options = {
 				url: options.url
 				, status: res.statusCode


jimmywarting:
Actually... `cancel()` is something that is part of the ReadableStream api not fetch per say...
The `cancel` method should actually do a little bit more then simply call `res.destroy()` It should only cancel the stream if there is nothing else that is consuming the stream (aka if it's locked to something else).

The stream gets consumed/locked when you call `.blob()`,  `.arrayBuffer()`,  `.text()` or `.body.getReader()`

```js
var body = new Response('abc').body
var reader = body.getReader()
console.log(body.locked) // true
// The reason i think it return's a promise is cuz it can throw a error (Perhaps they thought it's nicer if it silence reject then throwing and making the rest crash)
body.cancel() // Uncaught TypeError: ReadableStreamReader constructor can only accept readable streams that are not yet locked to a reader
```

What needs to be done is 
```js
var body = new Response('abc').body
var reader = body.getReader()
reader.releaseLock() // Release the lock so no stream are using it
console.log(body.locked) // false
body.cancel() // now it will cancel
```

But the ReadableStream API are different then the Node stream Api
I'm not sure where we stand on implementing ReadableStream api in v2... The hole node eco system is built around node-streams so what would be the benefit of having ReadableStream api if you can't pipe it to a file?

voxpelli:
@jimmywarting I'm mostly just trying to figure out a solution to #203, this might not be it, but the ability to cancel a never-ending infinite response is something I need no matter what kind of streams model it would be modeled around. Suggestion on how to do this with node streams?

jimmywarting:
Yea, I agree, something is needed. We may need some differentiation in node-fetch since it's built on node-stream

i don't know if it's possible to detect if a stream has been locked to something
if you where to example do

```js
body = new Response('abc').body
body.pipe(new fs.createWriteStream('cache.bin'))
// Do we know if there is something that is consuming the stream?
// ideally `body.locked` should return true and therefore cancel should not work
```

I would be fine with a less stricter api and allow canceling a response even if that means something else is consuming the stream. After all it's two different streaming api

bitinn:
I might be wrong as I haven't used this personally.

- The original `cancel()` is really intended to be used when you don't care about the result *at all*. And as far as I know, this won't be added to spec.
- The new approach appear to be *cancel token*, see [proposal-cancelable-promises](https://github.com/tc39/proposal-cancelable-promises) and an [implementation](https://github.com/mzabriskie/axios/pull/452) from [axios](https://github.com/mzabriskie/axios#cancellation), though it was not clear [how they handle streaming and cancelling](https://github.com/mzabriskie/axios/issues/499).

So to sum up the current status quo:

- If you want to abort and discard the body, *cancel token* can help.
- But if you want to stop the stream somehow, this is really a [Node.js problem](http://stackoverflow.com/questions/19277094/how-to-close-a-readable-stream-before-end).

I don't think we should add `res.body.cancel()` because it confuses people, but if the case for real-time JSON streaming is really popular, we might consider adding an extension to allow developers to control termination of underlying http request...

TimothyGu:
The `cancel()` is already in the spec as part of the [`ReadableStream`](https://streams.spec.whatwg.org/#rs-cancel) interface. However I'm not convinced we should augment a Node.js `ReadableStream` object with WHATWG `ReadableStream` semantics -- doing so might trick the user into thinking the `res.body` we provide is actually a WHATWG `ReadableStream`, which is not the case.

I wouldn't mind adding support for Promise-based cancelation, but I don't think we should have our own implementation of `CancelToken` either. Nor is there any standalone polyfill for `CancelToken` available on npm AFAICT.

I'd agree with exposing the original `ClientRequest` object. I think it can be beneficial to Node.js users in general to have access to that object.

bitinn:
@TimothyGu I haven't tried this, but does calling `destroy()` on a custom `agent` passed to `node-fetch` works? If so, we might not need to expose ClientRequest (I think 99% use case of Fetch is to avoid dealing with http.request directly?).

voxpelli:
I'm aware of cancelable promises, so just to be clear: This and #203 is _not_ about cancelable _requests_, it's about fully resolved requests where one wants to _cancel the response_.

An example of a properly fulfilled request where the response needs to be canceled:

```javascript
fetch('http://example.com/infinite-response').then(res => {
  // The request is done, but now I want to abort the response
});
```

This type of streaming through an infinite responses is rather common and was in the form of [newline delimited JSON](https://en.wikipedia.org/wiki/JSON_Streaming#Line_delimited_JSON) popularised by Twitter's [streaming API](https://dev.twitter.com/streaming/overview/processing) and is made easy in Node.js through packages like [ndjson](https://www.npmjs.com/package/ndjson).

It all works great with node-fetch – apart from the fact that the response  can't ever be destroyed.

Abusing HTTP agents to ensure that the response is destroyed feels wrong to me. HTTP agents are responsible for many other things, including keep-alive, which may be useful for also this type of request.

It's also not enough to just stop the stream – the response also needs to be destroyed so that the http connection is released or destroyed.

The simplest way of ensuring that the response is destroyed would be to let the one who knows that the request should be destroyed also be the one who destroys it.

Therefore a `.cancel()` (perhaps with another name) would probably be the easiest solution to this problem.

bitinn:
@voxpelli

I believe status quo is:

- WHATWG stream `cancel()` can't be used when you are actively piping to something.
- NODE stream doesn't have a `cancel()`, however HTTP request do have `destroy()` and `abort()`.

I don't think `res.destroy()` is more clean than `agent.destroy()`. And by that I mean they both forcefully close the socket, see [`socket.destroy() doc`](https://nodejs.org/api/net.html#net_socket_destroy_exception).

[`request.abort()`](https://nodejs.org/api/http.html#http_request_abort) is nicer, it just throw away any remaining data and then close socket. Not sure if it works for your case though.

@TimothyGu 

> I'd agree with exposing the original ClientRequest object. I think it can be beneficial to Node.js users in general to have access to that object.

How do you propose we expose this object? Through [event like got](https://github.com/sindresorhus/got#onrequest-request)?

TimothyGu:
What I had on my mind was as a property on the `Response` object:

```js
fetch('...').then(res => res.clientRequest.cancel());
```

> it just throw away any remaining data and then close socket. Not sure if it works for your case though.

Don't see a reason why it wouldn't...

voxpelli:
The suggestion of @TimothyGu would work for me, I updated the PR to match that with the one exception that I named the property `clientResponse` as it's a response object rather than the request object that's exposed.

This way one can do:

```javascript
fetch('...').then(res => res.clientResponse.destroy());
```

index.js
```diff
@@ -176,6 +176,9 @@ function Fetch(url, opts) {
 
 			// prepare response
 			var body = res.pipe(new stream.PassThrough());
+
+			body.clientResponse = res;
```
TimothyGu: When decompression is requested, `body` is overwritten, so this `clientResponse` will get lost.

I was thinking of passing `req` as `clientRequest` through `response_options` into the `Response` constructor, and make `clientRequest` a property of the `Response` class. This way, not only would one be able to do:

```js
fetch('...').then(res => res.clientRequest.abort());
```

and cancel both the request and the response, which is especially helpful if chunked encoding is used.

heri16:
@TimothyGu this was closed without being merged?

PR #212, opened 2016-12-06
Use ES2015 export syntax

TimothyGu:
The following description is now irrelevant and incorrect.

----

This should be the last functional change before a 2.0.0-alpha.1 release. It makes it possible to use this module directly with [Rollup](http://rollupjs.org/), which we are already using to generate the output bundles.

The main concern can be that it requires more typing on the user's part. However, it is not always true:

Before:

```js

import fetch, { Headers, Response } from 'node-fetch';

const { Headers, Response } = fetch;

const Headers = fetch.Headers;
const Response = fetch.Response;
```

After:

```js
import { fetch } from 'node-fetch';

import { fetch, Headers, Response } from 'node-fetch';

const { fetch } = require('node-fetch');

const fetch = require('node-fetch').fetch;

const { fetch, Headers, Response } = require('node-fetch');

const fetch = require('node-fetch').fetch;
const Headers = require('node-fetch').Headers;
const Response = require('node-fetch').Response;
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -16,11 +16,8 @@
       "presets": [
         [ "es2015", { "loose": true } ]
       ],
-      "ignore": [
-        "build"
-      ],
       "plugins": [
-        [ "istanbul", { "exclude": [ "src/blob.js", "test" ] } ],
+        [ "istanbul", { "exclude": [ "src/blob.js", "build", "test" ] } ],
 	"./build/babel-plugin"
       ]
     },
--- a/.babelrc
+++ b/.babelrc
@@ -6,19 +6,22 @@
     "test": {
       "presets": [
         [ "es2015", { "loose": true } ]
+      ],
+      "plugins": [
+        "transform-runtime",
+	"./build/babel-plugin"
       ]
     },
     "coverage": {
       "presets": [
         [ "es2015", { "loose": true } ]
       ],
+      "ignore": [
+        "build"
+      ],
       "plugins": [
-        [ "istanbul", {
-          "exclude": [
-            "src/blob.js",
-            "test"
-          ]
-        } ]
+        [ "istanbul", { "exclude": [ "src/blob.js", "test" ] } ],
+	"./build/babel-plugin"
       ]
     },
     "rollup": {
--- a/build/babel-plugin.js
+++ b/build/babel-plugin.js
@@ -1,0 +1,48 @@
+// This Babel plugin makes it possible to do CommonJS-style function exports
+
+const walked = Symbol('walked');
+
+module.exports = ({ types: t }) => ({
+  visitor: {
+    Program: {
+      exit(program) {
+        if (program[walked]) {
+          return;
+        }
+
+        for (let path of program.get('body')) {
+          if (path.isExpressionStatement()) {
+            const expr = path.get('expression');
+            if (expr.isAssignmentExpression() &&
+                expr.get('left').matchesPattern('exports.*')) {
+              const prop = expr.get('left').get('property');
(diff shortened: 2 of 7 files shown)


bitinn:
Personally, I don't enjoy this, but mostly because it breaks my own production server :/

Is there really no other way to work with Rollup? `const fetch = require('node-fetch');` isn't this the default for most other npm modules? I must be missing some details as I haven't worked with Rollup.

bitinn:
OK so Rollup basically trade CommonJS support for smaller build:

https://github.com/rollup/rollup/wiki/ES6-modules

I don't know how I feel about this.

bitinn:
I think a more appropriate title for this should be: Changing module from CommonJS export style to ES6 module export style.

TimothyGu:
@bitinn, it doesn't _trade_ CommonJS support for ES2015 modules, but rather _adds_ the ability to have smaller builds for downstream users.

Right now, the transpilation apparatus creates two files: `lib/index.js` and `lib/index.es.js`. The first uses standard CommonJS style imports/exports, while the second uses ES2015 style imports, but CommonJS style exports, which does not allow downstream Rollup/node-fetch users to take advantage of the features Rollup offers.

After this PR, `lib/index.js` still uses CommonJS patterns, but sets `exports.fetch` instead of `module.exports` to `function fetch() {}`. In fact, after the PR `lib/index.js` is arguably more CommonJS-compliant (since `module.exports` is not part of CommonJS's Modules/1.1 specification). However, `lib/index.es.js` will no longer be the odd CJS/ES2015 hybrid it is currently, since the exports will have ES2015 syntax as well.

bitinn:
@TimothyGu If you read the wiki I linked, this does [*trade*](https://github.com/rollup/rollup/wiki/ES6-modules#es6-friendly-libraries).

Let's say you are right, the default npm way is *less compliant* in terms of CommonJS spec. You are asking every single production use of `node-fetch` out there to update their code. Which I think is unlikely to happen.

bitinn:
To be honest, I don't want to be that guy that stands in the way of `node-fetch` moving to ES6 module, I am just saying, please consider the implication. And if there is another solution, I am interested to hear it.

jimmywarting:
I like this. I think it's nicer to do
```js
const { fetch, Headers, Response } = require('node-fetch');
```
then
```js
const { Headers, Response } = fetch;
```

if you only need fetch then you only need you one extra `{}` wrapt around it
It's breaking changes but this is the time to do it when releasing a new major release 

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.

bitinn:
From the lack of replies I guess there isn't much one can do to allow both?

```
// Style most people use on Node before ES6
var fetch = require('node-fetch');

// ES6 preferred style
import { fetch, Headers, Response } from 'node-fetch';
```

Another question being, is it a requirement for these 2 to match:

```
// If we support this
import { fetch, Headers, Response } from 'node-fetch';

// Must we also support this (or would it be weird if require returns only fetch)
const { fetch, Headers, Response } = require('node-fetch');
```

Basically I am looking for a way for users to opt-in v2 as smoothly as possible.

TimothyGu:
Changed in such a way that existing code bases will continue to work, _and_ Rollup will begin to work beautifully with this module. Therefore, this should now be uncontroversial and will be merged.

bitinn:
Just a question on this: what would various export style be at this point? I understand the old way still works, how does it affect the new way?

TimothyGu:
The "new way" does not work.

PR #215, opened 2016-12-13
proper stack first line for FetchError instances

TooTallNate:
The `.stack` property gets cached in the `captureStackTrace()`
call, so whatever is set as the `name` and `message` at that time
will be used for the first line of the stack trace.

Before this patch, FetchError's stack would just say "Error"
as the first line. Now they correctly display the
"${name}: ${message}" of the error instances.

Test case included.

### DIFF
--- a/lib/fetch-error.js
+++ b/lib/fetch-error.js
@@ -16,10 +16,6 @@ module.exports = FetchError;
  * @return  FetchError
  */
 function FetchError(message, type, systemError) {
-
-	// hide custom error implementation details from end-users
-	Error.captureStackTrace(this, this.constructor);
-
 	this.name = this.constructor.name;
 	this.message = message;
 	this.type = type;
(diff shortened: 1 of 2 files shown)


TimothyGu:
Merged to v2 branch in fa22529.

TooTallNate:
Hey @TimothyGu. Just wanted to chime in and say that I've bumped into this error multiple times from modules in the wild since this patch has been merged. Since v2 still hasn't ben released, can this patch be backported to a v1.x release? It's annoying having to dig into the code when this happens and log the `err.message` instead of `err.stack`.

TimothyGu:
cc @bitinn, who more or less *de facto* maintains the v1.x line.

bitinn:
@TooTallNate do you mind sending a PR to branch `1.x`? or open an issue so I remember to get around to it.

PR #216, opened 2016-12-14
Url to string

denar90:
[fetch method](https://fetch.spec.whatwg.org/#fetch-method) gets [RequestInfo](https://fetch.spec.whatwg.org/#requestinfo) parameter which can either be `Request` or be a `String`.
Using [URL](https://url.spec.whatwg.org/#URL-stringification-behavior) in that case is not possible. 
Because of `stringifier`, which
> indicates that objects that implement the interface **have a non-default conversion to a string**
> https://heycam.github.io/webidl/#idl-stringifiers

Hence, when param is not an instanceof `Request` it has to be converted to string.

Special thanks to @shvaikalesh

UPD:
> Using [URL](https://url.spec.whatwg.org/#URL-stringification-behavior) in that case is not possible. 

It's possible natively in browsers, but in scope of `node-fetch` it's not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
   - "node"
 env:
   - FORMDATA_VERSION=1.0.0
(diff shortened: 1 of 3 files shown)


TimothyGu:
This should be fixed in the v2 branch already. Can you file a new PR against v2, with only the test case? Thanks.

PR #217, opened 2016-12-14
simplify request input convertation

denar90:
Continuing #216 
I've found out that [tests](https://github.com/bitinn/node-fetch/blob/v2/test/test.js#L1462-L1482) for Node.js Url and WHATWG's URL are present. 

But I propose to simplify input convertation regarding to `template literals` [specification](http://www.ecma-international.org/ecma-262/6.0/#sec-template-literals-runtime-semantics-evaluation)
> The string conversion semantics applied to the Expression value are like String.prototype.concat rather than the + operator.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -24,15 +24,8 @@ export default class Request extends Body {
 
 		// normalize input
 		if (!(input instanceof Request)) {
-			if (input && input.href) {
-				// in order to support Node.js' Url objects; though WHATWG's URL objects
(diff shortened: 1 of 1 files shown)


TimothyGu:
The template literal change is fine.

The removal of if statement is not, since `input` can technically be `null`, in which case it should be stringified.

Please just convert the `+ ''` to template literals, and leave the if statement.

TimothyGu:
Closed by 7970491.

PR #220, opened 2017-01-09
Array Type-Checking

thomashuston:
In migrating an HTTP test suite that uses `node-fetch` from Mocha to Jest, a number of tests starting failing when trying to make assertions on cookies. When I dug into it, I traced it back to this Array type check in `Headers`.

Basically, because [Jest executes tests in a vm](https://github.com/facebook/jest/issues/2048), the class instances of core types like `Array` are different from those within Node's libraries like `http`, so `instanceof` checks against them fail.

The [recommended approach](http://stackoverflow.com/questions/22289727/difference-between-using-array-isarray-and-instanceof-array) for checking if an object is an array is to instead use the `Array.isArray` function.

This PR also includes a fallback for old versions of Node that don't include `Array.isArray` for maximum backwards compatibility.

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -7,6 +7,10 @@
 
 module.exports = Headers;
 
+var isArray = Array.isArray || function isArray(obj) {
+    return Object.prototype.toString.call(obj) == '[object Array]';
+};
+
 /**
  * Headers class
  *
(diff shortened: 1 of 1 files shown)


bitinn:
This makes sense, I will look into it when available.

(From my phone)

bitinn:
@TimothyGu I don't consider this a bug per se, do you want to fix it in v2 branch? It got babel for transpiling right?

TimothyGu:
@bitinn, yeah, sure.

@thomashuston, is the `Object.prototype.toString.call(obj) == '[object Array]'` check necessary? In other words, is `Array.isArray` sufficient for it to work?

thomashuston:
@TimothyGu You're correct, we can just use `Array.isArray`. I just verified that it was present as far back as Node 0.10. Will update.

TimothyGu:
Seems like this problem is already fixed in v2 branch, so PR closed. Thanks for the heads-up though.

Issue #146, opened 2016-08-14
Sending cookies


papigers:
Taken from [`fetch`](https://github.com/github/fetch) documentation:

> Sending cookies
> 
> To automatically send cookies for the current domain, the credentials option must be provided:
> 
> ``` js
> fetch('/users', {
>   credentials: 'same-origin'
> })
> ```
> 
> This option makes fetch behave similar to XMLHttpRequest with regards to cookies. Otherwise, cookies won't get sent, resulting in these requests not preserving the authentication session.
> 
> Use the include value to send cookies in a cross-origin resource sharing (CORS) request.
> 
> ``` js
> fetch('https://example.com:1234/users', {
>   credentials: 'include'
> })
> ```

This functionality should be added to `node-fetch`. I'll explain why I need this:

I'm developing an isomoprhic application which uses react, graphql, and relay. This isomorphic application, upon client requests, renders the view, and fetches the required data using relay on the server, while following requests and renders are done on the client.

As you may or may not know, `relay` uses `isomorphic-fetch` which uses node-fetch on the server. This makes the cookies not available on my graphql endpoint.
My current solution is to set the cookies manually:

``` js
{
  headers:{
    cookie: req.headers.cookie,
  },
}
```

TimothyGu:
node-fetch doesn't have a built-in cookie store, so you have to manually specify where the cookies are for node-fetch to be aware of them. Since GitHub's fetch runs in the browser, it does have access to the browser cookie store. CORS of course [doesn't apply](https://github.com/bitinn/node-fetch/blob/master/LIMITS.md) either.

Yes, we [do](https://github.com/bitinn/node-fetch#motivation) know about isomorphic-fetch.

papigers:
And what about adding a cookie store? (#94)

bitinn:
Cookie store is a big can of worms on server-side. While your request to emulate browser is fair, it's quite likely other people will see cookie store and think of different usages.

I do know 2 scenarios where having a cookie store will help (#8 and #94), but I have been trying to leave implementation to third-party, for example, [a simple solution](https://github.com/valeriangalliat/fetch-cookie) was mentioned in #94.

I understand some developers want isomorphism, but not distinguishing client/server side creates serious dependency problems down the road (client/server implementation need to support the same set of features, on an evolving spec). So I lean more towards having familiar API but knowing the gaps between implementations.

jimmywarting:
[...]

Issue #94, opened 2016-03-26
cookie jar


jimmywarting:
Maybe could be useful. what do you think? request uses it

https://www.npmjs.com/package/tough-cookie

bitinn:
While I agree this is likely a solid implementation of server-side cookie jar, I am currently leaning towards allowing extraction of headers during redirection, which should allow extraction of cookie headers, then you can parse or store them or pass them to the next request as you may see fit.

It's more elaborate than `request`'s solution, but it allows us to not depends on any single cookie implementation.

See #8 item 2 and 3.

jimmywarting:
You gave me an idea there...

It would be grate If we could extend node-fetch in some way with our own small module (like a cookie jar module or a cache module) (more or less like express, koa or connect `use` method) but in a way that reflects how browsers [service worker fetch event](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent) works

If we had a fetchEvent then we could mock data using [respondWith](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith). It would hook into all request being made (even all redirects). And we could build our own cookie jar

heck we could even modify all post data before it ever get sent (i think)

bitinn:
Fetch Event is really intended for Service Worker, we don't currently support that on the server side (for the face that you do have full control of your server).

Another approach would be to add a custom flag that ignores Fetch Spec, ie. people have argue for a `nofollow` option where it won't reject the fetch promise, but just don't follow redirect and returns the headers (which is useful if you POST and expect a redirect and set-cookie headers)

jimmywarting:
Anyway, I found this npm [fetch-cookie](https://www.npmjs.com/package/fetch-cookie) that solves my use case. However I still think it would be cool to have a fetchEvent that modules like fetch-cookie can hook into instead having to decorate it like he did

It could be used for authentication strategies in all request for a rest-api
If you were to use node-fetch as a proxy then it would be easy to just add cors headers to all response
if you want to decode/encode a different content-type (say [googles Protocol Buffers](https://developers.google.com/protocol-buffers/) `application/protobuf`) then it could be as easily as to just hook it into a fetchEvent
You could write your own network logger using fetchEvent instead of having to do it everywhere in the code
You could write your own cache mechanism that caches stuff and you would only have to write it once

think:
```
const fetch = require("node-fetch")
const fetchCache = require("fetch-cache")
const fetchCookie = require("fetch-cookie")({expire: '1Day'})
[...]

PR #229, opened 2017-02-05
Add a `fetch` option to allow wrappers

valeriangalliat:
This way a wrapper library like [fetch-cookie] could transparently leverage node-fetch way of handling redirects without having to force `{ redirect: 'manual' }` and re-implement the logic.

[fetch-cookie]: https://github.com/valeriangalliat/fetch-cookie

Instead, a wrapper could just pass the `fetch` option and point it to the wrapped function. And this would also allow wrappers to be combined as long as they also support being given the `fetch` option, e.g.:

```js
function fetchCookie(url, opts) {
    return fetch(url, Object.assign({ fetch: fetchCookie }, opts));
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,8 @@ The default values are shown after each option key.
 	timeout: 0,         // req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
-	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
+	agent: null,        // http(s).Agent instance, allows custom proxy, certificate etc.
+	resolver: null      // fetch-compatible function to use when following redirects (to support wrappers)
 }
 ```
 
--- a/src/index.js
+++ b/src/index.js
@@ -95,7 +95,10 @@ export default function fetch(url, opts) {
 
 				request.counter++;
 
-				resolve(fetch(resolve_url(request.url, res.headers.location), request));
+				// allow to pass a fetch-compatible function for supporting wrappers
+				const resolver = opts && opts.resolver || fetch;
+
+				resolve(resolver(resolve_url(request.url, res.headers.location), request));
 				return;
 			}
 
--- a/test/test.js
+++ b/test/test.js
@@ -396,6 +396,19 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should allow redirect resolver', function() {
+		url = `${base}redirect/301`;
+		opts = {
+			resolver: (url, opts) => fetch(url, opts).then(res => {
+				res.customResolver = true;
+				return res;
+			})
+		};
+		return fetch(url, opts).then(res => {
+			expect(res.customResolver).to.be.true;
+		});
+	});
+
 	it('should handle client-error response', function() {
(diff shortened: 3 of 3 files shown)


bitinn:
From what I see this saves you from [duplicating code](https://github.com/valeriangalliat/fetch-cookie/blob/master/node-fetch.js), basically allow you to hook into the redirection process which is where automate cookie handling is desired, true?

@TimothyGu I think this makes sense, just wondering if "fetch" option is the way to go. I think it solves #146 and #94, or at least make implementing them trivial to users.

valeriangalliat:
> From what I see this saves you from duplicating code, basically allow you to hook into the redirection process which is where automate cookie handling is desired, true?

Exactly. :+1:

ahmadnassri:
`resolver` instead of `fetch` would be my choice for naming.

valeriangalliat:
I can rename it to `resolver`. Would that be fine @bitinn?

bitinn:
@valeriangalliat I think `resolver` should do just fine, would you kindly add some comment, test and document for the change?

@TimothyGu what do you think, I do want this to happen, because it's one of the most discussed pain point of `node-fetch`, given we are unlikely ever going to implement server-side cookies ourselves.

valeriangalliat:
@bitinn I added comment, test and documentation for the `resolver` option. Does it looks good to you?

TimothyGu:
For now this PR looks fine to me. On the long term, however, I envision a mechanism to have a "store" of Web interfaces including `fetch`, `Promise`, `FormData`, `URL`, `Blob`, `URLSearchParams`, `CookieJar` etc., that the users can pass in, and for `node-fetch` to use.

The approach I have on my mind is to exports an additional `createFetch()` function that allows users to bind specific available features. The traditional `{ fetch, Headers, Request, Response }` API is created with `createFetch()` with a default set of available utilities (native Promise, etc.).

So to implement a fuller-featured `isomorphic-fetch`, something like this would work:

```js
// isomorphic-fetch-node.js
import { URL, URLSearchParams } from 'url';

import { Blob } from 'file-api';
import { FormData } from 'form-data';
import { createFetch } from 'node-fetch';
import { CookieJar } from 'tough-cookie';

export createFetch({
  Blob,
  CookieJar,
  FormData,
  URL,
  URLSearchParams
});
```

```js
// isomorphic-fetch-browser.js
import 'whatwg-fetch';
export { fetch, Headers, Request, Response };
```

And for traditional uses of the API can be extended:

```js
import { Blob, readBlob } from 'file-api';
import createFetchCookie from 'fetch-cookie';
import originalFetch from 'node-fetch';

const fetch = createFetchCookie(originalFetch);

(async () => {
  const res = await fetch(url, {
    method: 'POST',
    body: 'abc',
    interfaces: {
      Blob,
      fetch // this line might not even be necessary
    }
  });
  const blob = await res.blob();
  readBlob(blob).then(buf => /* Buffer.isBuffer(buf) */);
})();
```

bitinn:
> The approach I have on my mind is to exports an additional createFetch() function that allows users to bind specific available features.

I agree this would be the superior approach, but significant work needs to be done to support this modularize approach. Anyone up to the task?

(Also color me a bit wary of adding features most don't use)

TimothyGu:
Closing, as we need a bit more time to figure out the best way to tackle this issue.

PR #231, opened 2017-02-07
Fix code error

haocenxu:


ahmadnassri:
this is not an "error", rather a syntax choice, if the project requires a linting step, there are plenty of options to adopt. suggest closing this as un-needed change.

PR #236, opened 2017-02-16
feat(http.Agent): allow passing agent option as an object

### LINKED ISSUE #15: Allow to pass all available https options

ahmadnassri:
allows for controlling `http.Agent` without manually creating a new external object, this is useful

to prevent "requiring" `http`/`https` in SPA and isomorphic apps

Fixes #15

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1319,6 +1319,20 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should create http.Agent class if options.agent is provided as an object', function() {
+		url = `${base}inspect`;
+		opts = {
+			agent: {
+				keepAlive: true
+			}
+		};
+		return fetch(url, opts).then(res => {
+			return res.json();
+		}).then(res => {
+			expect(res.headers['connection']).to.equal('keep-alive');
+		});
+	});
+
 	it('should ignore unsupported attributes while reading headers', function() {
 		const FakeHeader = function () {};
 		// prototypes are currently ignored
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Note that only `method`, `headers`, `redirect` and `body` are allowed in `window
 	, compress: true     // support gzip/deflate content encoding. false to disable
 	, size: 0            // maximum response body size in bytes. 0 to disable
 	, body: empty        // request body. can be a string, buffer, readable stream
-	, agent: null        // http.Agent instance, allows custom proxy, certificate etc.
+	, agent: null        // http.Agent instance or options object, allows custom proxy, certificate etc.
 }
 ```
 
--- a/src/index.js
+++ b/src/index.js
@@ -39,12 +39,17 @@ export default function fetch(url, opts) {
 		const request = new Request(url, opts);
 
 		const options = getNodeRequestOptions(request);
+		const protocol = options.protocol === 'https:' ? https : http;
+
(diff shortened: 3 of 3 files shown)


test/test.js
```diff
@@ -1319,6 +1319,20 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should create http.Agent class if options.agent is provided as an object', function() {
```
TimothyGu: Is there a way to test `https.Agent` also?

ahmadnassri: yes, we can add an https specific test. 

I didn't think it was necessary since `https.Agent` inherits `http.Agent` and the logic is mainly concerned with initializing the object internally.

can add a test later today.

ahmadnassri: upon further study, I'm sticking with my initial comment, since `https.agent` inherits `http.agent` and the added logic is only concerned with initializing the objects.

I do think adding an `https.agent` test in general is great, but that is outside the scope of this PR and can be made into a separate branch & PR

TimothyGu: @ahmadnassri, agreed. I'll try to merge this PR later today. Thanks for your work!

ahmadnassri: thanks!

TimothyGu:
@ahmadnassri, actually, on a second thought I'm not sure if this is the best idea. After this patch, if an options object is passed, a new agent will be created on every fetch. I don't think we should encourage that.

ahmadnassri:
@TimothyGu I could see that concern. However it is also an optimization that can be left in user-land.

this change does not prevent performancr aware users from creating and using a global agent.

also, by creating different agents for each request it facilitates more user-level, fine-tune grained control, that would otherwise not be possible with a global agent.

TimothyGu:
Rebased and applied in ec29e3d2640586bc07cddf894e2aa10461b20ec0. Thanks!

Jakobud:
So, this is an old PR, but I see this was reverted. So does this no longer work? I can see that you can still pass an `agent` option into `fetch`. What does it do? Does this still allow passing an `https.Agent()` into it?

Also regarding the above PR, instead of creating an Agent for each request, you should have just had the Agent created when instantiating the client, and then just used that Agent on each request, so you aren't make a new one each time.

TimothyGu:
@Jakobud

> Does this still allow passing an `https.Agent()` into it?

Yes! We’ve always allowed this.

Jakobud:
Oh okay well then what code got reverted exactly? Automatic creation of an agent if an object is passed in?

TimothyGu:
@Jakobud 

> Automatic creation of an agent if an object is passed in?

Yes, that was reverted.

> Also, the agent is cached so that it doesn't get created over again for every request, correct?

We don't control the Agent object you pass in, so it is your responsibility to use the same Agent object yourself if you don't want to create it over again.

faceleg:
I've just run into a problem that this PR would solve - is there anything I could do to help get this merged?

jimmywarting:
@faceleg with v3 you can supply a function that can return an agent based on the protocol being used.
```js

fetch(url, {
  agent (url) { 
    return url.protocol === 'https:' ? httpsAgent : httpAgent
  }
})
```

PR #237, opened 2017-02-16
docs(defaults): list of default header values

### LINKED ISSUE #226: Document (or remove) user-agent header
It is not expected that the HTTP library is sending a user-agent header by default.

We only observed this when analysing the logs.

For others reference, `note-fetch` adds:

```
user-agent:       node-fetch/1.0 (+https://github.com/bitinn/node-fetch)
```

header to all requests, unless overwritten.

This can be a major privacy, security concern.

ahmadnassri:
Fixes #226

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,10 +193,10 @@ If no values are set, the following request headers will be sent automatically:
 
 Header            | Value                                                   
 ----------------- | --------------------------------------------------------
-`Accept-Encoding` | `gzip,deflate`                                          
+`Accept-Encoding` | `gzip,deflate` _(when `compress === true`)_             
 `Accept`          | `*/*`                                                   
 `Connection`      | `close`                                                 
-`Content-Length`  | (automatically calculated)                              
+`Content-Length`  | (automatically calculated, if possible)                 
 `User-Agent`      | `node-fetch/1.0 (+https://github.com/bitinn/node-fetch)`
 
 
--- a/README.md
+++ b/README.md
@@ -193,10 +193,10 @@ If no values are set, the following request headers will be sent automatically:
 
 Header            | Value                                                   
 ----------------- | --------------------------------------------------------
-`Accept-Encoding` | `gzip,deflate` _(when `compress === true`)_             
+`Accept-Encoding` | `gzip,deflate` _(when `options.compress === true`)_     
 `Accept`          | `*/*`                                                   
-`Connection`      | `close`                                                 
-`Content-Length`  | (automatically calculated, if possible)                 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -187,6 +187,18 @@ Note that only `method`, `headers`, `redirect` and `body` are allowed in `window
 }
 ```
 
+#### Default Headers
+
+If no values are set, the following request headers will be sent automatically:
+
+Header            | Value                                                   
+----------------- | --------------------------------------------------------
+`Accept-Encoding` | `gzip,deflate`
```
TimothyGu: This is done only when `compress` is true: https://github.com/bitinn/node-fetch/blob/560d5d6a023f069c37497f633fe65955182fa3d5/src/request.js#L145-L147

README.md
```diff
@@ -187,6 +187,18 @@ Note that only `method`, `headers`, `redirect` and `body` are allowed in `window
...
+#### Default Headers
+
+If no values are set, the following request headers will be sent automatically:
+
+Header            | Value                                                   
+----------------- | --------------------------------------------------------
+`Accept-Encoding` | `gzip,deflate`                                          
+`Accept`          | `*/*`                                                   
+`Connection`      | `close`                                                 
+`Content-Length`  | (automatically calculated)
```
TimothyGu: (automatically calculated, if possible)

README.md
```diff
@@ -187,6 +187,18 @@ Note that only `method`, `headers`, `redirect` and `body` are allowed in `window
...
 
+#### Default Headers
+
+If no values are set, the following request headers will be sent automatically:
+
+Header            | Value                                                   
+----------------- | --------------------------------------------------------
+`Accept-Encoding` | `gzip,deflate`                                          
+`Accept`          | `*/*`                                                   
+`Connection`      | `close`
```
TimothyGu: Specify that this is set automatically only if the user is not using an agent (https://github.com/bitinn/node-fetch/blob/560d5d6a023f069c37497f633fe65955182fa3d5/src/request.js#L148)

ahmadnassri:
will make appropriate changes as suggested :+1:

README.md
ahmadnassri: I feel that this is clearly implied (by declaring "default" in the section title)

TimothyGu: @ahmadnassri, no, the default only implies that it is set automatically in all cases when the header is not explicitly set. This header is special in that it is set automatically, when the header isn't explicitly set **and** `agent` isn't set.

ahmadnassri: I'm mistaken, somehow I conflated this with `User-Agent`! my bad.

PR #239, opened 2017-02-22
Decompress slightly invalid gzip response

gregory90:
Prevents following error when compressed response is slightly invalid/truncated (it happens rarely, but still)
```
Error: unexpected end of file
  File "zlib.js", line 370, col 17, in Zlib._handle.onerror
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -148,9 +148,18 @@ export default function fetch(url, opts) {
 			// otherwise, check for gzip or deflate
 			let name = headers.get('content-encoding');
 
+			// Be less strict when decoding compressed responses, since sometimes
+			// servers send slightly invalid responses that are still accepted
+			// by common browsers.
+			// Always using Z_SYNC_FLUSH is what cURL does.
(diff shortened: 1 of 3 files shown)


bitinn:
Note, the original handling of raw inflate response is a hack for older Microsoft IIS/Apache servers, not required by spec, and not for specific websites (I waited a longtime before adding the hack).

I want to know which old servers would intentionally send broken gzip? If it's pretty widespread we would consider it. Otherwise we think the site owner is better off fixing them.

(From my phone)

gregory90:
Some reference:
https://github.com/nodejs/node/issues/8701
https://github.com/request/request/pull/2492

I'm seeing these errors sometimes, and did get to issues mentioned above. node-fetch doesn't have this implemented. If that's not what you want, don't hesitate to close this PR, but many clients are using this approach. I'm not an expert by any means though:)

TimothyGu:
Merged in a1e76b9. Thanks!

PR #240, opened 2017-02-25
transpilling to Node 4 instead of dropped 0.x

tinovyatkin:
Your docs says that `2.x` dropping support for `node 0.x` while Babel transpilling still targets that version. This PR fixes that as well as dropping `0.x` from travis, updates `babel-preset-env` to latest version and adding appropriate `engines` to `package.json`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
   "env": {
     "test": {
       "presets": [
-        [ "env", { "loose": true, "targets": { "node": 0.12 } } ]
+        [ "env", { "loose": true, "targets": { "node": 4 } } ]
       ],
       "plugins": [
         "transform-runtime",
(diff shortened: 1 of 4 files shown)


bitinn:
We have only dropped `0.10.x`

tinovyatkin:
Why somebody who doesn't want to upgrade a `node` from 5 versions back `0.x` will want to upgrade `node-fetch` to next major version?

TimothyGu:
I'm fine with dropping 0.12.x as well, since it ran out of service at the end of last year.

TimothyGu:
@tinovyatkin, please check the [coverage report](https://codecov.io/gh/bitinn/node-fetch/pulls/240/changes) and remove a Node.js v0.12-specific hack.

jimmywarting:
🎉  Finally, RIP 0.x

bitinn:
I agree 0.12 can be dropped. (Though I should note the reason we target 0.12.x was because `2.x` was schedule to release in 2016, but anyway.)

tinovyatkin:
So, it's hopefully will be merged somewhere soon? I have another PR waiting (removing a dependency)...

PR #244, opened 2017-02-27
adopting style auto enforcement configs

tinovyatkin:
I'm sure that any repo with more than one contributor must have code style policy enforced. As this module is spec compliant I think the best possible style guide is `node` repo itself, so, I adopted `eslint` and `editorconfig` files from node repository as a starting point.
It's not completely following current style, as most source files are using tabs instead of spaces, but I'm not sure what part is better to change. Anyway, I think the best way is to introduce these gradually.
Having these files in project will also help to prevent autoformating at contributors IDE setup.


TimothyGu:
Currently the code base is mixed: files in `src/` are indented with tabs (which is what @bitinn prefers IIRC) but other files like `.travis.yml` are done with 2 spaces (though indentation of `package.json` is forced).

I personally prefer indentation with 2 spaces, which seems to be the norm in most of the JS world. If we were to switch to tabs wholesale, we might still want to specify a `indent_size = 2` so that tabs don't take up 8 character-spaces each.

.eslintrc.yaml
```diff
@@ -0,0 +1,132 @@
...
+  constructor-super: 2
+  no-class-assign: 2
+  no-confusing-arrow: 2
+  no-const-assign: 2
+  no-dupe-class-members: 2
+  no-new-symbol: 2
+  no-this-before-super: 2
+  prefer-const: [2, {ignoreReadBeforeAssign: true}]
+  rest-spread-spacing: 2
+  template-curly-spacing: 2
```
TimothyGu: I feel we should be using some pre-established ESLint config rather than rolling our own, just so that it'd be easier for new developers to get started. I've had good experiences with `xo` which supports both spaces and tabs.

bitinn:
I usually do tab with indent size 4. I know, it's not exactly the de facto style in nodejs world, but I want to see how everyone feels here. Is using 2 spaces a big deal for new contributors? If so, I can live with using 2 spaces.

I personally feel ok as long as the `src/` JS files are consistently formatted, I don't have a problem with yml/json files using other formatting/indent style.

tinovyatkin:
I'd love to use 2 spaces, but the biggest deal is when `eslint/editorconfig` is missing at all - as many IDE starts to reformat code to a developer default.
As I stated above, I've pulled `node` repo code convention because I think that a ultimate target for any standard compliant module is to be merged into `node` repo, so, it's good to be prepared even it's not an idea :) However, I can switch to just `eslint:recommended` if you feel that better.

tinovyatkin:
OK, in this case it will not affect hard decisions of refactoring tabs to spaces :), but will at least create a foundation and prevent autoformating by IDE

tinovyatkin:
I have a spec breaking test waiting for PR and want either merge or close this, please give your opinion...

PR #253, opened 2017-03-18
fix: do not re-initialize Headers (fixes #251)

### LINKED ISSUE #251: How to get an array of header values now that getAll has been removed?
> Major: remove headers.getAll(); make get() return all headers delimited by commas (per spec)

Server response (`set-cookie`) might include a comma, e.g.

```
Set-Cookie:              ASP.NET_SessionId=4pxd5fecgvg2vaq1nyetxk5z; domain=foo.com; path=/; HttpOnly
Set-Cookie:              SC_ANALYTICS_GLOBAL_COOKIE=a17765ea49144afdb045dac1294323ca|False; domain=foo.com; expires=Wed, 17 Mar
                         2027 17:12:37 -0000; path=/; HttpOnly
Set-Cookie:              UserSessionId=7c47a144-6a12-4ffc-907a-db8a281cca6a|10002|181845|161471|2017-03-17; domain=foo.com; path=/;
                         HttpOnly
```

How do I get individual `set-cookie` values? Breaking by comma as suggested in the change log note is going to produce malformed results.

gajus:
Without this change the headers get flattened in [`/src/headers.js#L42-L70`](https://github.com/bitinn/node-fetch/blob/a1e76b97e18f28395bdb46e3a849800f52ee8027/src/headers.js#L42-L70). This is a result of an iterator [`/src/headers.js#L197-L199`](https://github.com/bitinn/node-fetch/blob/a1e76b97e18f28395bdb46e3a849800f52ee8027/src/headers.js#L197-L199) using getter [`/src/headers.js#L89-L96`](https://github.com/bitinn/node-fetch/blob/a1e76b97e18f28395bdb46e3a849800f52ee8027/src/headers.js#L89-L96) that joins multiple header values into a comma separated list.

### DIFF
--- a/src/response.js
+++ b/src/response.js
@@ -23,7 +23,12 @@ export default class Response {
 		this.url = opts.url;
 		this.status = opts.status || 200;
 		this.statusText = opts.statusText || STATUS_CODES[this.status];
-		this.headers = new Headers(opts.headers);
+
+		if (opts.headers instanceof Headers) {
+			this.headers = opts.headers;
+		} else {
+			this.headers = new Headers(opts.headers);
+		}
 
 		Object.defineProperty(this, Symbol.toStringTag, {
 			value: 'Response',
--- a/src/response.js
+++ b/src/response.js
@@ -24,11 +24,7 @@ export default class Response {
 		this.status = opts.status || 200;
 		this.statusText = opts.statusText || STATUS_CODES[this.status];
 
-		if (opts.headers instanceof Headers) {
-			this.headers = opts.headers;
-		} else {
-			this.headers = new Headers(opts.headers);
-		}
+		this.headers = new Headers(opts.headers);
 
 		Object.defineProperty(this, Symbol.toStringTag, {
 			value: 'Response',
--- a/src/headers.js
+++ b/src/headers.js
@@ -34,6 +34,16 @@ export default class Headers {
 	constructor(init = undefined) {
 		this[MAP] = Object.create(null);
 
+		if (init instanceof Headers) {
+			for (const [key, values] of Object.entries(init.raw())) {
+				for (const value of values) {
+					this.append(key, value);
+				}
+			}
+
+			return;
+		}
+
 		// We don't worry about converting prop to ByteString here as append()
 		// will handle it.
 		if (init == null) {
--- a/src/headers.js
+++ b/src/headers.js
@@ -35,9 +35,12 @@ export default class Headers {
 		this[MAP] = Object.create(null);
 
 		if (init instanceof Headers) {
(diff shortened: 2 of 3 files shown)


src/response.js
```diff
@@ -23,7 +23,12 @@ export default class Response {
 		this.url = opts.url;
 		this.status = opts.status || 200;
 		this.statusText = opts.statusText || STATUS_CODES[this.status];
-		this.headers = new Headers(opts.headers);
+
+		if (opts.headers instanceof Headers) {
+			this.headers = opts.headers;
+		} else {
+			this.headers = new Headers(opts.headers);
+		}
```
TimothyGu: This goes against the Fetch Standard which explicitly says to make a copy of the headers ([<code>Response(_body_, _init_)</code> constructor](https://fetch.spec.whatwg.org/#dom-response), step 6). Instead, you should add a special case in the `Headers` constructor. It should provide a performance boost as well.

gajus: Updated the pull request.

bitinn:
Note: this PR breaks older node environment, the codecov report is outdated.

TimothyGu:
Other than the use of `Object.entries`, the rest LGTM.

gajus:
> Other than the use of Object.entries, the rest LGTM.

Whats wrong with `Object.entries`?

TimothyGu:
It is [not supported](http://node.green/#ES2017-features-Object-static-methods-Object-entries) on older versions of Node.js.

gajus:
> It is not supported on older versions of Node.js.

You are using `babel-preset-env` that transpiles `Object.entries` down to ES3.

bitinn:
@gajus if you are seeing something wrong with our travis integration/transpiler config that breaks your valid code then please fix it.

gajus:
> @gajus if you are seeing something wrong with our travis integration/transpiler config that breaks your valid code then please fix it.

Right. Sorry. I didn't notice that Travis was failing in the first place.

Upon further inspection, it looks like Babel does not transpile `Object.entries`. I thought it does because Google gave me https://www.npmjs.com/package/babel-plugin-transform-es2017-object-entries result, which appears as an official package at first sight. It isn't.

Let me update the PR using `Object.keys`.

TimothyGu:
Thanks a lot!

gajus:
@TimothyGu Please release the patch.

gajus:
@TimothyGu It is a bit of a joke to contribute something if it doesn't get released for 2 months.

bitinn:
@gajus I apologise for the delay, would you consider collaborating on the 2.x release?

https://github.com/bitinn/node-fetch/issues/252

PR #256, opened 2017-03-27
Pass family and localAddress props to http / https module

ewoerner:
Passes these two attributes through to the underlying http / https module:
* **family:** IP address family to use when resolving host and hostname. Valid values are 4 or 6. When unspecified, both IP v4 and v6 will be used.
* **localAddress:** Local interface to bind for network connections

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -77,6 +77,8 @@ export default class Request {
 			input.compress : true;
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
+		this.family = init.family || input.family;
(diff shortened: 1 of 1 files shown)


bkw:
You may want to add these to the options of the request issued for redirects as well (index.js).

TimothyGu:
Superseded by #474.

Legogris:
We have a use-case where we need to set `localAddress`. Since #474 is now closed in favor of #476, would you be open to a PR with adding just the `localAddress` part of this @TimothyGu ?

PR #257, opened 2017-04-01
headers.get: add space to join

zkat:
See spec for combined values: https://fetch.spec.whatwg.org/#concept-header-value-combined

I don't know if a different spec clause overrides the need for that space.


TimothyGu:
It was recently changed in whatwg/fetch#504 it seems.

PR #264, opened 2017-04-09
use browser natives when bundled for the browser

m59peacemaker:
This makes `node-fetch` reference the browser native fetch API when bundled for the browser.

https://github.com/bitinn/node-fetch/issues/262

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,0 +1,4 @@
+module.exports = exports = window.fetch;
+exports.Headers = window.Headers;
+exports.Request = window.Request;
+exports.Response = window.Response;
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "2.0.0-alpha.3",
   "description": "A light-weight module that brings window.fetch to node.js",
   "main": "lib/index.js",
+  "browser": "./browser",
   "module": "lib/index.es.js",
   "files": [
     "lib/index.js",


bitinn:
While this makes sense, I just wonder why you didn't opt for https://github.com/matthew-andrews/isomorphic-fetch ?

I always think `node-fetch` shouldn't assume browser environment (whether you need polyfill etc.), so I need some information on whether this may cause unintended problems.

m59peacemaker:
isomorphic-fetch is a polyfill when bundled in the browser. Unless I'm missing something, `node-fetch` can't possibly be used in a browser, so if someone were to bundle it for the browser, it will be creating a bundle that is just going to throw an error and not ever do anything useful for fetching. Using the browser field and exporting the native `fetch` is changing out the "throw an error" situation for "use the native fetch and work as expected". No loss, only gain. If this isn't behavior that interests you, I'm going to make a repo that just requires `node-fetch` and does this same thing for the browser. What's lame about that is that I would have to keep it up to date with every release of node-fetch. It's very awkward, considering node-fetch could just do this itself. Like I said on the other thread, `form-data`, which node-fetch depends on, does this already and it makes perfect sense to do so.

The only unfortunate thing here is that `Headers`, `Response`, etc become properties of `fetch` because of commonjs. As in - the native / global fetch object gets properties added to it. If you'll use the browser field, we should figure out a better approach. Maybe:

- browser
  - fetch.js
  - headers.js
  - request.js
  - response.js

```json
"browser": {
  "./index.js: "./browser/fetch.js",
  "./lib/headers.js: "./browser/headers.js",
  "./request.js: "./browser/request.js",
  "./response.js: "./browser/response.js",
}
```

```js
const Fetch = require('node-fetch')
const Headers = require('node-fetch/lib/headers')
const Request = require('node-fetch/lib/request')
const Response = require('node-fetch/lib/response')
```

It's not perfect, but either way is better than nothing.

m59peacemaker:
I just found that this hac come up before: https://github.com/bitinn/node-fetch/pull/89

bitinn:
@TimothyGu I am leaning towards merging this. Could you double check it doesn't have much drawback?

(I don't see any drawback myself and the problem does pops up quite often when people try to build with node-fetch)

TimothyGu:
@bitinn, none that I am aware of. I do prefer the property name to contain `.js` suffix though, other than the nitpick LGTM.

bitinn:
Alright I will merge and fix it on GitHub, hope you don't mind :)

PR #265, opened 2017-04-12
expose url property on request object

m59peacemaker:
In browsers, the `url` property appears when iterating over the properties of an instance of Request. I think that is the proper behavior. `node-fetch` is currently accessing `url` with a getter, so it is hidden when iterating the properties or logging the object. I removed the getter, set `this.url` in the constructor, and added a test that `url` is exposed. All the tests continue to pass after my change with no additional changes needed. I'm unsure why it was a getter, as no tests depended on it being so.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -79,16 +79,14 @@ export default class Request {
 		this.agent = init.agent || input.agent;
 
 		this[PARSED_URL] = parsedURL;
+		this.url = format_url(parsedURL);
+
 		Object.defineProperty(this, Symbol.toStringTag, {
 			value: 'Request',
 			writable: false,
 			enumerable: false,
 			configurable: true
 		});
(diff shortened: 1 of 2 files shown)


bitinn:
Spec wise, they are all readonly, so using a getter is correct:

https://fetch.spec.whatwg.org/#request-class

(in fact we should turn other properties to readonly getter as well)

On the other hand, I can't find a word on whether Request object should be explicit enumerable. So question here: https://github.com/whatwg/fetch/issues/525

bitinn:
Hi @m59peacemaker, because change needed to proceed (see my reply) is quite different from your current PR, I would think it's best to re-open a new PR for this change.

I will open a ticket for this issue and you may submit a PR against it if you have time, thx!

PR #267, opened 2017-04-30
Document response.ok convenience property

AlexTes:
Just noticed this was left out while looking through a response instance. Outside of whether I think this property is useful or not I think we can agree it's good to document our public properties 😄.

Thanks for a great lib!


bitinn:
Again, thx for adding that.

PR #269, opened 2017-05-11
Small improvements to the error-handling wording

AlexTes:
Sidenote. I feel this sentence:
> All errors originating from Node.js core are marked with `err.type = 'system'`, and in addition contain an `err.code` and an `err.errno` for error handling. These are aliases for error codes thrown by Node.js core.

Could be a little more clear on the distinction between `code` and `errno`, or at least link to the relevant Node documentation. I'd add it myself but I'm not clear on the distinction myself 😅 .

Thanks for a great lib!

### DIFF
--- a/ERROR-HANDLING.md
+++ b/ERROR-HANDLING.md
@@ -2,17 +2,17 @@
 Error handling with node-fetch
 ==============================
 
-Because `window.fetch` isn't designed to transparent about the cause of request errors, we have to come up with our own solutions.
+Because `window.fetch` isn't designed to be transparent about the cause of request errors, we have to come up with our own solutions.
 
 The basics:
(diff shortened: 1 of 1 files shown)


TimothyGu:
LGTM. I must have been sleeping the last time I edited these docs 😟, so thanks a lot!

The relevant Node.js documentation should be found here: https://nodejs.org/api/errors.html#errors_system_errors. Let me know if that's what you are looking for.

bitinn:
LGTM, so I merged it. Thx for the check!

AlexTes:
@TimothyGu No problem 😄 . Been there.
Ah, those core Error properties are quite easy to find.

Cheers!

PR #285, opened 2017-06-03
Add the missing comma

chinesedfan:
Just to keep all codes the same.


bitinn:
Thx, I will merge it this time. But 1.x is a backport maintenance branch, so we generally only accept important bugfix.

Issue #210, opened 2016-12-05
body conversion


jimmywarting:
Haven't tested v2 if this works but i would like for this to work:
It currently doesn't in 1.x

```js
stream = new Response('hi').body
str = await new Response(Uint8Array.from([97])).text() // a
stream = new Response(Uint8Array.from([97])).body // returns a Uint8array [97] (should return a stream)
```

Ideally every acceptable thing you put into Response should work with<br> `arraybuffer() json() text() and .body`

TimothyGu:
To implement this efficiently one would need to add a `body` getter function, so as to prevent having to go through the process of creating streams when the body is read synchronously anyway.

@bitinn, what are your thoughts on this?

bitinn:
I will say we hold off from this feature for now. We will get around to it when available.

garthk:
Fixing this _might_ justify a major semver bump. It wouldn't surprise me if people who took a dependency on `node-fetch` directly or via `isomorphic-fetch` somehow rely on `Response.body` being whatever was passed to the `Response` constructor, not necessarily a stream.

Arguing against that, for a sec: most customers wouldn't follow us across a major semver bump, and would miss out on bug and security fixes unless we back-ported them, which might require more effort than we've got spare.

I worked around it in fetch-hooks by wrapping a `Proxy` around the final `Response`, modifying `body`, and binding all methods to the original object so their view of `this.body` remained unchanged.

jimmywarting:
This have been hold of for a long time now

bitinn:
With #547 merged I think we are now a step towards making `req.body` and `req.body` more consistent even when you create it with Request/Response constructor.

I don't currently have time to make the rest of conversion, @jimmywarting let me know if you are up to the task.

I also think the next release should trigger a major version bump (and release into `next` branch) due changes on Request/Response constructor (even though I expect few to depend on it).

Given there are quite a few issues pending for V3 release, it's now a good time to fix this problem along with them.

jimmywarting:
I'm up for it

bitinn:
@jimmywarting just `master`,  I have created a new branch `2.x` with latest code `2.3.0`, but not our last commit.

xxczaki:
@jimmywarting As it is in our [v3 Roadmap](https://github.com/bitinn/node-fetch/issues/668), wouldn't you mind to implement it in your free time? Thanks ;)

jimmywarting:
[...]

PR #294, opened 2017-06-07
Adds arrayBuffer method

tomas-polach:
I noticed that `arrayBuffer`was missing from body in 1.x.
I think #210 talked about this as well so I thought I contribute back :)
Let me know what you think.

### DIFF
--- a/lib/body.js
+++ b/lib/body.js
@@ -76,6 +76,19 @@ Body.prototype.buffer = function() {
 };
 
 /**
+ * Decode response as arrayBuffer
+ *
+ * @return  Promise
+ */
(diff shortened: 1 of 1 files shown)


bitinn:
Thx, but a few notes on why you should open another PR:

- Don't make changes on your `1.x` branch directly, that will make syncing it back very difficult. (`1.x` is like a legacy `master` branch).
- You need to add testcases for it.

gr2m:
looks like `node-fetch@v2.0.0` has `body.arrayBuffer()` implemented now: https://github.com/bitinn/node-fetch#bodyarraybuffer

PR #297, opened 2017-06-11
Accept URLSearchParams as body and detect for setting form-urlencoded header

### LINKED ISSUE #296: Set Content-Type to 'application/x-www-form-urlencoded' when body is an instance of url.URLSearchParams
Seting `Content-Type` to 'application/x-www-form-urlencoded' when body is an instance of `url.URLSearchParams` would match spec as per https://fetch.spec.whatwg.org/#concept-bodyinit-extract 

As far as I can tell, this is not something that `node-fetch` currently does.. and it caught me up on what I consider to be a somewhat obscure bug for 30 minutes.

Interestingly (although possibly unrelated), `curl` sets this by default when using a simple data cli arg (-d foo=bar) as opposed to using text/plain, but I guess we're more interested in the fetch spec than how curl works.

If this is an acceptable proposal, I can get the PR put together in a fairly short time.

Edit: fixed incorrect whatwg link anchor

jkantr:
This fixes #296

- [x] tests with 100% branch coverage
- [x] readme updated

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -353,7 +353,7 @@ export function extractContentType(instance) {
 		return 'text/plain;charset=UTF-8';
 	} else if (isURLSearchParams(body)) {
 	 	// body is a URLSearchParams
-		return 'application/x-www-form-urlencoded';
+		return 'application/x-www-form-urlencoded;charset=UTF-8';
 	} else if (body instanceof Blob) {
 		// body is blob
 		return body.type || null;
@@ -380,6 +380,9 @@ export function getTotalBytes(instance) {
 	} else if (typeof body === 'string') {
 		// body is string
 		return Buffer.byteLength(body);
+	} else if (isURLSearchParams(body)) {
+		// body is URLSearchParams
+		return Buffer.byteLength(String(body));
 	} else if (body instanceof Blob) {
 		// body is blob
 		return body.size;
@@ -410,6 +413,10 @@ export function writeToStream(dest, instance) {
 		// body is string
 		dest.write(body);
 		dest.end();
+	} else if (isURLSearchParams(body)) {
+		// body is URLSearchParams
+		dest.write(new Buffer(String(body)));
+		dest.end()
 	} else if (body instanceof Blob) {
 		// body is blob
 		dest.write(body[BUFFER]);
--- a/src/body.js
+++ b/src/body.js
@@ -30,6 +30,8 @@ export default function Body(body, {
 		body = null;
 	} else if (typeof body === 'string') {
 		// body is string
+	} else if (isURLSearchParams(body)) {
+		// body is a URLSearchParams
 	} else if (body instanceof Blob) {
 		// body is blob
 	} else if (Buffer.isBuffer(body)) {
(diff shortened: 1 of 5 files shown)


src/body.js
```diff
@@ -324,6 +351,9 @@ export function extractContentType(instance) {
 	} else if (typeof body === 'string') {
 		// body is string
 		return 'text/plain;charset=UTF-8';
+	} else if (isURLSearchParams(body)) {
+	 	// body is a URLSearchParams
+		return 'application/x-www-form-urlencoded';
```
TimothyGu: The spec specifies `application/x-www-form-urlencoded;charset=UTF-8`

src/request.js
```diff
@@ -63,6 +63,9 @@ export default class Request {
 
 		if (init.body != null) {
 			const contentType = extractContentType(this);
+			if (contentType === 'application/x-www-form-urlencoded') {
+				this.body = String(this.body);
+			}
```
TimothyGu: This belongs to `consumeBody()` and `writeToStream()` in `src/body.js`, where the appropriate actions for other types of `body`types are listed as well. You might also have to implement `getTotalBytes()` there.

test/test.js
```diff
@@ -853,6 +855,67 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should allow POST request with URLSearchParams as body', function() {
```
TimothyGu: You can only enable these tests on Node.js v7.0.0+. To do so you could put

```js
const itUSP = typeof URLSearchParams === 'function' ? it : it.skip;
```

and use `itUSP` instead of `it` here.

test/test.js
```diff
@@ -853,6 +855,67 @@ describe('node-fetch', () => {
...
+			return res.json();
+		}).then(res => {
+			expect(res.method).to.equal('POST');
+			expect(res.headers['content-type']).to.equal('application/x-www-form-urlencoded');
+			expect(res.headers['content-length']).to.equal('3');
+			expect(res.body).to.equal('a=1');
+		});
+	});
+
+	it('should still recognize URLSearchParams when extended', function() {
```
TimothyGu: ditto

src/body.js
```diff
@@ -380,6 +413,10 @@ export function writeToStream(dest, instance) {
 		// body is string
 		dest.write(body);
 		dest.end();
+	} else if (isURLSearchParams(body)) {
+		// body is URLSearchParams
+		dest.write(new Buffer(String(body)));
```
TimothyGu: `new Buffer` is deprecated. Use `Buffer.from` instead.

src/body.js
```diff
@@ -380,6 +413,10 @@ export function writeToStream(dest, instance) {
...
+		dest.end()
```
TimothyGu: Semicolon...

src/body.js
jkantr: So I was actually going to submit another PR after this to replace all of the new Buffers... I didn't want to mix concerns, hehe. But yes, I understand now... don't do bad stuff in the new code :p

src/body.js
jkantr: y is there no .eslintrc ? :p

PR #298, opened 2017-06-15
refactor deprecated `new Buffer` with Buffer.from and Buffer.alloc

jkantr:
My only question is whether we should be creating the buffer from the #buffer prop in the case of an ArrayBuffer (like from a Uint8Array) because it's a read only reference. I don't completely understand the implications here in the case of Blob so I just left the deprecation fix that keeps the most consistent behavior to how it was previously.

### DIFF
--- a/src/blob.js
+++ b/src/blob.js
@@ -31,13 +31,13 @@ export default class Blob {
 				if (element instanceof Buffer) {
 					buffer = element;
 				} else if (ArrayBuffer.isView(element)) {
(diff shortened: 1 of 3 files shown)


jkantr:
I messed up the git history with another PR by accident, so codecov report is incorrect... will reopen

TimothyGu:
Don't worry about codecov. If you force push a new commit it should fix itself.

PR #299, opened 2017-06-19
refactor deprecated `new Buffer` with Buffer.from and Buffer.alloc

jkantr:
Just need a sanity check:
blob.js line 34... Buffer.from(ArrayBuffer) or Buffer.from(ArrayBuffer#buffer) ?

### DIFF
--- a/src/blob.js
+++ b/src/blob.js
@@ -31,9 +31,9 @@ export default class Blob {
 				if (element instanceof Buffer) {
 					buffer = element;
 				} else if (ArrayBuffer.isView(element)) {
-					buffer = Buffer.from(new Uint8Array(element.buffer, element.byteOffset, element.byteLength));
+					buffer = Buffer.from(element.buffer, element.byteOffset, element.byteLength);
 				} else if (element instanceof ArrayBuffer) {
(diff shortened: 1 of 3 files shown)


src/blob.js
```diff
@@ -31,13 +31,13 @@ export default class Blob {
 				if (element instanceof Buffer) {
 					buffer = element;
 				} else if (ArrayBuffer.isView(element)) {
-					buffer = new Buffer(new Uint8Array(element.buffer, element.byteOffset, element.byteLength));
+					buffer = Buffer.from(new Uint8Array(element.buffer, element.byteOffset, element.byteLength));
```
TimothyGu: `Buffer.from(element.buffer, element.byteOffset, element.byteLength)` should be enough. This way you are not copying the TypedArray twice (here and in the `Buffer.concat` below).

src/blob.js
```diff
@@ -31,13 +31,13 @@ export default class Blob {
...
 				} else if (element instanceof ArrayBuffer) {
-					buffer = new Buffer(new Uint8Array(element));
+					buffer = Buffer.from(new Uint8Array(element));
```
TimothyGu: Same here: `Buffer.from(element)`

PR #302, opened 2017-06-21
remove `encoding` as a dependency, throw descriptive Error

### LINKED ISSUE #273: Make dependency on encoding optional
`node-fetch` has 1 dependency that's [encoding](https://github.com/andris9/encoding), which is really legacy dependency from 1.x era (see [upgrade guide](https://github.com/bitinn/node-fetch/blob/master/UPGRADE-GUIDE.md#text-no-longer-tries-to-detect-encoding)).

My problem with `encoding` is it depends on [iconv-lite](https://github.com/ashtuchkin/iconv-lite), which is kinda huge in size because of what it needs to do.

So given many aren't using such features, I would consider dropping this hard dependency and throw errors when calling `textConverted` without `encoding` installed.

jkantr:
Fixes #273

So the code change itself is quite simple.

Trying to write a test to maintain 100% codecov with `encoding` as Schrodinger's package, however, proved interesting. I'm somewhere between 'proud' and 'horrified' with the result.

Shifted `encoding` from deps to dev-deps.


src/body.js
```diff
@@ -231,6 +232,10 @@ function consumeBody(body) {
  * @return  String
  */
 function convertBody(buffer, headers) {
+	if (typeof convert !== 'function') {
+		throw new FetchError('The package `encoding` must be installed to use the textConverted() function', 'missing-encoding');
```
bitinn: I think Error is better here as it's a programming error instead of operational error.

https://github.com/bitinn/node-fetch/blob/master/ERROR-HANDLING.md

test/test.js
```diff
@@ -1090,105 +1094,156 @@ describe('node-fetch', () => {
...
-		return fetch(url).then(res => {
-			expect(res.status).to.equal(200);
-			return res.text().then(result => {
-				expect(result).to.equal('<?xml version="1.0" encoding="EUC-JP"?><title>\ufffd\ufffd\ufffd\u0738\ufffd</title>');
+	describe('external encoding', () => {
+		const hasEncoding = typeof convert === 'function';
+
+		describe('with optional `encoding`', () => {
+			before(function() {
+				if (!hasEncoding) this.skip();
```
bitinn: not sure why this is necessary, could just use a flag to change the `expect` clause instead.

`skip` is not designed to be used this way, I think.

src/body.js
jkantr: Oh.. yes this makes sense. I'll adjust :)

test/test.js
jkantr: I... guess I was using skip for its intended purpose (to test the tests lol) but you're right, I shouldn't have kept it in the commit.

src/body.js
```diff
@@ -231,6 +232,10 @@ function consumeBody(body) {
  * @return  String
  */
 function convertBody(buffer, headers) {
+	if (typeof convert !== 'function') {
+		throw new Error('The package `encoding` must be installed to use the textConverted() function');
```
jkantr: Should I still be setting a 'type' property here on the Error, even tho it isn't an instance of FetchError? The error document was a little vague on that (It just said 'all errors originating from node-fetch' ...)

TimothyGu: Nah this is fine.

src/body.js
```diff
@@ -5,13 +5,14 @@
...
  */
 
-import {convert} from 'encoding';
 import Stream, {PassThrough} from 'stream';
 import Blob, {BUFFER} from './blob.js';
 import FetchError from './fetch-error.js';
 
 const DISTURBED = Symbol('disturbed');
 
+try { var convert = require('encoding').convert } catch(e) {}
```
TimothyGu: Please no implicit `var`... declare `convert` outside the `try`. (Also, use `let`.)

rollup.config.js
```diff
@@ -21,7 +21,5 @@ export default {
     if (isBuiltin(id)) {
       return true;
     }
-    id = id.split('/').slice(0, id[0] === '@' ? 2 : 1).join('/');
-    return !!require('./package.json').dependencies[id];
```
TimothyGu: Instead of doing this, why don't you explicitly whitelist `encoding`?

jkantr:
@TimothyGu @bitinn if there are no further comments I'll merge this :)

bitinn:
Now I have time to go through this PR, I have a few concerns:

- Why is `whatwg-url` added as a hard dependency?
- I think you really shouldn't run the with/without `encoding` test in a single test run. The preferable way should be to run the full test suites without `encoding` package, and then run it again with `encoding`.
- Yes it sounds like we are making test suites run more than it should, but this is exactly what Travis CI is here for. You want fast test run, instead of `this.timeout(12000);`

bitinn:
For reference, please take a look at this file:

- https://github.com/bitinn/node-fetch/blob/master/.travis.yml

Also for clarity sake:

- You may want to move these test cases for `encoding` outside of the main `describe` block, and use it's own block instead, because it's really a backward compatible test.
- You may want to rebase it when you merge.

jkantr:
Definitely didn't add `whatwg-url`, I think it just looks funny in the diff because `encoding` got shifted to dev-deps from deps.

I'm totally with you on the test thing. I have limited FOSS experience so I'm not up to speed completely on Travis. If it's just as or more trivial to runa separate test in an alternate environment with or without `encoding`, that's probably highly preferred. I'd just need a small nudge in the right direction as to implementation.

TimothyGu:
> because `encoding` got shifted to dev-deps from deps.

It should be less confusing if you can alphabetize the list.

jkantr:
Is this an okay way to add that env var in travis? This way doesn't seem to matrix each possibility (each version of form-data with and without encoding).. is that important to test or you think no?

.travis.yml
```diff
@@ -6,8 +6,10 @@ node_js:
 env:
   - FORMDATA_VERSION=1.0.0
   - FORMDATA_VERSION=2.1.0
+  - ENCODING=yes
 before_script:
   - 'if [ "$FORMDATA_VERSION" ]; then npm install form-data@^$FORMDATA_VERSION; fi'
+  - 'if [ "$ENCODING" = "yes" ]; then npm install encoding; fi'
```
jkantr: Is this an okay way to add that env var in travis? This way doesn't seem to matrix each possibility (each version of form-data with and without encoding).. is that important to test or you think no?

TimothyGu: You could do

```yaml
env:
  - FORMDATA_VERSION=1.0.0
  - FORMDATA_VERSION=1.0.0 ENCODING=yes
  - FORMDATA_VERSION=2.1.0
  - FORMDATA_VERSION=2.1.0 ENCODING=yes
```

so that the entire matrix will run. See https://docs.travis-ci.com/user/environment-variables/.

test/server.js
```diff
@@ -2,9 +2,10 @@ import * as http from 'http';
 import { parse } from 'url';
 import * as zlib from 'zlib';
 import * as stream from 'stream';
-import { convert } from 'encoding';
 import { multipart as Multipart } from 'parted';
 
+try { var convert = require('encoding').convert } catch(e) {}
```
TimothyGu: Do what you did with `body.js`:

```js
let convert;
try { convert = require('encoding').convert; } catch (e) {}
```

test/test.js
```diff
@@ -16,6 +16,9 @@ import * as http from 'http';
 import * as fs from 'fs';
 import * as path from 'path';
 
+let convert;
+try { convert = require('encoding').convert } catch(e) { }
```
TimothyGu: Missing `;` after `convert = require('encoding').convert`.

src/body.js
```diff
@@ -5,13 +5,15 @@
...
 
-import {convert} from 'encoding';
 import Stream, {PassThrough} from 'stream';
 import Blob, {BUFFER} from './blob.js';
 import FetchError from './fetch-error.js';
 
 const DISTURBED = Symbol('disturbed');
 
+let convert;
+try { convert = require('encoding').convert } catch(e) {}
```
TimothyGu: Missing `;` after `convert = require('encoding').convert`.

bitinn:
@jkantr I have gone through this PR and it looks good. I would leave you with the privilege of merging it yourself :)

(is it a privilege or is it just work?)

schickling:
@bitinn can you please release a new alpha version including this PR?

bitinn:
@schickling released as `v2.0.0-alpha.6`, been busy, I apologise.

PR #308, opened 2017-07-07
explicitly exclude node 5 from engines directive

### LINKED ISSUE #266: slightly invalid gzip decompression fails on node 5
This test works on node 4, 6, 7, but fails on 5.

https://github.com/bitinn/node-fetch/blob/master/test/test.js#L516-L525

> Z_BUF_ERROR unexpected end of file

jkantr:
fixes #266

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "lib/index.es.js"
   ],
   "engines": {
-    "node": ">=4"
+    "node": "4.x || >=6.0.0"
   },
   "scripts": {
     "build": "cross-env BABEL_ENV=rollup rollup -c",


package.json
```diff
@@ -10,7 +10,7 @@
     "lib/index.es.js"
   ],
   "engines": {
-    "node": ">=4"
+    "node": "4.x || >=6.0.0"
```
bitinn: Since it's quite a hassle to test a warning message in CI, could you try use [n
](https://github.com/tj/n) to verify it produces the warning message correctly (use `npm install <folder>`)

Other than that, this change LGTM

bitinn:
I just tested this with node v5, no warning was printed.

And I am starting to believe this will never work in most scenario due a subtle npm v3 bug (which comes with all node v5 install)

ref: https://github.com/npm/npm/issues/10024

TimothyGu:
@bitinn :(

Well this is the most we can do for that particular Node.js version. I say we just merge it.

PR #311, opened 2017-07-11
add data uri support

caub:
Just proposing this, to behave like browsers fetch which can accept a data uri

not finished, but I don't have precise ideas on how to do

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,7 @@
     "url-search-params": "^0.10.0",
     "whatwg-url": "^5.0.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "data-urls": "^1.0.0"
+  }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@
  * All spec algorithm step numbers are based on https://fetch.spec.whatwg.org/commit-snapshots/ae716822cb3a61843226cd090eefc6589446c1d2/.
  */
 
+import parseDataURL from 'data-urls';
 import Body, { writeToStream, getTotalBytes } from './body';
 import Response from './response';
 import Headers, { createHeadersLenient } from './headers';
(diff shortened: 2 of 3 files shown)


TimothyGu:
`data` URLs are a can of worms as they are [notoriously underspecified](https://simonsapin.github.io/data-urls/). I would prefer to use an established parser on npm or something instead of rolling our own. https://www.npmjs.com/package/data-uri-to-buffer looks pretty nice.

Besides that, you have to construct a Response object, and return a promise from `fetch` that will resolve to that.

You also have to pass the `type` along to the Response object as a `Content-Type` header [per spec](https://fetch.spec.whatwg.org/#concept-scheme-fetch).

Handling non-`base64` URLs is needed too. See https://tools.ietf.org/html/rfc2397#section-2 on the format for `data` URLs in general, though data-uri-to-buffer will do this for you.

TimothyGu:
Oh, and you need to be careful to wrap all the errors that may occur in the URL-to-data process in a promise instead of throwing synchronously in `fetch()`.

caub:
Ok, great, thanks, will do, I'll try to behave like spec/browser

caub:
normally it's ok now

caub:
I know it's not very beautiful, but is it ok now?

TimothyGu:
I'd like to put this on hold because of https://github.com/whatwg/fetch/pull/579.

TimothyGu:
Looks like the standard change has landed. @caub, can you update this module to use https://www.npmjs.com/package/data-urls instead?

caub:
@TimothyGu yup, done

bitinn:
Is `data-urls` now a dependency instead of devDependency?

caub:
it fails on node4 because `data-urls` contains some destructuring https://github.com/jsdom/data-urls/blob/master/lib/parser.js#L3

might add rollup there so

PR #312, opened 2017-07-12
add test for fetch(urlObj)

caub:
initially I changed to `parse_url(input.url + '')` in request.js because

```js
const {URL} = require('url');
fetch(new URL('https://nodejs.org'));
```
throws: <rejected> TypeError: Parameter "url" must be a string, not object

but I guess the npm version is not up-to-date yet? because the tests passes here

so I didn't do it, and just added a test. Can the npm version be updated to avoid this bug?

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -10,7 +10,7 @@ import * as stream from 'stream';
 import resumer from 'resumer';
 import FormData from 'form-data';
 import URLSearchParams_Polyfill from 'url-search-params';
-import {parse as parseURL, URLSearchParams} from 'url';
+import {parse as parseURL, URLSearchParams, URL as URLNode} from 'url';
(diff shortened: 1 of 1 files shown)


caub:
https://runkit.com/caub/5978d1598a496c0011c15e33 demo, that shows that he throws with the npm dep

any update?

TimothyGu:
@caub URL object support was added in node-fetch v2.0.0-alpha.\*. The default installed version is node-fetch v1.7.1, which doesn't have URL object support.

caub:
Ah ok

As you're here:), Why using require for nodejs core modules? https://github.com/bitinn/node-fetch/commit/60cf26c2f3baf566c15632b723664b47f5b1f2db and not import, it looks inconsistent

PR #320, opened 2017-07-26
Update deps

TimothyGu:


TimothyGu:
Not sure why Travis fails. Might be an npm glitch. Let's try this again in a few hours.

bitinn:
Re-run the build, `whatwg-url` seem to have problem on node v4, looks like they only test latest node.

TimothyGu:
Ah, right. whatwg-url is v6.x+ only so we shouldn't upgrade that atm.

PR #324, opened 2017-08-08
Fix cannot receive http header array issue.

H1Gdev:
On [jest](https://www.npmjs.com/package/jest) test framework, cannot receive http header array.

Test project
https://github.com/H1Gdev/test_set_cookie

1. npm install
2. npm test

**Expect**
-> receive 'set-cookie' header.
**Actual**
-> cannot receive 'set-cookie' header.

### DIFF
--- a/lib/headers.js
+++ b/lib/headers.js
@@ -35,7 +35,7 @@ function Headers(headers) {
 		} else if (typeof headers[prop] === 'number' && !isNaN(headers[prop])) {
 			this.set(prop, headers[prop].toString());
 
-		} else if (headers[prop] instanceof Array) {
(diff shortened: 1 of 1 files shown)


jkantr:
@TimothyGu I don't think we use a fallback here. Isn't it just "more correct" to drop the `instanceof` check entirely and just use `Array.isArray()`?

Assuming the PR is okay to begin with (it seems to make sense to me?), having both seems redundant.

bitinn:
It's for the 1.x branch, I will look into it.

@H1Gdev could you change the line to just use `Array.isArray`? It should cover all cases by `instanceof`.

Issue #75, opened 2016-02-06
file:// URLs


mgax:
Are there plans to support local filesystem URLs?

Klortho:
+1
Really surprised to discover this doesn't work.

TimothyGu:
There might be some security implications.

bitinn:
It's not on my roadmap at the moment, `node-fetch` sits on server-side and local url usage is limited.

stevenvachon:
This should work as it works in the browser.

TimothyGu:
I guess we've decided that `file://` will not be supported.

bitinn:
Just to clarify, if this works consistently across major browsers we will re-open.

But more as a clear warning that we don't support `file://` protocol, just like we don't support relative url.

matthewadams:
@bitinn @TimothyGu I don't I understand the aversion to `file:///` urls.

If we're on the server side (using Node.js + Express), and I have to call another rest endpoint whose swagger is not easily obtainable, why can't I take care to put it under source control and simply use it instead of the undownloadable spec from the service I'm calling?

This limitation seems arbitrary.  Please clarify its justification.

TimothyGu:
@matthew-andrews That sounds like a problem with how node-fetch is used, not with node-fetch. FYI Node.js v8 will be released with support for file URLs in `fs` module methods.

jimmywarting:
The Response constructor can take a Readable stream so why not simply do something like this?

```js
let {Response} = fetch
let stream = fs.createReadStream('myfile.txt')
let promise = new Response(stream).text()
```

I don't think file protocol should be supported, sorry
I for once use node-fetch as a proxy, user can request any url they want with my api
wouldn't want them to get access to the filesystem...

stevenvachon:
I don't see why it can't be supported behind an option to enable it.

matthewadams:
_NB: I'm @matthewadams, not @matthew-andrews. Sorry for the noise, @matthew-andrews  _

@TimothyGu I'm confused by your response.  Please see https://github.com/swagger-api/swagger-js/issues/1044 for justification.  Do you have a pointer for Node.js v8 supporting `file://` URLs in `fs`?

matthewadams:
@jimmywarting I still don't understand why `node-fetch` shouldn't support `file://` URLs.  You could easily prevent your users from using `file://` URLs simply by testing for it in your API.

jimmywarting:
> You could easily prevent your users from using

[...]

PR #326, opened 2017-08-10
support file URLs

joshua-gould:


jimmywarting:
Thought we decided not to implement file scheme #75

bitinn:
Thx for the PR, Joshua. But we have no plan to add `file` protocol support. Fetch Spec offers no definition for such interaction, and we would rather not argue with them on that.

For filesystem related operations, I would suggest you use a separate module (and there are many on npm).

joshua-gould:
For those that would like to access file URLs with fetch, please see https://gist.github.com/joshua-gould/58e1b114a67127273eef239ec0af8989

PR #328, opened 2017-08-12
Remove the `charset` parameter from the `application/x-www-form-urlencoded` MIME type

### LINKED ISSUE #327: Do not add `charset` parameter to `x-www-form-urlencoded` content-type
The `application/x-www-form-urlencoded` MIME type does not support parameters (such as `charset`). If you look at [this section of the HTML5 spec](http://www.w3.org/TR/html5/forms.html#url-encoded-form-data), you will see how charset is determined (it's complicated). In particular, there is a note at the bottom of the section mentioning how charset cannot be specified as a parameter to the MIME type:

> Note: Parameters on the application/x-www-form-urlencoded MIME type are ignored. In particular, this MIME type does not support the charset parameter.

When working with some web services (like Akismet or Smartling), if I send a POST request using a `URLSearchParams` body, I always get an error from these services.

For Smartling, I get the explicit response:
`Content type 'application/x-www-form-urlencoded;charset=UTF-8' not supported`.

If I set my own `Content-Type` header to `application/x-www-form-urlencoded` (e.g. without a `charset`), everything works fine.

cedx:
Fixes #327

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -367,7 +367,7 @@ export function extractContentType(instance) {
 		return 'text/plain;charset=UTF-8';
 	} else if (isURLSearchParams(body)) {
 	 	// body is a URLSearchParams
-		return 'application/x-www-form-urlencoded;charset=UTF-8';
+		return 'application/x-www-form-urlencoded';
 	} else if (body instanceof Blob) {
 		// body is blob
 		return body.type || null;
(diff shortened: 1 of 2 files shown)


cedx:
Closing this PR as the [Fetch spec](https://fetch.spec.whatwg.org/#body-mixin) requires the presence of `charset` parameter.

PR #331, opened 2017-08-24
Support trailers

### LINKED ISSUE #330: Support trailers
The `Response` class has a [`trailer`](https://fetch.spec.whatwg.org/#dom-response-trailer) attribute that we don't currently implement. Node.js exposes trailers through [`res.trailer`](https://nodejs.org/api/http.html#http_message_trailers) that is only populated when `'end'` event is fired.

TimothyGu:
Fixes #330.

### DIFF
--- a/src/headers.js
+++ b/src/headers.js
@@ -284,6 +284,20 @@ const HeadersIteratorPrototype = Object.setPrototypeOf({
 	Object.getPrototypeOf([][Symbol.iterator]())
 ));
 
+export function nodeToHeaders(headersObj) {
+	const headers = new Headers();
+	for (const name of Object.keys(headersObj)) {
+		if (Array.isArray(headersObj[name])) {
+			for (const val of headersObj[name]) {
+				headers.append(name, val);
+			}
+		} else {
+			headers.append(name, headersObj[name]);
+		}
+	}
+	return headers;
+}
+
 Object.defineProperty(HeadersIteratorPrototype, Symbol.toStringTag, {
 	value: 'HeadersIterator',
 	writable: false,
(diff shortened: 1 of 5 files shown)


bitinn:
while this look fine, I haven't used trailer header before; need another pair of eyes to come in and check this is legit.

TimothyGu:
@jkantr?

jkantr:
LGTM

(sorry for the tardy response, i slipped and rather fell out of our plane of existence for a while)

bitinn:
I am honestly happy to merge.

But it seems like major browsers are resisting this feature:

https://github.com/whatwg/fetch/issues/473 (see linked bug tracker)

TimothyGu:
Switched to standard pending per https://github.com/bitinn/node-fetch/pull/331#issuecomment-347779333.

phamann:
@TimothyGu What is the status of this? 

Whilst yes, Chrome etc have not implemented Trailer support in their fetch implementations. It is still been accepted as part of the spec and useful within server-side environments regardless of browser implementation status. 

Trailers have become more useful since this PR was opened due to their adoption within the [Server-Timing API](https://w3c.github.io/server-timing/) as a mechanism to return timing information about a response after it is closed, such as the final round-trip time etc.

Firefox have also recently implemented trailer support specifically for Server-Timing: https://bugzilla.mozilla.org/show_bug.cgi?id=1413999

Therefore, would you consider reopening and merging this PR? I am happy to do the work to fix the merge conflicts etc if so :)

xxczaki:
Closing this, as https://github.com/bitinn/node-fetch/pull/690 is on hold.

PR #333, opened 2017-08-24
Use built in req.setTimeout

nordsimon:
Use built in set timeout to simplify code and avoid dealing with timeout key manually

https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,25 +48,19 @@ export default function fetch(url, opts) {
 
 		// send request
 		const req = send(options);
-		let reqTimeout;
 
 		if (request.timeout) {
-			req.once('socket', socket => {
-				reqTimeout = setTimeout(() => {
-					req.abort();
-					reject(new FetchError(`network timeout at: ${request.url}`, 'request-timeout'));
-				}, request.timeout);
-			});
+			req.setTimeout(request.timeout, () => {
+				req.abort();
+				reject(new FetchError(`network timeout at: ${request.url}`, 'request-timeout'));
+			})
 		}
 
 		req.on('error', err => {
-			clearTimeout(reqTimeout);
 			reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
 		});
 
 		req.on('response', res => {
-			clearTimeout(reqTimeout);
-
 			// handle redirect
 			if (fetch.isRedirect(res.statusCode) && request.redirect !== 'manual') {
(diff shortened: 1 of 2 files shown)


nordsimon:
While this is not an issue it simplifies code and makes the implementation easier

TimothyGu:
@nordsimon Can you take a look at the Travis CI failures on Node.js 4.x? We do indeed support versions that far back for now...

nordsimon:
Absolutely, i will check it out

nordsimon:
I think the issue is resolved with using flushHeaders instead of write('test') on /slow

However an interesting behaviour should up that we might want to look at. For some reason few tests in node 4 actually resolved with body sent over. I take a look the code now and seems like expected behaviour is to abort even though some payload has been flushed down the pipe.

TimothyGu:
There's one last issue it seems: https://travis-ci.org/bitinn/node-fetch/jobs/268230717#L1368-L1369

> expected behaviour is to abort even though some payload has been flushed down the pipe.

@bitinn Could you take a look?

bitinn:
@nordsimon @TimothyGu 

A few comments while I investigating:

- I want to keep the original test server config, as I explicitly want to test a slow body (where you have some part of body already, but due to slow networking, the rest is still incoming), so when user set "timeout", even a slow body won't hold the client for more than the "timeout" value.

- I believe it's different from [flushHeaders](https://nodejs.org/api/http.html#http_request_flushheaders) where the intended usage is when the whole body is not available.

- ref: https://github.com/bitinn/node-fetch/blob/master/src/body.js#L190-L196

- As for why the change make Node 4 fail, I suspect ClientRequest.setTimeout is doing something we don't expect, I haven't figured out why just yet. https://github.com/nodejs/node/blob/v4.8.4/lib/_http_client.js#L574-L605

bitinn:
I re-run Travis for the previous commit (the one without test server change), it's failing more than Node 4, Node 6 also failed. I think the change created a subtle timing issue.

So my current take is do not merge.

https://travis-ci.org/bitinn/node-fetch/jobs/268113460

nordsimon:
Thanks for your feedback @bitinn 

Honestly my only intention was to simply this piece of logic with using internals but yes, there are some mysterious timing issues. I took a look at the native implementation and it's very similar but has a few differences. I might come back later with a complete solution using internal timers but for now I will leave this where it is

PR #340, opened 2017-09-07
proper stack first line for FetchError instances (#215)

TooTallNate:
The `.stack` property gets cached in the `captureStackTrace()` call, so
whatever is set as the `name` and `message` at that time will be used
for the first line of the stack trace.

Before this patch, FetchError's stack would just say "Error" as the
first line. Now they correctly display the "${name}: ${message}" of the
error instances.

Test case included.

Signed-off-by: Timothy Gu <timothygu99@gmail.com>

### DIFF
--- a/lib/fetch-error.js
+++ b/lib/fetch-error.js
@@ -17,9 +17,6 @@ module.exports = FetchError;
  */
 function FetchError(message, type, systemError) {
 
-	// hide custom error implementation details from end-users
-	Error.captureStackTrace(this, this.constructor);
-
 	this.name = this.constructor.name;
 	this.message = message;
 	this.type = type;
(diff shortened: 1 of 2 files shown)


TooTallNate:
@bitinn Hey, this is https://github.com/bitinn/node-fetch/commit/fa225291280e161dedca5b923c73f82cda23c3fe rebased for the `1.x` branch. See https://github.com/bitinn/node-fetch/pull/215#issuecomment-327371526. Cheers!

bitinn:
@TooTallNate Nathan, could you make the test working in 1.x?

- Pretty sure it's a either a syntax error or because we use older chai
- Also backtick is not available for node 0.10/0.12, so you might want to use an old school solution instead.

(Yeah, both sucks but that's the baggage we carry.)

TooTallNate:
Fixed (it was older Chai).

bitinn:
1.7.3 is out, thx!

TooTallNate:
Thank you @bitinn!

PR #350, opened 2017-10-04
Add an "additionalRequestProperties" option to requests

voxpelli:
This is to allow eg. AWS XRay instrumentation, which requires the addition of an "XRaySegment" value to be sent to the `http.request()` call.

The reasoning for adding `additionalRequestProperties` is to preserve backwards compatibility – just assigning all unknown properties from `init` could result in unwanted values being forwarded to `http.request()`.

I am aware that this module wants to mimik the browser `fetch()` and this property doesn't align with that, but it do align with properties like `agent` and is useful when eg. there's a need to better trace calls across microservices in a server environment.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -79,6 +79,10 @@ export default class Request {
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
 
+		if (init.additionalRequestProperties) {
+			Object.assign(this, init.additionalRequestProperties);
+		}
+
 		this[PARSED_URL] = parsedURL;
 		Object.defineProperty(this, Symbol.toStringTag, {
 			value: 'Request',


TimothyGu:
As I understand, AWS X-Ray's Node.js API requires using a separate `http` module anyway, which node-fetch does not support. I'm going to close this PR for now, but would be happy to discuss and potentially reopen.

voxpelli:
@TimothyGu Xray instruments http modules when in automatic mode, this PR was targeting their manual mode.

I’ve been using this PR in production on a pretty large site since I submitted it and it works well

voxpelli:
At least that’s what I remember, I can take a closer look on Monday

josh--newman:
@voxpelli, Did you manage to get X-Ray to play nicely with node-fetch? I tried using the [`captureHTTPsGlobal`](https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/module-http_p.html#.captureHTTPsGlobal) function with no luck.

voxpelli:
@josh--newman We're running https://github.com/Sydsvenskan/node-fetch/tree/1.x-fork internally to achieve this behavior and will probably move to another module eventually, that better suits our node.js needs. This module is more geared towards modules that should work both in in node.js and the browser (no x-ray there)

josh--newman:
@voxpelli thanks for your response! I'm considering reevaluating our use of `node-fetch` as well.

PR #354, opened 2017-10-16
Request and Response should define read only attributes

bseber:
recently my team wrote test code like this:

```js
// business code
const jsonBody = response.json ? await response.json().catch(() => null) : null;
const responseWithJsonBody = Object.assign ({}, response, { jsonBody: jsonBody });
return responseWithJsonBody.ok ? responseWithJsonBody : Promise.reject (responseWithJsonBody);

// test
expect(errorResponse.status).toBe(400);
expect(errorResponse.ok).toBe(false);
expect(errorResponse.jsonBody).toBeNull();
```

I don't want to discuss this code snippet...
but using `Object.assign` to create a new response object with additional stuff doesn't break the test currently. However, in the real world all `Response`attributes are not enumerable and therefore are not copied with `Object.assign`.

imho this lib should ensure the real world behaviour =)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -84,18 +84,22 @@ export default function fetch(url, opts) {
 					return;
 				}
 
+				let opts = {
+					headers: new Headers(request.headers),
+					follow: request.follow,
+					counter: request.counter + 1,
+					agent: request.agent,
+					compress: request.compress,
+				};
 				// per fetch spec, for POST request with 301/302 response, or any request with 303 response, use GET when following redirect
 				if (res.statusCode === 303
 					|| ((res.statusCode === 301 || res.statusCode === 302) && request.method === 'POST'))
 				{
-					request.method = 'GET';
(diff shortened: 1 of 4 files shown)


TimothyGu:
Per [Web IDL](https://heycam.github.io/webidl/#es-attributes) it should really be

```js
Object.defineProperty(Class.prototype, "property", {
  get() {
    return this[/* a private symbol behind which the property is hidden */].property;
  },
  // no set()
  enumerable: true,
  configurable: true
});
```

bseber:
hey @TimothyGu thank you for the fast feedback 👍

I've hacked a bit to protect the readOnly fields of the Request and Response objects. Let me know about what you think. Thanks!

TimothyGu:
This has been folded into b1cd2dd438a17e4e4412365bc33ab4d542e34930, soon to be released as v2.0.0.

PR #356, opened 2017-10-20
Check for illegal keys in the options argument

vinz243:
I was writing some script and I spent literally 1 to 2 hours pulling my head out as to why the server was responding with 500 when same curl was working just fine. Turns out I used the field `data` instead of `body`. This could have been avoided if I used TypeScript, or if node-fetch checked for illegal options.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -10,7 +10,7 @@ import Body, { clone, extractContentType, getTotalBytes } from './body';
 
 const { format: format_url, parse: parse_url } = require('url');
 
-const ALLOWED_OPTIONS = ['method', 'headers', 'body', 'redirect', 'follow', 'timeout', 'compress', 'size', 'agent'];
+const ALLOWED_OPTIONS = ['method', 'headers', 'body', 'redirect', 'follow', 'timeout', 'compress', 'size', 'agent', 'counter'];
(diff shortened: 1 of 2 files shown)


jimmywarting:
This is kind of unusual to whitelist stuff on a object in the javascript world.
I would certainly not expect an error if i throw in a extra property. 
Eventually we are going to have a abort signal and and ppl are going to send that in even if node-fetch don't support it the browser may support something that the node version don't support and that would make isomorphic fetch don't work well together

```js
req = new Request(url, { signal })
```

I have heard about backward compatible, but I would say that this isn't very feature compatible.
This is something that belongs more to typescript and should only be meant to do IDE autocompletion or suggestion of how to write your code - not throw errors all around

jimmywarting:
at the very least i would agree upon a console warning saying "did you mean body instead of data in your request?" and only if it's not a get/head request and if the object has a `data` property and not a `body` property. And that should only be logged once

I could imagine some ppl do that mistake think i also have done it once long before

TimothyGu:
Sorry, but this change goes against the behavior of browsers and Node.js APIs as well the most popular JavaScript idioms.

jimmywarting:
"idioms" was the word i was looking for, thanks

PR #358, opened 2017-10-30
Document default redirect policy

riking:
Source: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch

### DIFF
--- a/LIMITS.md
+++ b/LIMITS.md
@@ -12,6 +12,8 @@ Known differences
 
 - `res.url` contains the final url when following redirects.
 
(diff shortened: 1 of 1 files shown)


bitinn:
I don't think the spec has changed yet, if it changes we are likely going to change in v2 as well.

https://fetch.spec.whatwg.org/#requests

> A request has an associated redirect mode, which is "follow", "error", or "manual". Unless stated otherwise, it is "follow".

PR #362, opened 2017-11-09
fix a small oversight in request class 

bitinn:
where lowercase method name does not trigger type error


bitinn:
FYI, the travis failure is due to: https://github.com/nodejs/node/issues/16649

the latest stable npm is not compatible with latest stable node.js;

(also reminds me we need to add node 8 lts test in travis too...)

Issue #284, opened 2017-06-02
Debugging


dandv:
Would be great to have a debugging feature to examine requests before they go down the wire, [a la `request`](https://github.com/request/request#debugging).

jimmywarting:
Think this has been brought up before...

lostpebble:
Yep, debugging requests on Node is tough with this module. Would really like a way to inspect / debug things.

sabrehagen:
Yes, this would be a major help and has been requested before #150.

bitinn:
A few suggestions:

1. Be more specific about what you want from *debugging*; we are pretty transparent about exposing errors (unlike Fetch in browser, which doesn't give you much to investigate), but maybe you want more.

2. We are careful not to expose too much control related to native HTTP(S) module, because at that point you might as well be working with native modules. Nevertheless, we like to hear your pain point.

3. We are very understaffed (see #252 and [our contributors graph](https://github.com/bitinn/node-fetch/graphs/contributors)), if you got a great idea, open an PR or issue, it starts the conversation.

4. There are literally PR opened related to debugging (see #363 #364), but we hear nothing from people about if they want it or not.

lostpebble:
Sorry for not being specific enough.

Basically I came here looking for debugging options because my Authentication headers were not working. Turns out I was using the wrong syntax (thanks to me trying to work around incorrect TypeScript definitions https://github.com/Microsoft/TypeScript/issues/19656 - which had appeared to be working at first):

```
headers: [["Content-Type", "application/json"], ["Authorization", `Basic ${Buffer.from(`vibescout:${ServiceConfig.MailChimp.apiToken}`).toString("base64")}`]]
```
Instead of:
```
headers: {
    "Content-Type": "application/json",
    Authorization: `Basic ${Buffer.from(`vibescout:${ServiceConfig.MailChimp.apiToken}`).toString("base64")}`,
}
```

But what would have been helpful to debug it was if the final / created request being sent out, including headers etc. was able to be printed to console - then I could explore it with Google Chrome devtools and see that my headers were not being created properly somehow. Instead I was left wondering if my API token had expired.

So my specification for basic debugging would be a switch to say: print the request object and the response object to the debug console for further exploration.

Understandably you don't seem to have enough time to implement every little thing now. I'll see if I can dig in and help out at some point.

jimmywarting:
Just something that pop in to my mind is: wouldn't it be cool if you could see network request in chrome's devtool?

[...]

PR #364, opened 2017-11-17
add timings

### LINKED ISSUE #363: Add timings for http calls
Would be good to expose timings of http calls for things like dns lookups and so on without introducing a shim to achieve this.

E.g 
https://blog.risingstack.com/measuring-http-timings-node-js/

https://github.com/RisingStack/opentracing-auto/blob/master/src/instrumentation/httpClient.js

ajohnstone:
fixes #363

### DIFF
--- a/src/fetch-error.js
+++ b/src/fetch-error.js
@@ -18,6 +18,7 @@ export default function FetchError(message, type, systemError) {
 
 	this.message = message;
 	this.type = type;
+	this.timings = (systemError || {}).timings;
 
 	// when err.type is `system`, err.code contains system error code
 	if (systemError) {
--- a/src/index.js
+++ b/src/index.js
@@ -76,13 +76,14 @@ export default function fetch(url, opts) {
 			if (request.timeout) {
 				reqTimeout = setTimeout(() => {
 					req.abort();
-					reject(new FetchError(`network timeout at: ${request.url}`, 'request-timeout'));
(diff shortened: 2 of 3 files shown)


bitinn:
I am not against it per se but we try very hard to avoid extending Fetch Spec too much. I would keep it open for future discussion.

ajohnstone:
@bitinn Not sure I follow your reservations, and how to proceed?

bitinn:
@ajohnstone I believe timing data should be completely optional to users. And #284 is discussing some possible options to debug requests.

I believe `request` approach of accepting every features are wrong, so we usually stick with Fetch Spec. And it certainly doesn't expose this information through API.

@TimothyGu might have a different opinion but mine are not merging.

ajohnstone:
Would be happy if req/res was exposed :+1:

TimothyGu:
I agree with @bitinn on this.

lostthetrail:
@TimothyGu @bitinn How about @ajohnstone suggestion of exposing the raw Node req/res objects, so that users can extend the module with the timing information we want/need?

bitinn:
I agree that we need to get a move on at #284, but I am fully occupied with other projects, so unlikely the PR would come from me :(

Of course if anyone want to tackle #229, which discusses a more elegant way to do this, I am all ears.

TimothyGu:
No movement on this; closing.

> How about @ajohnstone suggestion of exposing the raw Node req/res objects, so that users can extend the module with the timing information we want/need?

I'm fine with this.

PR #365, opened 2017-11-19
Test against Node.js v8

MarkHerhold:
Node.js v9 is out so we need to specify testing against v8


bitinn:
@MarkHerhold Thx a lot for the reminder! There are some minor things we need to workaround on the Travis CI integration part so I did it with another PR #366 

I will close this for now.

PR #367, opened 2017-11-19
Adding comment to recommend 2.x branch

bitinn:
I am making some changes to Readme:

- Adding comment to recommend 2.x releases;
- Adding an alternative to `isomorphic-fetch`;

One thing I am considering:

- Now that MDN is basically the official doc for every major browsers, should we just link to it instead of listing all spec compliant interfaces/classes in Readme? We probably only need to document the differences.

- Since we are not going to take 2.x out of alpha due to Fetch Spec itself is constantly evolving; shall we at some point switch to 2.x alpha as our latest (mainline) npm release?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 node-fetch
 ==========
 
-[![npm version][npm-image]][npm-url]
+[![npm stable version][npm-image]][npm-url]
+[![npm next version][npm-next-image]][npm-url]
 [![build status][travis-image]][travis-url]
 [![coverage status][codecov-image]][codecov-url]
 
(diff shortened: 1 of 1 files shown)


bitinn:
@TimothyGu @jkantr @jimmywarting @kirill-konshin @matthew-andrews 

Love to hear your thought on this (sorry for the spam, just trying to ping some major contributors).

matthew-andrews:
If you are recommending using v2.x, is now the right time to publish v2.x versions to npm?

bitinn:
@matthew-andrews 2.x has been there for about a year, it's behind a `next` tag https://github.com/bitinn/node-fetch#install

I believe quite a few projects are using it. Like https://github.com/mikeal/r2

matthew-andrews:
Ohh, I did not know you could do that with NPM.  My bad for misunderstanding NPM.  👍

PR #373, opened 2017-11-28
Add 'rawBody' options for preventing *zip decoding

dimapaloskin:
Hello.

This POC pr brings ability to prevent gzip decoding even the `compress` option was set up as `true`. It's can be useful if you are want to work with raw body (decompress it yourself or just pass through).

What are you think about this, guys?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -131,12 +131,13 @@ export default function fetch(url, opts) {
 			// HTTP-network fetch step 16.1.3: handle content codings
 
 			// in following scenarios we ignore compression support
-			// 1. compression support is disabled
-			// 2. HEAD request
-			// 3. no Content-Encoding header
-			// 4. no content response (204)
-			// 5. content not modified response (304)
-			if (!request.compress || request.method === 'HEAD' || codings === null || res.statusCode === 204 || res.statusCode === 304) {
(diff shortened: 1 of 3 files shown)


bitinn:
@dimapaloskin you sentence is self-contradictory, please provide an example on what can't be done within current release.

either you want compression handling or you don't. use `compress: false`.

casdevs:
@bitinn: I think this could be useful for download progress calculation (received bytes vs. total bytes from content-length header) while at the same time using compressed content encoding.

in the current implementation using e.g. gzip encoding, we can only count the received *decompressed* bytes, which are always more than the content-length header (size of compressed response) states.

Or are there any other ways of doing exact progress calculations in case of compressed responses I'm not aware of?

bitinn:
@casdevs Fetch isn't designed to do this, you are better off calculating the body size. Also, generally speaking, a properly setup server shouldn't compress large binary files when serving them, so you can still use content-length safely in those cases.

https://github.com/nginx-boilerplate/nginx-boilerplate/blob/795a1b5e9d24d7dfcb0f8008fbb5cbd05e921a44/boilerplate/enable/gzip.conf

jimmywarting:
@casdevs fyi, https://github.com/whatwg/fetch/issues/607

PR #379, opened 2017-12-09
Add error event hander for the body stream even if the body isn't accessed

### LINKED ISSUE #378: Errors on the body pipe are only handled if the requests body is accessed which crashes the node process
When requesting a site with `node-fetch` and the response sets the `content-encoding` header to `gzip`, the body gets piped through Gunzip:
```js
// for gzip
if (name == 'gzip' || name == 'x-gzip') {
	body = body.pipe(zlib.createGunzip());
	output = new Response(body, response_options);
	resolve(output);
	return;
}
```
https://github.com/bitinn/node-fetch/blob/1.7.3/index.js#L208

In cases where the encoding is errorsome Gunzip will emit an error event on the pipe:
```js
function zlibOnError(message, errno) {
  // there is no way to cleanly recover.
  // continuing only obscures problems.
  _close(this);
  this._hadError = true;

  var error = new Error(message);
  error.errno = errno;
  error.code = codes[errno];
  this.emit('error', error);
}
```
https://github.com/nodejs/node/blob/v8.9.1/lib/zlib.js#L156

If I already called `response.text()` (`fetch(url).then(response => response.text())`) or similar (e.g. `.json()`), `Body.prototype._decode` gets called and sets an error handler on the body steam:

```js
Body.prototype._decode = function() {
	[...]
	return new Body.Promise(function(resolve, reject) {
		[...]
		self.body.on('error', function(err) {
			reject(new FetchError('invalid response body at: ' + self.url + ' reason: ' + err.message, 'system', err));
		});
		[...]
	});
};
```
https://github.com/bitinn/node-fetch/blob/1.7.3/lib/body.js#L122

However if I don't call `.text()` (for example because I only care for the status code), there will be no error handler and node will throw an exception:

```js
EventEmitter.prototype.emit = function emit(type) {
  [...]
  var doError = (type === 'error');

  events = this._events;
  if (events)
    doError = (doError && events.error == null);
  [...]
  // If there is no 'error' event listener then throw.
  if (doError) {
    if (arguments.length > 1)
      [...]
    } else if (er instanceof Error) {
      throw er; // Unhandled 'error' event
    }
  }
  [...]
};
```
https://github.com/nodejs/node/blob/v8.9.1/lib/events.js#L169

Because this exception is thrown so deep in the core of node, it is not possible to catch the response and nod crashes because of an uncaught exception.

A workaround for this is to always call `response.text()` even if you don't use it.

This is a very nasty error because you get little to no infos on what it the cause. All you get is:
```
events.js:183
      throw er; // Unhandled 'error' event
      ^
Error: unexpected end of file
    at Gunzip.zlibOnError (zlib.js:153:15)
```

I'm using:
Node: `v8.9.1`
Node Fetch: `1.7.3`
OS: maxOs 10.12.6

MoritzKn:
Fixes #378

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -12,6 +12,7 @@ const Stream = require('stream');
 const { PassThrough } = require('stream');
 
 const DISTURBED = Symbol('disturbed');
+const ERROR = Symbol('error');
 
 let convert;
 try { convert = require('encoding').convert; } catch(e) {}
@@ -49,8 +50,15 @@ export default function Body(body, {
 	}
 	this.body = body;
 	this[DISTURBED] = false;
+	this[ERROR] = null;
 	this.size = size;
 	this.timeout = timeout;
+
+	if (this.body instanceof Stream) {
+		this.body.on('error', err => {
+			this[ERROR] = new FetchError(`Invalid response body while trying to fetch ${this.url}: ${err.message}`, 'system', err);
+		});
+	}
 }
 
 Body.prototype = {
@@ -152,6 +160,10 @@ function consumeBody(body) {
 	}
 
 	this[DISTURBED] = true;
+
+	if (this[ERROR]) {
+		return Body.Promise.reject(this[ERROR]);
+	}
 
 	// body is null
 	if (this.body === null) {
--- a/test/test.js
+++ b/test/test.js
@@ -587,6 +587,40 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should handle errors on the body stream even if it is not used', function(done) {
+		url = `${base}invalid-content-encoding`;
+		fetch(url)
+			.then(res => {
+				expect(res.status).to.equal(200);
+			})
+			.catch(() => {})
+			.then(() => {
+				// Wait a few ms to see if a uncaught error occurs
+				setTimeout(() => {
+					done();
+				}, 50);
+			});
+	});
+
+	it('should collect handled errors on the body stream to reject if the body is used later', function() {
+
+		function delay(value) {
+			return new Promise((resolve) => {
+				setTimeout(() => {
+					resolve(value)
+				}, 100);
+			});
+		}
+
+		url = `${base}invalid-content-encoding`;
+		return fetch(url).then(delay).then(res => {
+			expect(res.headers.get('content-type')).to.equal('text/plain');
+			return expect(res.text()).to.eventually.be.rejected
+				.and.be.an.instanceOf(FetchError)
+				.and.have.property('code', 'Z_DATA_ERROR');
+		});
+	});
+
 	it('should allow disabling auto decompression', function() {
 		url = `${base}gzip`;
 		opts = {


bitinn:
I think this make sense but I am not certain how browser actually handle this. Fetch Spec seems to suggest a different approach (see step 10):

https://fetch.spec.whatwg.org/#http-network-fetch

MoritzKn:
I tested this in firefox and chrome and both seem to behave like I implemented it.
I have no good overview of the fetch spec, could you please explain what you think would be the right behavior.

bitinn:
@MoritzKn Sorry I haven't had time to look deep into browser behavior recently.

Do you happen to have any screenshots or output from browsers? Do they actually reject invalid encoding instead of just returning result without decompression or garbled result?

The last time I checked Fetch doesn't expose this sort of information.

MoritzKn:
Sorry I have no screenshots of the output or anything but I could make some this evening. In my tests with invalid encoding the fetch promise was resolve but the `.text()` promise was rejected.

MoritzKn:
@bitinn little later then promised, sorry. 

Running this code in node:
```js
const http = require('http');
const url = require('url');

const server = http.createServer((req, res) => {
  if (url.parse(req.url).pathname === '/invalid') {
    res.setHeader('Content-Encoding', 'gzip');
    res.end('not gzip');
    return;
  }

  res.setHeader('Content-Type', 'text/html; charset=utf-8');
  res.end(`
   <!doctype html>
   <title>Test</title>
   <script>

 fetch('http://localhost:12345/invalid')
   .then(result => console.log('simple request is fine', result))
   .catch(error => console.error('error with simple request', error))

 fetch('http://localhost:12345/invalid').then(res => res.text())
   .then(result => console.log('calling .text() is fine', result))
   .catch(error => console.error('error with calling .text()', error))

 fetch('http://localhost:12345/invalid').then(res => res.blob())
   .then(result => console.log('calling .blob() is fine', result))
   .catch(error => console.error('error with calling .blob()', error))

   </script>
  `);
});

server.listen(12345, 'localhost', () => {
  console.log('listening at http://localhost:12345');
});
```

And opening http://localhost:12345/ in Firefox show this in the console:
image

And this in Chrome:
image

MoritzKn:
The [github polyfill](https://github.com/github/fetch) however rejects the error directly;
image

I applied the polyfill [without the check](https://gist.github.com/MoritzKn/0c95194abacb1b888c9916612c6bb5e1) if `fetch` is already present.

If I have time this weekend, I will try to look into this further and perhaps file a spec issue. The behavior doesn't seem well defined.

MoritzKn:
foo @github/fetch fdg

MoritzKn:
Accidentally closed

bitinn:
Thx for looking into this and producing a concise test case.

My guess is browsers doesn't consider this an important case to normalize because there will be no break-the-world exceptions.

TimothyGu:
Without looking at the current PR, from a spec standpoint this is pretty well defined. Other than parameter validation errors, `fetch()` will never return a promise that’s rejected. Instead we should have some kind of signaling mechanism to body consumption methods for them to exit early in case some errors have already occurred.

MoritzKn:
@TimothyGu indeed that is what the PR does.

I got around to this in Safari and Edge too. 

Running the above snippet and opening the page in Safari results in these logs:
image

It logs that something went wrong (Failed to load resource: decoding the raw data not possible.) in the console but resolves the promise. `.text()` simply resolve to an empty string and `.blob()` and to a zero size blob.

In Edge the behavior is similar to Chrome and Firefox however the `.text()` call returns a `Promise` rejected with a `TypeError` "Body has already been read" which is not the case.
image

Nevertheless I think an uncaught exception that can't be handled and causes the node process to crash isn't desirable in any case.

MoritzKn:
I now also created an issue on the fetch spec: whatwg/fetch#657
For now I think we should go with the behavior implemented in this PR since it's the most in line with browser behavior and makes the most sense.

Issue #399, opened 2018-02-11
bodyUsed bug


jamesplease:
Affected `node-fetch` versions: ^1.0, 2.0

**Issue type:**
bug (possibly)

**Summary:**
Passing a used `body` into the constructor of a `Response` does not set the `bodyUsed` property of the new `Response ` to `true`. The native `Response ` _does_ set the `bodyUsed` property as `true`.

**Reproducing:**

Run the following code:

```js
fetch('https://jsonplaceholder.typicode.com/posts/1')
  .then(res => {
    const text = res.text();
    console.log(res.bodyUsed);
    // => true
    
    var newResponse = new Response(res.body, res);
    console.log(newResponse.bodyUsed);
    // => `true` if using Chrome, `false` if using node-fetch, Safari, Firefox
  });
```

Source of this behavior within `node-fetch`: https://github.com/bitinn/node-fetch/blob/c0950b7b9f0720cc9a0915968d5585d87286fba1/src/body.js#L52

This behavior isn't in the MDN docs, so I'm going to dig into the spec to see if it's described there.

----

Interestingly, both Chrome and node-fetch share the underlying stream between the two responses:

```js
fetch('https://jsonplaceholder.typicode.com/posts/1')
  .then(res => {    
    var newResponse = new Response(res.body, res);
    console.log(newResponse.body === res.body);
    // => `true` for Chrome and node-fetch
  });
```

I'm reasonably certain this is a bug but I'm no expert at the fetch spec, so maybe someone more knowledgeable can confirm :v: Thanks!

jmeas:
I sought out an expert opinion on Twitter (thanks @jakearchibald !). He is also thinking it is a bug:

https://twitter.com/jaffathecake/status/962609913305403392

bitinn:
https://streams.spec.whatwg.org/#is-readable-stream-disturbed

I am not sure if Node.js stream has any equivalent.

An interesting question really: What's left of a stream object, after it has been consumed? I have never thought about it.

bitinn:
By the way `res.body` is just a Node.js stream, nothing more. The `disturbed` flag is not in the `res.body`. The stream must provide a way to tell node-fetch it has been consumed. Otherwise we need to figure it out from `res` (but that's not spec compliant?)

TimothyGu:
@jmeas [Sought further clarifications](https://twitter.com/Timothy_Gu/status/962739344640376832) from @jakearchibald. I am not sure I agree with his reading of the spec.

[...]

PR #402, opened 2018-02-17
Add bodyUsed limitation to Known Differences doc

jamesplease:
For more, see #399 

---

If you'd like me to change the wording here, let me know. I tried to explain it in enough detail while matching the succinctness of the other items on the list, but there may be a better way to explain this.

Thanks for reading!


bitinn:
I might have worded it differently but that's largely correct.

I would probably say: 

> Because node.js stream doesn't expose a [*disturbed*](https://fetch.spec.whatwg.org/#concept-readablestream-disturbed) property like Stream spec, using a consumed stream for `new Response(body)` will not set `bodyUsed` flag correctly.

Issue #247, opened 2017-03-01
Add support for TypedArray, DataView, and ArrayBuffer as body


TimothyGu:
The Fetch Standard [allows](https://fetch.spec.whatwg.org/#bodyinit) `BufferSource` as a `BodyInit`, which is [defined](https://heycam.github.io/webidl/#BufferSource) to include all TypedArray types, DataView, and ArrayBuffer.

gr2m:
I _think_ I run into a problem because of the lack of `ArrayBuffer` support, with a test that runs in the browser but fails in Node.

Can I work on this? Could you give me pointers on where to start?

TimothyGu:
@gr2m Seems like you figured it out :)

For any other people interested in this, one could do the same thing in #408, but for the remaining BufferSource types (TypedArray and DataView).

PR #408, opened 2018-03-01
ArrayBuffer body

gr2m:
adds support for ArrayBuffer body in requests. Part of #247

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,8 +54,12 @@
     "resumer": "0.0.0",
     "rollup": "^0.55.1",
     "rollup-plugin-babel": "^3.0.3",
+    "string-to-arraybuffer": "^1.0.0",
     "url-search-params": "^0.10.0",
     "whatwg-url": "^5.0.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "is-array-buffer": "^1.0.0",
+    "typedarray-to-buffer": "^3.1.2"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -58,7 +58,5 @@
     "url-search-params": "^0.10.0",
     "whatwg-url": "^5.0.0"
   },
-  "dependencies": {
-    "is-array-buffer": "^1.0.0"
-  }
+  "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "whatwg-url": "^5.0.0"
   },
   "dependencies": {
-    "is-array-buffer": "^1.0.0",
-    "typedarray-to-buffer": "^3.1.2"
+    "is-array-buffer": "^1.0.0"
   }
 }
--- a/src/body.js
+++ b/src/body.js
@@ -40,7 +40,7 @@ export default function Body(body, {
 		// body is blob
 	} else if (Buffer.isBuffer(body)) {
 		// body is buffer
-	} else if (body instanceof ArrayBuffer) {
+	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
 		// body is array buffer
 	} else if (body instanceof Stream) {
 		// body is stream
@@ -205,7 +205,7 @@ function consumeBody() {
 	}
 
 	// body is buffer
-	if (this.body instanceof ArrayBuffer) {
+	if (Object.prototype.toString.call(this.body) === '[object ArrayBuffer]') {
 		return Body.Promise.resolve(this.body);
 	}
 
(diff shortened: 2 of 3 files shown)


bitinn:
LGTM

@TimothyGu this might be in your wheelhouse, anything else you were placing for #247?

src/body.js
```diff
@@ -483,6 +490,10 @@ export function writeToStream(dest, instance) {
 		// body is buffer
 		dest.write(body);
 		dest.end()
+	} else if (isArrayBuffer(body)) {
+		// body is array buffer
+		dest.write(typedarrayToBuffer(body));
```
TimothyGu: Why not just `Buffer.from(body)`? Plus, `body` isn't exactly a TypedArray here, as the function call seems to imply.

test/test.js
```diff
@@ -771,6 +772,23 @@ describe('node-fetch', () => {
...
 	});
 
+	it('should allow POST request with ArrayBuffer body', function() {
+		url = `${base}inspect`;
+		opts = {
+			method: 'POST'
+			, body: stringToArrayBuffer('Hello, world!\n')
+		};
+		return fetch(url, opts).then(res => {
+			return res.json();
```
TimothyGu: nit: `res => res.json()`

src/body.js
jimmywarting: also, why not use `ArrayBuffer.isView(body)` instead of `is-array-buffer`?

package.json
```diff
@@ -54,8 +54,12 @@
     "resumer": "0.0.0",
     "rollup": "^0.55.1",
     "rollup-plugin-babel": "^3.0.3",
+    "string-to-arraybuffer": "^1.0.0",
```
jimmywarting: I took a look at [string-to-arraybuffer](https://github.com/dfcreative/string-to-arraybuffer) at what it dose, think it can lead to unintentional behavior since it dose different things depending on what kind of string it is (dataUri, base64 or plain strings)

What if the server only accepts base64 strings? the lib will convert it to none-base64 buffer...?
can't we just use `Buffer.from(str, 'utf-8')`?

TimothyGu:
I'm fine with adding a dependency as lightweight as is-array-buffer, though inlining the check (`Object.prototype.toString.call(value) === '[object ArrayBuffer]'`) isn't too much work either.

src/body.js
TimothyGu: string-to-arraybuffer is only used in the testing server.

> also, why not use `ArrayBuffer.isView(body)` instead of `isArrayBuffer(body)`?

If we do the same thing for ArrayBufferViews as for ArrayBuffer, we would then be copying the view, which we don't want. We need special handling for that (`Buffer.from(body.buffer, body.byteOffset, body.byteLength)`), and I believe that should be in a separate PR.

gr2m: > Why not just `Buffer.from(body)`

No idea that this would just work, but looks like it does, I tested it against [`@octokit/rest`](https://github.com/octokit/rest.js/blob/5a8fca6e69cd5786b58e95933dc789a6dacb2131/test/integration/release-assets-buffer-test.js). Done via f9f328b

> why not use ArrayBuffer.isView(body) instead of isArrayBuffer(body)?

That’s not working, but `body instanceof ArrayBuffer`. Any reason against using that? See 8d3752d

> string-to-arraybuffer 

I’ll leave that as is as per Timothy’s comment

gr2m:
> consumeBody() (return Body.Promise.resolve(Buffer.from(this.body)), which converts the ArrayBuffer to a Buffer)

Can you clarify? That change is not working

```diff
diff --git a/src/body.js b/src/body.js
index 8a3bbf4..9e24db9 100644
--- a/src/body.js
+++ b/src/body.js
@@ -79,7 +79,7 @@ Body.prototype = {
 	 * @return  Promise
 	 */
 	arrayBuffer() {
-		return consumeBody.call(this).then(buf => buf.buffer.slice(buf.byteOffset, buf.byteOffset + buf.byteLength));
+		return Body.Promise.resolve(Buffer.from(this.body));
 	},
 
 	/**
```

> extractContentType() (returning null directly is fine)

👍 cf86d15

gr2m:
> > why not use ArrayBuffer.isView(body) instead of isArrayBuffer(body)?

> That’s not working, but `body instanceof ArrayBuffer`. Any reason against using that? See 8d3752d

Used `Object.prototype.toString.call(body) === '[object ArrayBuffer]'` instead after discussing with Timothy

PR #409, opened 2018-03-02
307 redirect with PATCH & body

gr2m:
When the server returns a 307 redirect, `node-fetch` does not post the request body to the new URL, but `fetch` does, at least in Chrome.

Here is a failing test from https://github.com/octokit/rest.js/pull/764

https://travis-ci.org/octokit/rest.js/jobs/348045840#L4095

The same test passes when run in Chrome

https://travis-ci.org/octokit/rest.js/jobs/348045843#L4040

I tried to create a test case to reproduce the problem, does that look good? I’m not sure how to fix it though, I’d appreciate pointers :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,8 @@
  * index.js
  *
  * a request API compatible with window.fetch
+ *
+ * All spec algorithm step numbers are based on https://fetch.spec.whatwg.org/commit-snapshots/ae716822cb3a61843226cd090eefc6589446c1d2/.
  */
 
 import Body, { writeToStream, getTotalBytes } from './body';
@@ -141,10 +143,10 @@ export default function fetch(url, opts) {
 				, timeout: request.timeout
 			};
 
-			// HTTP-network fetch step 16.1.2
+			// HTTP-network fetch step 12.1.1.3
 			const codings = headers.get('Content-Encoding');
 
-			// HTTP-network fetch step 16.1.3: handle content codings
+			// HTTP-network fetch step 12.1.1.4: handle content codings
 
 			// in following scenarios we ignore compression support
 			// 1. compression support is disabled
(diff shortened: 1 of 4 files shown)


bitinn:
make sense, looks like for all the tests we did, we didn't actually test for 307 redirect with post/patch, which do require you to keep the request method and body.

without looking too closely the fix might be just a 2 line change (add body when creating new request and remember to drop it if it's 301/302)

https://github.com/bitinn/node-fetch/blob/master/src/index.js#L87-L103

bitinn:
(btw, the 301/302 POST redirect to GET isn't very usefully at the moment because deliberate lack of built-in cookie store; might be something we work on in the future)

gr2m:
@bitinn thanks David, I’ll look into it! If you could have a look at #408, too, that'd be great 🙏

gr2m:
All set, I think ☝️👀 :octocat:

bitinn:
LGTM, I will let @TimothyGu take a pass on this (he is the maintainer for v2 branch)

src/index.js
```diff
@@ -102,6 +102,12 @@ export default function fetch(url, opts) {
 					requestOpts.headers.delete('content-length');
 				}
 
+				// for 307 redirects, the body of the original request is reused to perform the redirected request
+				// https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307
+				if (res.statusCode === 307) {
+					requestOpts.body = request.body;
```
TimothyGu: Two comments:

This step should be executed for all status codes other than 303. This spec-mandated behavior is echoed in Chrome.

We should reject the promise with an FetchError here if `request.body` is a readable stream. See https://fetch.spec.whatwg.org/#concept-http-redirect-fetch, step 9. ("*request*’s body’s source is null" is an oblique way of saying "*request* is backed by a stream.) It makes sense too not to consume the same stream multiple times.

PR #413, opened 2018-03-04
Harmonize style

TimothyGu:
Most of the code base already consistently uses the tab + comma-at-the-end style. This converts the rest.


gr2m:
Maybe add linting to CI to make sure it stays that way?

TimothyGu:
@gr2m See https://github.com/bitinn/node-fetch/issues/303. Let's do that in a separate step.

bitinn:
I am happy with this.

PR #415, opened 2018-03-06
Reject when stream accumulation fails (Fix #414)

### LINKED ISSUE #414: Fetch promise never completes due to swallowed RangeError
I'm in a scenario where the following happens:

1. I make a request to a particular API using node-fetch, telling the server to do some processing and stream back results as JSON as they become ready.
2. The API responds successfully and begins streaming back data.
3. There is much more data coming back than I expected, so I tell the API (via another HTTP request) to stop the processing.
4. The server complies (stops processing), and closes the initial stream.
5. Because the stream has closed, [the `end` event is emitted on `body`](https://github.com/bitinn/node-fetch/blob/fb86ab34a6e6f00bfae5fd8cbebda9af4a0f412f/src/body.js#L254).
6. While handling the `end` event, node-fetch tries to create a new `ArrayBuffer` from the data that has streamed in so far, but the amount of data is too large, and a `RangeError` exception is thrown.
7. [node-fetch does _not_ `reject` with the error in the `end` event handler](https://github.com/bitinn/node-fetch/blob/fb86ab34a6e6f00bfae5fd8cbebda9af4a0f412f/src/body.js#L260), so the error is effectively lost, and the `Promise` that I received from node-fetch when I made the initial request is never resolved or rejected. The result is that the initial request continues to wait for the `Promise` to complete in some way, and (as a result) my UI currently shows the request as continuously ongoing.

Issue #203 is _somewhat_ similar to this, though not identical. What I'd like to happen here is for node-fetch to `reject` with the error, [as it does for other streaming errors](https://github.com/bitinn/node-fetch/blob/fb86ab34a6e6f00bfae5fd8cbebda9af4a0f412f/src/body.js#L246), so that the `Promise` completes and my UI can update as expected. Is there any reason for that not to happen? If it sounds reasonable, I can submit a simple PR for this.

jmrog:
I'm not totally sure whether we should reject this as a `FetchError`, so please let me know, if not (and I'll update the PR accordingly). The underlying error here should always be a `RangeError`.

As for testing: unfortunately, there doesn't appear to be any great way to test this (the `RangeError` is thrown when the buffer size is larger than the maximum allowed by v8 itself, which isn't configurable from within the tests). The best I could do here was to temporarily overwrite `Buffer.concat`, which is used by node-fetch internally, to force it to throw an error, and then insure that the error is passed on via a `Promise` rejection.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1436,12 +1436,15 @@ describe('node-fetch', () => {
 		const restoreBufferConcat = () => Buffer.concat = bufferConcat;
 		Buffer.concat = () => { throw new Error('embedded error'); };
 
-		return res.text().then(() => chai.assert(false)).catch(err => {
-			expect(err).to.be.an.instanceOf(FetchError)
-				.and.include({ type: 'system' })
-				.and.have.property('message').that.includes('Could not create Buffer')
-				.and.that.includes('embedded error');
-		}).then(restoreBufferConcat, restoreBufferConcat);
+		const textPromise = res.text();
+		// Ensure that `Buffer.concat` is always restored:
+		textPromise.then(restoreBufferConcat, restoreBufferConcat);
+
+		return expect(textPromise).to.eventually.be.rejected
+			.and.be.an.instanceOf(FetchError)
+			.and.include({ type: 'system' })
+			.and.have.property('message').that.includes('Could not create Buffer')
+			.and.that.includes('embedded error');
 	});
 });
 
--- a/test/test.js
+++ b/test/test.js
@@ -16,6 +16,7 @@ const http = require('http');
 const fs = require('fs');
 const path = require('path');
 const stream = require('stream');
+const buffer = require('buffer');
 const { parse: parseURL, URLSearchParams } = require('url');
 
 let convert;
@@ -1427,6 +1428,21 @@ describe('node-fetch', () => {
 		});
 	});
 
+	// issue #414
+	it('should reject if attempt to accumulate body stream throws', function () {
+		let body = resumer().queue('a=1').end();
+		body = body.pipe(new stream.PassThrough());
+		const res = new Response(body);
+		const bufferConcat = Buffer.concat;
+		Buffer.concat = () => { throw new Error('embedded error'); };
+
+		return res.text().then(() => chai.assert(false)).catch(err => {
+			expect(err).to.be.an.instanceOf(FetchError)
+				.and.include({ type: 'system' })
+				.and.have.property('message').that.includes('Could not create Buffer')
+				.and.that.includes('embedded error');
+		}).then(() => Buffer.concat = bufferConcat);
+	});
 });
 
 describe('Headers', function () {
--- a/test/test.js
+++ b/test/test.js
@@ -16,7 +16,6 @@ const http = require('http');
 const fs = require('fs');
 const path = require('path');
 const stream = require('stream');
-const buffer = require('buffer');
 const { parse: parseURL, URLSearchParams } = require('url');
 
 let convert;
@@ -1434,6 +1433,7 @@ describe('node-fetch', () => {
 		body = body.pipe(new stream.PassThrough());
 		const res = new Response(body);
 		const bufferConcat = Buffer.concat;
+		const restoreBufferConcat = () => Buffer.concat = bufferConcat;
 		Buffer.concat = () => { throw new Error('embedded error'); };
 
 		return res.text().then(() => chai.assert(false)).catch(err => {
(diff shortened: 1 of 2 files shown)


bitinn:
Thx for the PR. I think the catch and re-throw as FetchError is right, as this is not a code/programmer error.

I might need to think about how to restructure the test, if at all possible.

jmrog:
> I might need to think about how to restructure the test, if at all possible.

Sure, I agree that the test isn't ideal. The main difficulty with testing here is that the error occurs inside of the 'end' handler on the body stream, which does nothing relevant for this case other than call `Buffer.concat` on an inaccessible private array (`accum`). It looks like `Buffer.concat` will generally throw only if either `accum` is not an array of buffers (which is not really the case we want to test here, and isn't possible in this case anyway) or the size of `accum` is too large for the engine to create a new `ArrayBuffer` (which is not something we can modify -- _unless_ the test actually does pipe an enormous amount of data into the body stream, which I am reluctant to do in a test -- since the engine's size limit isn't configurable inside the test AFAIK, and `accum` is private). It might be necessary to modify the source itself to make this more testable.

test/test.js
```diff
@@ -1427,6 +1428,21 @@ describe('node-fetch', () => {
...
+		const res = new Response(body);
+		const bufferConcat = Buffer.concat;
+		Buffer.concat = () => { throw new Error('embedded error'); };
+
+		return res.text().then(() => chai.assert(false)).catch(err => {
+			expect(err).to.be.an.instanceOf(FetchError)
+				.and.include({ type: 'system' })
+				.and.have.property('message').that.includes('Could not create Buffer')
+				.and.that.includes('embedded error');
+		}).then(() => Buffer.concat = bufferConcat);
```
TimothyGu: We need to reset `Buffer.concat` whether or not the test succeeded or failed.

test/test.js
```diff
@@ -16,6 +16,7 @@ const http = require('http');
 const fs = require('fs');
 const path = require('path');
 const stream = require('stream');
+const buffer = require('buffer');
```
TimothyGu: Is this used anywhere?

jmrog: Nope. Accidentally left over from an earlier iteration. Removed!

test/test.js
jmrog: Good catch (no pun intended!). I think I changed some of the inner logic before submitting the PR and just forgot to update the final part of the chain. In any case, I've updated the PR now to reset in both cases. Thanks!

jmrog:
Just checking whether there's anything further I can do here to push this along. Thanks!

test/test.js
```diff
@@ -1427,6 +1427,22 @@ describe('node-fetch', () => {
...
+		const bufferConcat = Buffer.concat;
+		const restoreBufferConcat = () => Buffer.concat = bufferConcat;
+		Buffer.concat = () => { throw new Error('embedded error'); };
+
+		return res.text().then(() => chai.assert(false)).catch(err => {
+			expect(err).to.be.an.instanceOf(FetchError)
+				.and.include({ type: 'system' })
+				.and.have.property('message').that.includes('Could not create Buffer')
+				.and.that.includes('embedded error');
+		}).then(restoreBufferConcat, restoreBufferConcat);
```
TimothyGu: Specifying these `.then()`s would mask the errors, since `restoreBufferConcat()` doesn't forward errors. This problem would be solved if we move to a Node.js version supporting [`promise.finally()`](https://github.com/tc39/proposal-promise-finally#why-not-thenf-f).

This should work:

```js
		const res = new Response(body);
		const bufferConcat = Buffer.concat;
		const restoreBufferConcat = () => Buffer.concat = bufferConcat;
		Buffer.concat = () => { throw new Error('embedded error'); };
 
		const textPromise = res.text();
		textPromise.then(restoreBufferConcat, restoreBufferConcat);
		return textPromise.then(() => chai.assert(false)).catch(err => {
			expect(err).to.be.an.instanceOf(FetchError)
				.and.include({ type: 'system' })
				.and.have.property('message').that.includes('Could not create Buffer')
				.and.that.includes('embedded error');
		});
```

TimothyGu:
@jmrog Sorry about the delay. It's finals season for me, but just finished my last one this morning so would be able to get this in soon after https://github.com/bitinn/node-fetch/pull/415#discussion_r176602522 gets fixed :)

test/test.js
jmrog: D'oh, silly oversight. Fixed. Thanks! (I did it slightly differently using `chai-as-promised`, but to essentially the same effect.)

jmrog:
@TimothyGu No problem, I understand. Congrats on finishing your last final!

TimothyGu:
Thanks!

Issue #412, opened 2018-03-04
.textConverted needs to install the package encoding separately


motss:
Why the package `encoding` is not part of `node-fetch`'s dependencies to use `.textConverted`?

TimothyGu:
It's intentional: we don't force people who don't use `textConverted` to install `encoding`. We should document this however.

motss:
Ya. A line or two stating this can be helpful! 👍

TimothyGu:
@motss Would you be interested in creating a pull request? ☺️

motss:
@TimothyGu Sure thing. Will work on that probably on the weekend.

peterbakonyi05:
Please note that since `encoding` is not part of the dependencies by default, when using this package with `webpack`, without installing `encoding`, webpack will fail with the usual error message: 
```
WARNING in ./node_modules/cross-fetch/node_modules/node-fetch/lib/index.es.js
Module not found: Error: Can't resolve 'encoding' in 'C:\git\halo.site\node_modules\cross-fetc
h\node_modules\node-fetch\
```

I think it would be nice to also add this to the `README`:
When using `node-fetch` with `webpack`. You either need to install `encoding` or add [ContextReplacementPlugin](https://webpack.js.org/plugins/context-replacement-plugin/)
```js
new webpack.ContextReplacementPlugin(
    /encoding$/,
    path.join(__dirname, "src"),  // location of your src
    {}
),
```

itonics-tbeauvais:
Somehow, using the `nodeExternals()` for the `externals` option in webpack found here https://www.npmjs.com/package/webpack-node-externals

motss:
I thought this had been fixed by one of the PRs. I can update the doc real quick with the findings from @peterbakonyi05 as well.

motss:
I've added briefly in the doc. Please help review #511. Thank you.

bitinn:
I have added guideline to readme and link to this issue.

Issue #375, opened 2017-11-29
Download file example


davideicardi:
I'm having some problems downloading file. In your readme file you have written the following example:

```js
fetch('https://assets-cdn.github.com/images/modules/logos_page/Octocat.png')
.then(res => {
  const dest = fs.createWriteStream('./octocat.png');
  res.body.pipe(dest);
});
```
But it my case it seems that destination file remain locked, it is not closed automatically I think. Probably is not related to `node-fetch` library but it is just how stream works. But maybe you can add a more complete example ...

I have written this code:

```js
const res = await fetch('https://assets-cdn.github.com/images/modules/logos_page/Octocat.png');
await new Promise<void>((resolve, reject) => {
  const fileStream = fs.createWriteStream('./octocat.png');
  res.body.pipe(fileStream);
  res.body.on("error", (err) => {
    fileStream.close();
    reject(err);
  });
  fileStream.on("finish", function() {
    fileStream.close();
    resolve();
  });
});
```

I'm right or I missing something?

thanks

bitinn:
Turns out you just need an `autoClose` option :)

https://nodejs.org/api/fs.html#fs_fs_createwritestream_path_options

PR welcomed to fix this code sample.

jkantr:
er.. was just looking at this for the readme update discussion... `createWriteStream`'s `autoClose` option defaults to `true`. It should be closing the fd itself on `error` or `finish` events. The example as is *should* work.

davideicardi:
The problem I think is that when the promise returns the stream is not actually closed, the pipe is not completed.
So probably a better example (for dummies like me 😄 ) would be:

```
async function download() {
	const res = await fetch('https://assets-cdn.github.com/images/modules/logos_page/Octocat.png');
	await new Promise((resolve, reject) => {
		const fileStream = fs.createWriteStream('./octocat.png');
		res.body.pipe(fileStream);
		res.body.on("error", (err) => {
			reject(err);
		});
		fileStream.on("finish", function() {
			resolve();
		});
	});
}
```

So the example IS correct but if you copy that code you will probably don't get what do you expect.
I'm right?

jdleesmiller:
In node 10+, I think this can be simplified using [stream.pipeline](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback):

```js
const util = require('util')
const fs = require('fs')
const streamPipeline = util.promisify(require('stream').pipeline)

const fetch = require('node-fetch')

async function download () {
[...]

Issue #368, opened 2017-11-21
Post body data example doesn't work, Content-Length is not specified


IjzerenHein:
Hi, great library, just found this issue with an example in the README. The `post with stream from file` example states that this should work:

```js
// post with stream from file
import { createReadStream } from 'fs';

const stream = createReadStream('input.txt');
fetch('http://httpbin.org/post', { method: 'POST', body: stream })
	.then(res => res.json())
	.then(json => console.log(json));
```

However, it doesn't. Httpbin returns with an internal error.

After inspection, I found that you'll need to specify `"Content-Length"` in the header. The following does work though:

```js
const fs = require('fs');
const fetch = require('node-fetch');

fs.readFile('test/assets/test.txt', (error, file) => {
	fetch('http://httpbin.org/post', {
		method: 'POST',
		headers: {
			"Content-Length": file.length
		},
		body: file
	})
	.then(res => res.json())
	.then(json => console.log(json));
});
```

Cheers  🤘

bitinn:
Thx for pointing this out, this is an issue for the receiving side though: without content-length you are using chunked transfer-encoding, and httpbin doesn't handle it.

I guess there was an error when moving from 1.x doc to 2.x; no one actually check whether it works.

I think this original code should work; or at least when I was testing it 2 years ago ;)

```
var resumer = require('resumer');
var stream = resumer().queue('a=1').end();
fetch('http://httpbin.org/post', { method: 'POST', body: stream })
	.then(function(res) {
		return res.json();
	}).then(function(json) {
		console.log(json);
	});
```

jimmywarting:
when creating a readableStream from a file it will include a path ([readstream.path](https://nodejs.org/api/fs.html#fs_readstream_path)) and a `readstream.fd`

With the help of path or `fd` it would be possible for node-fetch to retrieve the content size of the file synchronous with `fs.statSync(path)` or `fs.fstatSync(fd)`

bitinn:
readme was updated to address this issue.

PR #419, opened 2018-03-09
The readme update discussion PR

bitinn:
I am trying to start a PR that fixes many small issues with current readme.

Specifically, these issues: #412, #399, #375, #368

Love to hear some comments (or provide cleaner code example for above issues!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@ node-fetch
 ==========
 
 [![npm stable version][npm-image]][npm-url]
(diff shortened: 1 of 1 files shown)


bitinn:
Some changes I am thinking about:

- I think we should recommend `const fetch = require('node-fetch');` over `import fetch from 'node-fetch';`

- adding table of content if possible.

bitinn:
superseded by #448

PR #426, opened 2018-03-19
wrapped ArrayBuffer with buffer internally to fix body methods

wmsmacdonald:
Initializing `Request` and `Response` with array buffers causes issues for `.text()`:
```javascript
const { Response } = require('node-fetch');

const response = new Response(new Uint8Array[97]);
response.text().then(text => {
    // text === '[object ArrayBuffer]', should be 'a'
});
```

The problem is due to the fact that `ArrayBuffer.toString()` returns `[object ArrayBuffer]` instead of decoding the `ArrayBuffer`. Converting the `ArrayBuffer` to a `Buffer` fixes `.text()` and `.json()`.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -206,7 +206,7 @@ function consumeBody() {
 
 	// body is buffer
 	if (Object.prototype.toString.call(this.body) === '[object ArrayBuffer]') {
-		return Body.Promise.resolve(this.body);
+		return Body.Promise.resolve(new Buffer(this.body));
 	}
 
 	// istanbul ignore if: should never happen
--- a/src/body.js
+++ b/src/body.js
@@ -206,7 +206,7 @@ function consumeBody() {
 
 	// body is buffer
 	if (Object.prototype.toString.call(this.body) === '[object ArrayBuffer]') {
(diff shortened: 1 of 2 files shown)


src/body.js
```diff
@@ -206,7 +206,7 @@ function consumeBody() {
 
 	// body is buffer
 	if (Object.prototype.toString.call(this.body) === '[object ArrayBuffer]') {
-		return Body.Promise.resolve(this.body);
+		return Body.Promise.resolve(new Buffer(this.body));
```
gr2m: note that `new Buffer` is deprecated, use `Buffer.from` instead, see https://nodejs.org/api/buffer.html#buffer_new_buffer_buffer

This creates a copy of `this.body` as far as I know I’m not sure if that’s what we want?

wmsmacdonald: Thanks for taking a look. 

The docs say `Buffer.from()` "creates a view of the ArrayBuffer without copying the underlying memory." I think we can then decode (in `.text()` for example) without any additional memory overhead.

TimothyGu:
Thanks! I’ll try to get a release out this week.

PR #429, opened 2018-03-22
Fix redirection: response object must be consumed otherwise process gets stuck

rzaharenkov:
Fixes https://github.com/bitinn/node-fetch/issues/428

> If no 'response' handler is added, then the response will be entirely discarded. However, if a 'response' event handler is added, then the data from the response object must be consumed, either by calling response.read() whenever there is a 'readable' event, or by adding a 'data' handler, or by calling the .resume() method. Until the data is consumed, the 'end' event will not fire. Also, until the data is read it will consume memory that can eventually lead to a 'process out of memory' error.

https://nodejs.org/api/http.html#http_event_response

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -128,6 +128,7 @@ export default function fetch(url, opts) {
 
 						// HTTP-redirect fetch step 15
 						resolve(fetch(new Request(locationURL, requestOpts)));
+						res.on('data', () => {}) // response *must* be consumed otherwise it gets stuck forever
 						return;
 				}
 			}


pjg:
Thanks for fixing this!

src/index.js
```diff
@@ -128,6 +128,7 @@ export default function fetch(url, opts) {
 
 						// HTTP-redirect fetch step 15
 						resolve(fetch(new Request(locationURL, requestOpts)));
+						res.on('data', () => {}) // response *must* be consumed otherwise it gets stuck forever
```
TimothyGu: Would `res.resume();` be a cleaner solution? Just wondering based on the Node.js docs.

TimothyGu:
I've proposed my alternative in https://github.com/bitinn/node-fetch/pull/432. Instead of the slightly hacking empty callback, I am now explicitly aborting the request. Both results in the same effect (as `req.abort()` dumps the remaining data in the response as well), however. I also fixed up some surrounding code.

rzaharenkov:
@TimothyGu , I think your solution is cleaner 👍 I am going to close my PR

Issue #425, opened 2018-03-13
Setting Host headers causes: TypeError: hostHeader.startsWith is not a function


woky:
Since version 2.1.0, setting `Host` header causes `fetch` call to fail. This is a minimal example

    const fetch = require('node-fetch');
    (async () => {
        const resp = await fetch('https://google.com', { headers: { 'Host': 'example.com' } });
        console.log(resp);
    })();

that throws following exception:

    (node:5895) UnhandledPromiseRejectionWarning: TypeError: hostHeader.startsWith is not a function
        at calculateServerName (_http_agent.js:240:20)
        at Agent.addRequest (_http_agent.js:155:26)
        at new ClientRequest (_http_client.js:257:16)
        at request (https.js:245:10)
        at /home/user/tmp/node-fetch-host-header/node_modules/node-fetch/lib/index.js:1355:15
        at new Promise (<anonymous>)
        at fetch (/home/user/tmp/node-fetch-host-header/node_modules/node-fetch/lib/index.js:1342:9)
        at /home/user/tmp/node-fetch-host-header/index.js:4:24
        at Object.<anonymous> (/home/user/tmp/node-fetch-host-header/index.js:6:3)
        at Module._compile (module.js:660:30)
    (node:5895) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
    (node:5895) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

In my production code I've also encountered `TypeError: hostHeader.replace is not a function` in almost identical code to the example above.

gr2m:
Here is a test case that reproduces the problem: https://runkit.com/gr2m/5aa826566fefc70013b04566

bitinn:
Weird, I thought this test should cover it?

https://github.com/bitinn/node-fetch/blob/master/test/test.js#L197-L209

woky:
@bitinn Lowercase header key, i.e. `host`, works. I don't know what the `fetch` spec[1] constraints the header keys, but it just felt natural to use what's on the wire in the end.

[1] https://fetch.spec.whatwg.org/

bitinn:
We needed a special handling on host header because nodejs http doesn’t allow it in headers (it’s set separately which we handled automatically)

[...]

Issue #416, opened 2018-03-06
Requests with a `Host` header cause an exception in Node v6


nathanielw:
Hiya, I haven't had a chance to dig into this much, so let me know if any more details are needed. Essentially using node-fetch to send a request with a `Host` header causes an exception in Node v6.13.1:

```javascript
const fetch = require('node-fetch');

fetch('https://api.github.com/users/github', {
		headers: { 'Host': 'api.github.com' }
	});
```

gives:
```
(node:96772) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: hostHeader.startsWith is not a function
```

It looks like this was introduced by the changes in 1592ca1 (the above code works fine in the parent commit).

gr2m:
@TimothyGu that was your change, any idea? Here is a runkit snippet showing that the regression was introduced in 2.1.0: https://runkit.com/gr2m/5a9f1c073815fc0012281973

TimothyGu:
I'll look at this. I do have to say though, I would partially consider it a bug in Node.js that no better diagnostic is emitted.

Palisand:
This is happening for me as well (Node v9.2).

PR #430, opened 2018-03-23
Fix custom Host header with arbitrary case

TimothyGu:
Regression since 1592ca1148e657110fd6f64dc8b29a576cd1da1c.

Fixes: #416
Fixes: #425

### DIFF
--- a/src/headers.js
+++ b/src/headers.js
@@ -324,6 +324,25 @@ Object.defineProperty(HeadersIteratorPrototype, Symbol.toStringTag, {
 });
 
 /**
+ * Export the Headers object in a form that Node.js can consume.
+ *
+ * @param   Headers  headers
+ * @return  Object
+ */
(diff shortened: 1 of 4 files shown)


bitinn:
Interesting so the header keeps its original value and request get the right host string.

But you might add a line in test to verify the original `hoSt` header name is not changed? That's what you are trying to preserve, right?

TimothyGu:
> But you might add a line in test to verify the original `hoSt` header name is not changed? That's what you are trying to preserve, right?

Yes, but we can't actually test for that since Node.js force-lowercases everything on the server side.

bitinn:
But we could check the *request* headers, no? It's not affected by node.js http module, right?

TimothyGu:
@bitinn But we can't get access to the request object given the current API surface, unless you meant check `exportNodeCompatibleHeaders` directly?

Issue #428, opened 2018-03-22
fetch resolves the promise but node process gets stuck


rzaharenkov:
I am faced with really weird issue. I do a fetch and it resolves the promise so I am getting page HTML. But nevertheless node.js process gets stuck. I guess it's something around node-fetch internals. It depends on the URL (I am running into this issue with 1 particular URL).

It's very easy to reproduce this issue.

```
// test.js

const fetch = require('node-fetch')
fetch('http://kjsubz.com').then(response => response.text()).then(console.log)
```

Then run it with `node test.js`

I've tried both v2.0 and v2.1 against different node.js versions (6.13.0, 8.9.1, 9.7.1)

rzaharenkov:
OK, I think I know what is the reason of this issue. But not sure how to fix it properly.

In case of redirect it doesn't consume the response body (https://github.com/bitinn/node-fetch/blob/master/src/index.js#L130). And according to node.js documentation it must do it:

> If no 'response' handler is added, then the response will be entirely discarded. However, if a 'response' event handler is added, then the data from the response object must be consumed, either by calling response.read() whenever there is a 'readable' event, or by adding a 'data' handler, or by calling the .resume() method. Until the data is consumed, the 'end' event will not fire. Also, until the data is read it will consume memory that can eventually lead to a 'process out of memory' error.

I've tried to patch the code and it helped:

```
    // HTTP-redirect fetch step 15
    resolve(fetch(new Request(locationURL, requestOpts)));
+   res.on('data', () => {}) // do nothing but release handler
    return;
```

bitinn:
Interesting, we will look into it.

(From my phone)

TimothyGu:
@bitinn Two pull requests have been opened that fix this :) (#429 and #432).

bitinn:
Yeah it’s not easy catching up via email.

Thx OP for going through the trouble of identifying the problem and making that PR though!

(From my phone)

PR #432, opened 2018-03-23
Make sure to finalize the request properly

TimothyGu:
Unfortunately, I could not write a test case that allows testing the bug
in #428.

Fixes: #428

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -52,18 +52,23 @@ export default function fetch(url, opts) {
 		const req = send(options);
 		let reqTimeout;
 
+		function finalize() {
+			req.abort();
+			clearTimeout(reqTimeout);
+		}
+
 		if (request.timeout) {
 			req.once('socket', socket => {
 				reqTimeout = setTimeout(() => {
-					req.abort();
 					reject(new FetchError(`network timeout at: ${request.url}`, 'request-timeout'));
+					finalize();
 				}, request.timeout);
 			});
 		}
 
 		req.on('error', err => {
-			clearTimeout(reqTimeout);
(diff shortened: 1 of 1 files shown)


bitinn:
I used to use a trick where I set the timeout to be quite short and make sure the request is done before that, otherwise the test fail, does that approach work for testing this PR?

bitinn:
I think I like your approach better as request is explicitly abort, but do make sure to credit @rzaharenkov for discovering this bug, it must have been there since one day of node-fetch?

TimothyGu:
@bitinn The issue is more about the fact that I am so far unable to replicate the behavior of http://kjsubz.com in test/server.js. I'll keep digging.

pjg:
Another URL we've had problems with was this redirection:
* http://tradeoptionswithme.com/ => https://tradeoptionswithme.com/

TimothyGu:
I didn't end up finding a way to test this. Timeout don't work, because the socket timeout won't be triggered since it is not TTFB that's the issue; when the redirect is followed, the body timeout will refer to the redirected page.

The only way to test it would be executing a separate file and observing its behavior (specifically, its runtime). It is a fairly fragile approach however, and I feel it's too much complexity for this one change.

I'll get this merged, and then try to fix the code coverage issue (which came before this PR).

pjg:
Hopefully you'll be able to cut a new release too for. Thanks!

TimothyGu:
@pjg Published as v2.1.2.

rzaharenkov:
I confirm that it solved our issue.

PR #332, opened 2017-08-24
Support TypeScript import

TimothyGu:
Fixes #307.

Issue #307, opened 2017-07-07
Can not import default member in 2.0.0-alpha.5


jkeylu:
```
import fetch from 'node-fetch';
console.log(typeof fetch); // 'undefined'
```

bitinn:
If `const fetch = require('node-fetch');` works, investigate your transpiler setup first.

jkantr:
Honestly, as the loader syntax is still very much up in the air (and all npm modules should still be distributing CommonJS style exports into the foreseeable future) I highly suggest just using `require` and avoiding the headache. You also get the bonus of being able to destructure on the same line you require.

TimothyGu:
Just checked: node-fetch@2.0.0-alpha.5/lib/index.es.js has

```js
export { Headers, Request, Response, FetchError };export default fetch;
```

so `import fetch from 'node-fetch'` would definitely work with Rollup and [rollup-plugin-node-resolve](https://www.npmjs.com/package/rollup-plugin-node-resolve) with `module` option turned on.

lib/index.js has

```js
module.exports = exports = fetch;
exports.Headers = Headers;
exports.Request = Request;
exports.Response = Response;
exports.FetchError = FetchError;
```

so Babel would work as well.

jkeylu:
@bitinn @TimothyGu 
I'm using Typescript to compile my code.
```typescript
import fetch from 'node-fetch';
```

will compile to

```javascript
var fetch = require('node-fetch').default;
```

but in `lib/index.js` does not export default

TimothyGu:
Argh, another module system. Reopening...

TimothyGu:
For now, you should be able to use

```ts
import fetch = require('node-fetch');
```

as documented at https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

bitinn:
Does Typescript definition file handle this type of things? If so might need to contact its author to support node-fetch 2.0

(From my phone)

TimothyGu:
> Does Typescript definition file handle this type of things?

I don't think so. Just specifying `fetch.default = fetch` should be enough to fix this issue, though I'd prefer handling it in the compiler plugins ([Babel](https://github.com/bitinn/node-fetch/blob/master/build/babel-plugin.js), [Rollup](https://github.com/bitinn/node-fetch/blob/master/build/rollup-plugin.js)).

PR #433, opened 2018-03-24
Support TypeScript import in browser

MarkHerhold:
Related to #307 and #332.

TypeScript is compiling
```
fetch(url, options)
```
into
```
const node_fetch_1 = require("node-fetch");
node_fetch_1.default(url, options)
```

This works great in Node.js thanks to #332 but doesn't work in the browser (I'm using Webpack, so it correctly uses the `browser.js` script specified in `package.json`).

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,4 +1,8 @@
 module.exports = exports = window.fetch;
+
+// Needed for TypeScript.
+exports.default = window.fetch;
+
 exports.Headers = window.Headers;
 exports.Request = window.Request;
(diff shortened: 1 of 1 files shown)


MarkHerhold:
@TimothyGu My apologies. It turns out that this is only a partial fix. This works well in Safari but doesn't work in Chrome because fetch isn't bound to `window` in all cases.

The folks developing `unfetch` encountered the same issue here: https://github.com/developit/unfetch/pull/47

I'll PR a fix for Chrome.

PR #434, opened 2018-03-24
Bind fetch to window in the browser

MarkHerhold:
Fixes `fetch` illegal invocation errors in Chrome. This is the same as https://github.com/developit/unfetch/pull/47

As mentioned [here](https://github.com/bitinn/node-fetch/pull/433#issuecomment-375917867), there's an issue in some browsers and Webpack with fetch not being bound to `window` in Webpack.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,7 +1,7 @@
 module.exports = exports = window.fetch;
 
-// Needed for TypeScript.
-exports.default = window.fetch;
+// Needed for TypeScript and Webpack.
+exports.default = window.fetch.bind(window);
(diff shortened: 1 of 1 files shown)


jimmywarting:
Still don't get why ppl want to make node-fetch available in browsers that already has the fetch api...

MarkHerhold:
It's really nice to write code once for both node.js and browser and node-fetch makes the platform switch transparent with the browser.js file. Without this, my code has to do platform detection which is simple but I am happy to abstract away.

PR #437, opened 2018-03-27
[WIP] Add abort support

TimothyGu:
This is still a WIP:

1. Tests are needed.
2. Use a better `AbortSignal` type check after https://github.com/mysticatea/abort-controller/pull/5 is applied.

Note: this PR will introduce a runtime dependency on https://github.com/mysticatea/abort-controller.

Fixes: #95

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,7 @@
     "url-search-params": "^0.10.0",
     "whatwg-url": "^5.0.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "abort-controller": "^1.0.1"
+  }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,7 @@
 import Body, { writeToStream, getTotalBytes } from './body';
 import Response from './response';
 import Headers, { createHeadersLenient } from './headers';
-import Request, { getNodeRequestOptions } from './request';
+import Request, { getAbortSignal, getNodeRequestOptions } from './request';
 import FetchError from './fetch-error';
 
 const http = require('http');
@@ -39,6 +39,12 @@ export default function fetch(url, opts) {
 	return new fetch.Promise((resolve, reject) => {
 		// build request object
 		const request = new Request(url, opts);
+		const signal = getAbortSignal(request);
+		if (signal.aborted) {
+			reject(new FetchError(`Fetch to ${request.url} has been aborted`, 'aborted'));
+			return;
+		}
+
 		const options = getNodeRequestOptions(request);
 
 		const send = (options.protocol === 'https:' ? https : http).request;
@@ -46,10 +52,22 @@ export default function fetch(url, opts) {
 		// send request
 		const req = send(options);
 		let reqTimeout;
+		let body;
+
+		function abortCallback() {
+			const error = new FetchError(`Fetch to ${request.url} has been aborted`, 'aborted');
+			reject(error);
+			if (body !== undefined) {
+				body.emit('error', error);
+			}
+			finalize();
+		}
+		signal.addEventListener('abort', abortCallback);
 
 		function finalize() {
 			req.abort();
 			clearTimeout(reqTimeout);
+			signal.removeEventListener('abort', abortCallback);
 		}
 
 		if (request.timeout) {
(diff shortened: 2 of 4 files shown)


src/request.js
```diff
@@ -82,11 +100,23 @@ export default class Request {
 			}
 		}
 
+		const abortController = new AbortController();
+		if (signal !== undefined) {
+			if (signal.aborted) {
+				abortController.abort();
+			} else {
+				signal.addEventListener('abort', () => {
```
mika-fischer: This eventListener is never removed, is it?

jimmywarting: If it could be of any help maybe adding {once: true} to addEventListener could help cleaning up things

TimothyGu: @mika-fischer No, and per spec we can't, so we'll just have to rely on GC to do its job. We could add a `{ once: true }` though.

mika-fischer: But won't this cause issues, for instance in the case where I have a long running Node.js application where I have a global AbortController (to be able to abort all requests when the application shuts down) and each request in the whole lifetime of the app (let's say hundreds of thousands) adds a listener, none of which is ever removed?

TimothyGu: 😔 yes. It’s also one of the reasons why this is a WIP.

mika-fischer: Can't you just save a cleanup closure (which just removes the listener from the parent AbortController) and attach this closure to the appropriate events of the Node.js Request/Response so that it gets called when the request is finished?

RikkiGibson:
Hi there. I would love to see this happen. Is there any way I can assist?

TimothyGu:
Hi @RikkiGibson, yes! The biggest thing right now is making sure the "abort" signal handlers get cleaned up correctly, i.e., https://github.com/bitinn/node-fetch/pull/437#discussion_r177984831. But I'd appreciate any help in adding tests as well.

You could submit a pull request against the `abort` feature branch.

src/request.js
RikkiGibson: A cleanup closure would remove the listener from the parent AbortController after fetch() terminates for any reason, right? Would that affect the ability to cancel a fetch that reuses a Request object for multiple fetches? Does that maybe mean that the listener should be added each time the fetch begins and removed when the fetch terminates?

RikkiGibson: `.once()` seems like it would only help if the request terminated by being aborted, but I might be missing something.

package.json
```diff
@@ -58,5 +58,7 @@
     "url-search-params": "^0.10.0",
     "whatwg-url": "^5.0.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "abort-controller": "^1.0.1"
```
heri16: Update to abort-controller@^1.0.2

src/request.js
```diff
@@ -82,11 +100,23 @@ export default class Request {
 			}
 		}
 
+		const abortController = new AbortController();
```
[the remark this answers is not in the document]
TimothyGu: Per spec, the `AbortSignal` we expose through `signal` must be a sanitized object, not be the `signal` passed in. To create a new `AbortSignal` we have to create another `AbortController`.

heri16:
I haven't been able to find the spec that require signal to be exposed. Where can I find this information?

RikkiGibson:
https://fetch.spec.whatwg.org/#request-class

Look for the Request constructor. Something like step 30 it will mention that the passed in signal must be used to create a “following” signal, i.e. one that subscribes to a parent signal (given by the user) but can also be canceled independent of the parent signal.

src/index.js
```diff
@@ -39,17 +39,35 @@ export default function fetch(url, opts) {
 	return new fetch.Promise((resolve, reject) => {
 		// build request object
 		const request = new Request(url, opts);
+		const signal = getAbortSignal(request);
+		if (signal.aborted) {
+			reject(new FetchError(`Fetch to ${request.url} has been aborted`, 'aborted'));
```
joaovieira: any way this could be more inline with the browser spec? rejecting with an `AbortError`? when using with isomorphic fetch this would result in having to catch two types of errors for node and browser

https://dom.spec.whatwg.org/#aborting-ongoing-activities

PR #445, opened 2018-04-14
Pass family

mjesun:
In order to be able to select the IP type when DNS querying, `family` needs to be proxied from the options passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,7 @@ The following node-fetch extension properties are provided:
 - `compress`
 - `counter`
 - `agent`
+- `family`
 
 See [options](#fetch-options) for exact meaning of these extensions.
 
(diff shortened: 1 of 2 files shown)


mjesun:
@bitinn friendly ping :)

mjesun:
We use the module at FB, which uses IPv6-only networks. However, we still return sometimes `A` registries from our DNS.

mjesun:
Yup!

TimothyGu:
Can you update https://github.com/mjesun/node-fetch/blob/e943b2273837fe4398ffb9d3951665db3380e996/README.md#options as well?

bkw:
I think you'll have to save the property to `this.family` [in the constructor](https://github.com/mjesun/node-fetch/blob/e943b2273837fe4398ffb9d3951665db3380e996/src/request.js#L100) as well for this to work.
Also to preserve this options through redirects, you'll probably want to add it to [the options for  following redirects](https://github.com/mjesun/node-fetch/blob/e943b2273837fe4398ffb9d3951665db3380e996/src/index.js#L116), too.

Take a look at #473 too see what I mean.

TimothyGu:
Superseded by #474.

PR #448, opened 2018-05-01
README 2.x update proposal... take 1

jkantr:
Just getting some feelers out there to see if this sort of thing would be received favorably. I reworked some of the structure and content of the docs, tried to make them a little more beginner friendly, take a few less liberties, be a little more consistent, and urge some best practices.

Let me know if you all think it's going in the right direction...

OH, right, also I have it identical here in a gist, if this is easier than trying to see the markdown otherwise lol:

https://gist.github.com/jkantr/1b7bcc0e660548476fbc2f184467c41e

(note, this doesn't directly refer to the points in #419 as tbh i wasn't super clear what those updates were meant for... but I could work them in I'm sure)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,52 @@ node-fetch
 
 A light-weight module that brings `window.fetch` to Node.js
 
+<!-- TOC -->
+
+- [Motivation](#motivation)
+- [Features](#features)
+- [Difference from client-side fetch](#difference-from-client-side-fetch)
+- [Installation](#installation)
+- [Loading and configuring the module](#loading-and-configuring-the-module)
+- [Common Usage](#common-usage)
+    - [Plain text or HTML](#plain-text-or-html)
+    - [JSON](#json)
+    - [Simple Post](#simple-post)
+    - [Post with JSON](#post-with-json)
+    - [Post with form parameters](#post-with-form-parameters)
+    - [Handling exceptions](#handling-exceptions)
+    - [Handling client and server errors](#handling-client-and-server-errors)
+- [Advanced Usage](#advanced-usage)
+    - [Streams](#streams)
+    - [Buffer](#buffer)
+    - [Accessing Headers and other Meta data](#accessing-headers-and-other-meta-data)
+    - [Post data using a file stream](#post-data-using-a-file-stream)
+    - [Post with form-data (detect multipart)](#post-with-form-data-detect-multipart)
+- [API](#api)
+    - [fetch(url[, options])](#fetchurl-options)
+    - [Options](#options)
+        - [Default Headers](#default-headers)
+    - [Class: Request](#class-request)
+    - [new Request(input[, options])](#new-requestinput-options)
+    - [Class: Response](#class-response)
+    - [new Response([body[, options]])](#new-responsebody-options)
+    - [response.ok](#responseok)
+    - [Class: Headers](#class-headers)
+    - [new Headers([init])](#new-headersinit)
+    - [Interface: Body](#interface-body)
+    - [body.body](#bodybody)
+    - [body.bodyUsed](#bodybodyused)
+    - [body.arrayBuffer()](#bodyarraybuffer)
+    - [body.blob()](#bodyblob)
+    - [body.json()](#bodyjson)
+    - [body.text()](#bodytext)
+    - [body.buffer()](#bodybuffer)
+    - [body.textConverted()](#bodytextconverted)
+    - [Class: FetchError](#class-fetcherror)
+- [License](#license)
+- [Acknowledgement](#acknowledgement)
+
+<!-- /TOC -->
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


bitinn:
Great job Jared! Rewriting manual is a tiresome job but they are critical to users, thank you for taking the time to do this.

Formatting and readability wise I think this is much better than the current one. @TimothyGu @jimmywarting @gr2m if you got time, give it a quick read and let us know if there are technical details you want to amend?

A few suggestions:

- You seem to be editing an older version of the readme? Some recent changes appear to be lost here (which might also explain why the PR is conflicting?)

> See Matt Andrews' isomorphic-fetch or Leonardo Quixada's cross-fetch for isomorphic usage (exports node-fetch for server-side, whatwg-fetch for client-side).

- The Table of Content could be shorter? Personally I would take out 3rd level items to make it shorter.

- `textConverted` API should mention an optional dependency on `encoding`, which users need to install by themselves, and this is done purely for backward compatibility with 1.x release. 

- we always said `content-length` will be "automatically calculated, if possible" in the default header section, but we never explain what's the condition for it to be calculated, and that chunked transfer-encoding will be used when they are not calculated or supplied.

As for my original PR #419 , I was trying to flag some issues I would like to work on (but never did).

- I have summarized issues related to readme above.
- #402 is for known limits doc, not directly related to readme doc.
- I am closing #419 in favorite of this PR.

jkantr:
> You seem to be editing an older version of the readme?

Oh dear, don't I feel sheepish.. usually I'm pretty good with making sure I'm at head... I'll get that merged in properly asap o.O

Good points all around, I'll dig into each when I find a few moments :) and will await input from others as well~

Thanks for the response!

jimmywarting:
> If you prefer to cache binary data in full, use buffer(). (NOTE: buffer() is a node-fetch only API)

Somewhere i think we also should mention `arrayBuffer()` also if you want to be cross-fetch compatible.  

> Use native stream for body, on both request and response.

I feel conflicted of what is a native stream now that both node and browser has 2 different spec. 
Seeing it as the fetch api is more of a web thing then it is a NodeJS thing i could maybe believe it would giving me a whatwg stream since that is what is used in the browser.
Think they should be called Node stream or whatwg stream to distinguish those to

README.md
```diff
@@ -1,411 +1,492 @@
...
+#### JSON
+```javascript
+fetch('https://api.github.com/users/github')
+    .then(res => res.json())
+    .then(json => console.log(json));
+```
+
+#### Simple Post
+```javascript
+fetch('http://httpbin.org/post', { method: 'post', body: 'a=1' })
```
TimothyGu: I would encourage uppercasing the method names.

README.md
```diff
@@ -1,411 +1,492 @@
...
+
+Current stable release (`2.x`)
+
+```sh
+$ npm install node-fetch --save
+```
+
+## Loading and configuring the module
+We suggest you load the module via `require`, pending the stabalizing of es modules in node:
+```javascript
```
TimothyGu: Can we be consistent about the language tag? I'd prefer just `js`.

README.md
```diff
@@ -1,411 +1,492 @@
...
+
+Instead of implementing `XMLHttpRequest` in Node.js to run browser-specific [Fetch polyfill](https://github.com/github/fetch), why not go from native `http` to `fetch` API directly? Hence `node-fetch`, minimal code for a `window.fetch` compatible API on Node.js runtime.
+
+See Matt Andrews' [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch) or Leonardo Quixada's [cross-fetch](https://github.com/lquixada/cross-fetch) for isomorphic usage (exports `node-fetch` for server-side, `whatwg-fetch` for client-side).
+
+
+## Features
+
+- Stay consistent with `window.fetch` API.
+- Make conscious trade-off when following [whatwg fetch spec][whatwg-fetch] and [stream spec](https://streams.spec.whatwg.org/) implementation details, document known difference.
```
TimothyGu: nit: WHATWG

README.md
```diff
@@ -1,411 +1,492 @@
...
+```
+
+## Common Usage
+
+NOTE: The documentation below is up-to-date with `2.x` releases, [see `1.x` readme](https://github.com/bitinn/node-fetch/blob/1.x/README.md), [changelog](https://github.com/bitinn/node-fetch/blob/1.x/CHANGELOG.md) and [2.x upgrade guide][UPGRADE-GUIDE.md] for the differences.
+
+#### Plain text or HTML
+```javascript
+fetch('https://github.com/')
+    .then(res => res.text())
```
TimothyGu: Style-wise, the code base uses tabs for indenting. I understand why you changed it to spaces, but I feel two spaces is more common in the JavaScript world anyway.

README.md
```diff
@@ -1,411 +1,492 @@
...
+
+If no values are set, the following request headers will be sent automatically:
+
+Header            | Value
+----------------- | --------------------------------------------------------
+`Accept-Encoding` | `gzip,deflate` _(when `options.compress === true`)_
+`Accept`          | `*/*`
+`Connection`      | `close` _(when no `options.agent` is present)_
+`Content-Length`  | _(automatically calculated, if possible)_
+`User-Agent`      | `node-fetch/1.0 (+https://github.com/bitinn/node-fetch)`
```
TimothyGu: Maybe also add `Transfer-Encoding: chunked`? That header is added by Node.js automatically if the input is a stream, I believe.

README.md
```diff
@@ -1,411 +1,492 @@
...
+
+- `body` A string or [Readable stream][node-readable]
+- `options` A [`ResponseInit`][response-init] options dictionary
+
+Constructs a new `Response` object. The constructor is identical to that in the [browser](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response).
+
+Because Node.js does not implement service workers (for which this class was designed), one rarely has to construct a `Response` directly.
+
+#### response.ok
+
```
TimothyGu: This one is also spec-compliant.

README.md
```diff
@@ -1,411 +1,492 @@
...
-[UPGRADE-GUIDE.md]: https://github.com/bitinn/node-fetch/blob/master/UPGRADE-GUIDE.md
-
-[whatwg-fetch]: https://fetch.spec.whatwg.org/
-[response-init]: https://fetch.spec.whatwg.org/#responseinit
-[node-readable]: https://nodejs.org/api/stream.html#stream_readable_streams
+node-fetch
+==========
+
+[![npm stable version][npm-image]][npm-url]	
+[![npm next version][npm-next-image]][npm-url]
```
TimothyGu: I don't think we need a `next` version any more, not until we prepare for v3.x.

TimothyGu:
Might be good to also mention the fact that we don't do any proxy business, à la #449.

README.md
jimmywarting: Minor thing that bugs me is the use of http and [protocol relative urls](https://jeremywagner.me/blog/stop-using-the-protocol-relative-url/). Should use https whenever possible to stop spying mitm attacks. Saying this cuz i know that httpbin has https

bitinn:
It might be a good idea to reformat the **options** section into a table layout (like the **headers** section), instead of current `code` block.

Mainly because `agent` is an easily overlooked feature that we support since v1, and you need it for:

- proxy
- self-signed certificate
- keep alive
- possibly other low level network stuffs

Using a table might give it a bit more awareness and give us more space to write (code block isn't great for line wrapping)

README.md
```diff
@@ -1,411 +1,492 @@
...
+## Difference from client-side fetch
+
+- See [Known Differences][LIMITS.md] for details.
+- If you happen to use a missing feature that `window.fetch` offers, feel free to open an issue.
+- Pull requests are welcomed too!
+
+
+## Installation
+
+Current stable release (`2.x`)
```
gr2m: maybe remove that? The npm badge on top shows the version, we could also move it here

README.md
```diff
@@ -1,411 +1,492 @@
...
+[codecov-image]: https://img.shields.io/codecov/c/github/bitinn/node-fetch.svg?style=flat-square
+[codecov-url]: https://codecov.io/gh/bitinn/node-fetch
+
+[ERROR-HANDLING.md]: https://github.com/bitinn/node-fetch/blob/master/ERROR-HANDLING.md
+[LIMITS.md]: https://github.com/bitinn/node-fetch/blob/master/LIMITS.md
+[UPGRADE-GUIDE.md]: https://github.com/bitinn/node-fetch/blob/master/UPGRADE-GUIDE.md
+
+[whatwg-fetch]: https://fetch.spec.whatwg.org/
+[response-init]: https://fetch.spec.whatwg.org/#responseinit
+[node-readable]: https://nodejs.org/api/stream.html#stream_readable_streams
```
gr2m: I would move `License` below `Acknowledgement`

bitinn:
@jkantr I know the suggestions are a bit overwhelming, if it's too much to fix, let's work on the factual updates first, and leave other suggestion to a later PR :)

jkantr:
Sorry I've been invisible for awhile, trying to surface.

Unfortunately in switching between devices and making updates in a less-than-ideal environment, I think I once again have merged the changes in a very non-optimal way. At this point should I just blow away this branch and submit a new PR with just the changes in their current state?

PR #452, opened 2018-05-08
Passing through original request timeout on new requests generated fr…

tsein-bc:
The timeout option wasn't working as I would expect it to when redirects are followed. If a custom timeout of, say 100ms is specified, and the second link in the redirect chain takes 500ms to respond, the custom timeout is NOT honored.

This change passes through the original request timeout to each subsequent redirect request so that a request-timeout error would be thrown instead of waiting longer than desired.

According to the documentation:
"req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)" it's unclear whether "resets" means unlimited timeouts on followed redirects, or if it's supposed to reset to the original request custom timeout. I think the latter makes more sense.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -113,7 +113,8 @@ export default function fetch(url, opts) {
 							agent: request.agent,
 							compress: request.compress,
 							method: request.method,
-							body: request.body
+							body: request.body,
+							timeout: request.timeout
 						};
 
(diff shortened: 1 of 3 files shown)


bitinn:
Sorry for the delay but this is now fixed and released as 2.4.0

PR #457, opened 2018-05-22
Added support for `ArrayBufferView`

### LINKED ISSUE #247: Add support for TypedArray, DataView, and ArrayBuffer as body

nazar-pc:
Fixes #247

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -39,9 +39,11 @@ export default function Body(body, {
 	} else if (body instanceof Blob) {
 		// body is blob
 	} else if (Buffer.isBuffer(body)) {
-		// body is buffer
-	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
-		// body is array buffer
+		// body is ArrayBuffer
+	} else if (body instanceof ArrayBuffer) {
+		// body is ArrayBufferView
+	} else if (ArrayBuffer.isView(body)) {
+		// body is array buffer view
 	} else if (body instanceof Stream) {
 		// body is stream
 	} else {
@@ -204,8 +206,8 @@ function consumeBody() {
 		return Body.Promise.resolve(this.body);
 	}
 
-	// body is buffer
-	if (Object.prototype.toString.call(this.body) === '[object ArrayBuffer]') {
+	// body is ArrayBuffer
+	if (this.body instanceof ArrayBuffer) {
 		return Body.Promise.resolve(Buffer.from(this.body));
 	}
 
@@ -416,8 +418,11 @@ export function extractContentType(instance) {
 	} else if (Buffer.isBuffer(body)) {
 		// body is buffer
 		return null;
-	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
-		// body is array buffer
+	} else if (body instanceof ArrayBuffer) {
+		// body is ArrayBuffer
+		return null;
+	} else if (ArrayBuffer.isView(body)) {
+		// body is ArrayBufferView
 		return null;
 	} else if (typeof body.getBoundary === 'function') {
 		// detect form data input from form-data module
@@ -457,8 +462,11 @@ export function getTotalBytes(instance) {
 	} else if (Buffer.isBuffer(body)) {
 		// body is buffer
 		return body.length;
-	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
-		// body is array buffer
+	} else if (body instanceof ArrayBuffer) {
+		// body is ArrayBuffer
+		return body.byteLength;
+	} else if (ArrayBuffer.isView(body)) {
+		// body is ArrayBufferView
 		return body.byteLength;
 	} else if (body && typeof body.getLengthSync === 'function') {
 		// detect form data input from form-data module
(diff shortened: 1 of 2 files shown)


nazar-pc:
There is some suspicious undescriptive error in Node.js 4, but I guess it is a bug in Node.js itself.
Since v4 is EOL, maybe just drop its support in next release?

cc @TimothyGu and @gr2m from relevant issue

bitinn:
Thx for PR.

But we can't drop support for Node v4 in non-major release, even though it's end-of-life for Node v4. Reason: semver doesn't know this, people will happily update and find it breaks their legacy environment.

From the error I think it's likely the test/server.js doesn't response, you might need to run it in a Node v4 environment locally to see what server.js is actually reporting. If it's indeed a Node v4 bug, we will find a way to workaround it or skip this test in CI.

nazar-pc:
Turns out Node.js v4 doesn't support `byteOffset` and `length` arguments in `Buffer.from()` API: https://nodejs.org/docs/latest-v4.x/api/buffer.html#buffer_class_method_buffer_from_arraybuffer

Workaround for this can be written, but why? Is there a reason that prevents you from releasing a major version that drops unsupported version of Node.js?

gr2m:
I’m in favor of releasing a new major release and officially drop support for Node 4. It should be easy for everyone to upgrade unless they still depend on Node 4 themselves, which they really shouldn't by now.

0xC0DEBA5E:
Oh, I was working on the same issue today without noticing. Maybe I solved it differently, because I didn't need to use `Buffer.from` and used the `.buffer` property of the ArrayBufferView instance.

bitinn:
@gr2m @TimothyGu @jimmywarting 

I will let you decide if v3 release is required here, personally I think a major release is likely going to split users, so you end up supporting both.

It might not be a huge issue if only node v4 users are affected, but I have some thoughts on this (apologize for going a bit off-topic below):

- I really want to get #448 in v3.
- We should clearly communicate to users, if you don't use nodejs v4, your should upgrade from node-fetch v2 to v3, and it will be a smooth upgrade.
- We need a key maintainer for v3 release. (Timothy has done a great job with v2, I am just not sure he has time to work on v3.)

gr2m:
For an independent project such as node-fetch, I’d recommend to focus on supporting the latest versions. Usually support for outdated environments is a requirement by companies. And if companies depend on it because they cannot upgrade, then they should be the ones maintaining these versions, not the people who put in their free time to create node-fetch in the first place.

This is a good read about that perspective: https://medium.com/@mikeal/stop-supporting-old-releases-70cfa0e04b0c

bitinn:
@gr2m @nazar-pc Alright, I think we are going to merge this PR either way.

How about this approach:

- Just skip the last test when testing in Node v4 CI environment (or if possible, feature detect ` Buffer.from` API, if it doesn't accept the 2 parameter, skip the test)
- Since this is simply an API issue, I don't think anyone will complain, and we can release a non-major update without more wait.
- We will drop Node v4 support in V3 release, but because #448 and a few other things are still pending, I prefer not to leave the build broken (hence my suggestion to skip the last test in node v4).

If it sounds good, @nazar-pc can you update the PR?

nazar-pc:
I'm fine with that, added corresponding commit and CI is green now.

src/body.js
```diff
@@ -39,9 +39,11 @@ export default function Body(body, {
 	} else if (body instanceof Blob) {
 		// body is blob
 	} else if (Buffer.isBuffer(body)) {
-		// body is buffer
-	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
-		// body is array buffer
+		// body is ArrayBuffer
+	} else if (body instanceof ArrayBuffer) {
```
TimothyGu: Was there a reason to change the ArrayBuffer check semantics? This makes it impossible to use e.g. ArrayBuffer from a VM context, while ArrayBufferView from a VM context would work fine.

src/body.js
```diff
@@ -204,8 +206,8 @@ function consumeBody() {
 		return Body.Promise.resolve(this.body);
 	}
 
-	// body is buffer
-	if (Object.prototype.toString.call(this.body) === '[object ArrayBuffer]') {
+	// body is ArrayBuffer
+	if (this.body instanceof ArrayBuffer) {
 		return Body.Promise.resolve(Buffer.from(this.body));
 	}
```
TimothyGu: ArrayBufferViews don't seem to be handled here.

src/body.js
nazar-pc: I'm not sure what you mean by VM context here. `instanceof` looks cleaner and is consistent with the rest of checks like `body instanceof Blob` or `body instanceof Stream`, this was the primary reason for the change.

src/body.js
nazar-pc: It was here initially, but code coverage showed that `ArrayBufferView` does't appear here. When `ArrayBufferView` is sent, only `ArrayBuffer` seems to appear on receiving side. Or am I missing something?

TimothyGu: It is there for cases like

```js
const req = new Request('about:blank', {
  method: 'POST',
  body: new Uint16Array([32123])
});
const j = await req.json();
```

src/body.js
TimothyGu:
```js
const vm = require('vm');
const ForeignArrayBuffer = vm.runInNewContext('ArrayBuffer');
const req = new Request('about:blank', {
  method: 'POST',
  body: new ForeignArrayBuffer(1)
});
```

`ArrayBuffer.isView` works on objects from VM contexts, so it would seem odd for `ArrayBuffer` objects to not support that.

In browsers, this behavior is observable through `iframe` contexts.

src/body.js
nazar-pc: Interesting, but I do not see any explicit use of `Request` in existing tests.

src/body.js
nazar-pc: Thanks for explanation, I think a test case for this tricky scenario should be added.

src/body.js
nazar-pc: I'm just not familiar enough with code base yet, sorry about that. Makes sense to add even more test cases here. As usual, seems so simple on the surface, but quite complex under the hood.

PR #460, opened 2018-05-24
Add support for ArrayBufferView as Request Body

### LINKED ISSUE #459: Wrong Content-Length when uploading binary data
When I'm uploading binary data from a Uint8Array using the POST-Method, more data is uploaded than is contained within the buffer. Instead of 2700000 Bytes there are 9199999 Bytes transfered. Maybe I'm doing something wrong, so here's an example:

```javascript

const sendPost = async (data, url) => {
    try {
        console.log(`byteLength of buffered data:`, data.byteLength);
        const response = await fetch(url,
            {
                method: 'POST', body: data,
                headers: {
                    'Content-Type': 'application/x-protobuf'
                }
            });
        console.log(`response`, response);
        return response;
    } catch (e) {
        console.error(e)
    }
};

const data = new Uint8Array(5);

for (let i = 0; i < 5; i++) {
    data[i] = i;
}

console.log(`byteLenght:`, data.byteLength);
sendPost(data, "http:/httpbin.org/post");
```

At the Server side I can see that the `Content-Length`-Header is set to 9 and node-fetch is trying to send this number of bytes to the server.

0xC0DEBA5E:
ArrayBuffers like Uint8Array were treated like strings and converted to strings, leading to wrong body payloads in i.e. POST requests.
I also added some tests for Uint8Array, Uint16Array and Float64Array.
This fixes #459 and maybe also #457, for which the build was failing.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -39,7 +39,10 @@ export default function Body(body, {
 	} else if (body instanceof Blob) {
 		// body is blob
 	} else if (Buffer.isBuffer(body)) {
-		// body is buffer
+        // body is buffer
+    } else if (body.buffer instanceof ArrayBuffer) {
+		// body is ArrayBuffer or inherits from ArrayBufferView
+        body = body.buffer
 	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
 		// body is array buffer
 	} else if (body instanceof Stream) {
--- a/test/test.js
+++ b/test/test.js
@@ -2051,35 +2051,55 @@ describe('Request', function () {
 
     it('should support Uint8Array as body', function() {
     	const data = new Uint8Array(3);
-        data[0] = 'a';
-        data[1] = '=';
-        data[2] = '1';
+        data[0] = 1;
+        data[1] = 2;
+        data[2] = 3;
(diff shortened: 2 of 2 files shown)


0xC0DEBA5E:
Closed, because it doesn't pass the following test:

```javascript
it('should allow POST request with ArrayBufferView (Uint8Array, offset, length) body', function() {
        		const url = `${base}inspect`;
        		const opts = {
            			method: 'POST',
            			body: new Uint8Array(stringToArrayBuffer('Hello, world!\n'), 7, 6)
        		};
        		return fetch(url, opts).then(res => res.json()).then(res => {
            			expect(res.method).to.equal('POST');
            			expect(res.body).to.equal('world!');
            			expect(res.headers['transfer-encoding']).to.be.undefined;
            			expect(res.headers['content-type']).to.be.undefined;
            			expect(res.headers['content-length']).to.equal('6');
            		});
        	});
});

```

Unfortunately, the `.buffer`-property I was using to convert the ArrayBufferView to a NodeJS Buffer  doesn't allow for an offset to be specified. So maybe the solution from #457 is better than this.

PR #462, opened 2018-05-26
Fix HTTPS for Node 8 & 9

### LINKED ISSUE #461: Proper support of HTTPS for Node 8.6.0 ~ 10.0.0
There was a breaking change back in 8.6.0 which for some reason took a long time to fix and it will land in Node 10, not sooner. Unfortunately, we are stuck with current LTS which is in 8.x line.

There is a detailed discussion here: https://github.com/nodejs/node/issues/16196#issuecomment-336647658 and the relevant PR, that is going to land in 10.x

---
# Quick workaround (otherwise continue reading)

I made my own fork of `node-fetch` which adds `ecdhCurve:auto` which is enough to have a valid cross-platform solution and it's done same way in NodeJS 10+. 

https://www.npmjs.com/package/node-fetch-auto-curve

If you are using `node-fetch` directly, just go ahead and use the fork directly. Otherwise, if you have a 3rd party dependency that is using `node-fetch`, you can win the fight with Yarn great feature called [selective dependency resolution](https://yarnpkg.com/en/docs/selective-version-resolutions). All you need to do is add this to your `package.json` and you run `yarn install`.

```
  "resolutions": {
    "node-fetch": "fredyc/node-fetch#auto-curve"
  },
```

Sadly, I cannot give you any useful advice if NPM is your choice.

FredyC:
Fixes #461 

I've tested this and it works just fine.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -98,6 +98,7 @@ export default class Request {
 			input.compress : true;
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
+		this.ecdhCurve = 'auto';
 	}
 
 	get method() {


FredyC:
Well, I don't understand it either, whole SSL thing is too complex for me. I simply tried to read outcome from those issues and that linked comment particularly is about using this property to fix the issue. I tried it and it indeed worked, so here I am :)

nazar-pc:
From what I understand `ecdhCurve` is an option that controls which particular elliptic curve is used in SSL. There are multiple options for those, but older Node.js version seems to have a single option hardcoded by default, the option which many people consider unsafe (and there is a good reason for that, but I would not go into details why here). From what I understand `auto` just allows to use whatever curve server and client both support/prefer, rather than single hardcoded option.

TL;DR: It will work in all previous cases + many other scenarios that would fail without this PR.

bitinn:
I think this is fine to merge.

Just 2 comments:

- To me adding it between these 2 line seem to be a better choice than adding it to request object, but not a big deal. https://github.com/bitinn/node-fetch/blob/master/src/index.js#L44-L47

- I am asking in the nodejs thread on whether this can be done with https.Agent, many TLS related operation can be done that way, not sure why `ecdhCurve` is left out and have to be passed via request options.

bitinn:
@FredyC so I have double-checked, I think `https.Agent` can pass the `ecdhCurve` option.

- You can create an instance of https.Agent with `ecdhCurve: auto` and pass it to node-fetch https://github.com/bitinn/node-fetch#options

- The relevant test: https://github.com/nodejs/node/blob/master/test/parallel/test-https-agent-additional-options.js

Given this option is available, I don't think we need to merge this fix (a fix for node.js breaking change, really)

FredyC:
Well, I don't agree, simply because this is a highly undocumented problem that will influence many users and it's not really easy to find a culprit. I encountered the error in upstream package and had to follow the path `graphql-cli` -> `graphql-request` -> `cross-fetch` -> `node-fetch`. It took me about 3 hours to figure out what to do.

I don't understand your reluctance given that this is not a breaking change, it's simply to ensure it won't affect users like me. I guess I will have to create my fork if is there some barrier for you :/

> In this PR, you are setting the ecdhCurve property on the Request object. We don't directly pass the Request object to https.request, only getNodeRequestOptions(req), so this PR in fact doesn't have any visible effect on the results.

I guess you are right, it just did not feel right to hide this somewhere so deep.

FredyC:
@bitinn Just realized, is it actually possible to create the `https.Agent` in cross platform manner? In theory if I would propose this in `graphql-request` package, however it aims to be cross-platform solution. My understanding is that not even `cross-fetch` package would expose cross platform variant.

bitinn:
@FredyC I replied in your original issue, let's take the discussion there for now.

FredyC:
Yea, but still I am convinced it would be much better to have it here so every other package depending on this doesn't need to be updated because of a small mishap in NodeJS.

bitinn:
closing this as my comment on #461.

PR #463, opened 2018-05-28
Refine README example

TimothyGu:
This reverts commit fa6548ed316a3c512147a3bb4f4d6e986d738be6 (#441).

The autoClose option has been true by default since at least Node.js
v6.0.0. There is no need to set it once more.

Instead, make the example more realistic by handling stream outcomes
using a promise.

See #375.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,10 +80,19 @@ fetch('http://domain.invalid/')
 
 fetch('https://assets-cdn.github.com/images/modules/logos_page/Octocat.png')
 	.then(res => {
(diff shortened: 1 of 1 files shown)


bitinn:
```
fetch('https://assets-cdn.github.com/images/modules/logos_page/Octocat.png')
.then(res => {
  const dest = fs.createWriteStream('./octocat.png');
  res.body.pipe(dest);
});
```

By the way, was the original code snippet wrong then? Like, it can run into situation where pipe isn't closed properly even though autoClose is default to true?

PR #464, opened 2018-05-28
Fix up ArrayBufferView support

TimothyGu:
Also add more test coverage.

Fixes: #459

/cc @nazar-pc

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -39,11 +39,11 @@ export default function Body(body, {
 	} else if (body instanceof Blob) {
 		// body is blob
 	} else if (Buffer.isBuffer(body)) {
+		// body is Buffer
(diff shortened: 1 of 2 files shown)


nazar-pc:
Checks similar to this are needed to skip some tests on Node.js v4:
https://github.com/bitinn/node-fetch/blob/5bc23d81cf7ebb97b0a7310b2c88c13cdaa706dc/test/test.js#L908-L909

bitinn:
Sorry I drop the ball there not seeing the line change for `body instanceof ArrayBuffer`.

There are usually some reasons for a weird line, I think we just need to document that line more thoroughly.

TimothyGu:
@nazar-pc Right, thanks for noticing. @bitinn No worries… these things are just really subtle (and most of the time not even observable). I should have been more active in reviewing PRs in the first place.

PR #466, opened 2018-05-30
fixed #465

### LINKED ISSUE #465: Can not set Accept-Encoding to gzip only
In my case , I need to set Accept-Encoding to gzip only, without deflate, it seems not work, when I add 'Accept-Encoding: gzip' to headers with method POST.
p.s. compress the body manually before fetching

eongoo:


gr2m:
Could you add a test reproducing the issue from #465 to make sure we won’t break it in future again?

bitinn:
2.2.1 contains this workaround, closing

PR #473, opened 2018-06-14
support option `lookup` for custom dns resolution

bkw:
This enables the use of modules like [lookup-dns-cache](https://www.npmjs.com/package/lookup-dns-cache) to cache dns requests and work around the [thread pool issues caused by dns.lookup](https://nodejs.org/api/dns.html#dns_implementation_considerations).

Adding this functionality here should eliminte the need for complex decorators like [@zeit/fetch-cached-dns](https://www.npmjs.com/package/@zeit/fetch-cached-dns).


bkw:
If it would speed up inclusion, I'd volunteer to pull in (and complete) #256  and #445 along with this, since they are touching the same code areas, do related things and would conflict with this PR anyways.

Comments?

bkw:
Closing in favor of #474

PR #474, opened 2018-06-15
support socket options (son of #256, #445 and #473)

bkw:
This MR unifies #256, #445 and #473.

All options have been added to the documentation and test suite. Support for these options during redirects has been implemented and added to the tests where possible.

Please let me know if You'd like anything changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,8 @@ The default values are shown after each option key.
 	size: 0,            // maximum response body size in bytes. 0 to disable
 	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
 	lookup: dns.lookup  // custom dns lookup function for socket
+	family: null        // numbers 4 or 6 to limit dns resolution to ip v4 or v6
+	localAddress: null  // local interface to bind to for network connections
 }
 ```
 
@@ -257,6 +259,8 @@ The following node-fetch extension properties are provided:
 - `counter`
 - `agent`
 - `family`
+- `lookup`
+- `localAddress`
 
 See [options](#fetch-options) for exact meaning of these extensions.
 
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ The default values are shown after each option key.
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
 	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
+	lookup: dns.lookup  // custom dns lookup function for socket
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -256,6 +256,7 @@ The following node-fetch extension properties are provided:
 - `compress`
 - `counter`
 - `agent`
+- `family`
 
 See [options](#fetch-options) for exact meaning of these extensions.
 
--- a/README.md
+++ b/README.md
@@ -216,9 +216,9 @@ The default values are shown after each option key.
 	timeout: 0,         // req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
-	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
-	lookup: dns.lookup  // custom dns lookup function for socket
-	family: null        // numbers 4 or 6 to limit dns resolution to ip v4 or v6
+	agent: null,        // http(s).Agent instance, allows custom proxy, certificate etc.
+	lookup: dns.lookup, // custom dns lookup function for socket
+	family: null,       // numbers 4 or 6 to limit dns resolution to ip v4 or v6
 	localAddress: null  // local interface to bind to for network connections
 }
 ```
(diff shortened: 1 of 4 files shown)


bkw:
I have no idea what's wrong with the [failing travis build](https://travis-ci.org/bitinn/node-fetch/jobs/392680258)

src/request.js
```diff
@@ -98,6 +98,9 @@ export default class Request {
 			input.compress : true;
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
+		this.lookup = init.lookup || input.lookup;
+		this.family = init.family || input.family;
+		this.localAddress = init.localAddress || input.localAddress;
```
TimothyGu: These should probably have a default value as well, like

```js
this.lookup = init.lookup || input.lookup || dns.lookup;
this.family = init.family || input.family || null;
this.localAddress = init.localAddress || input.localAddress || null;
```

README.md
```diff
@@ -217,6 +217,9 @@ The default values are shown after each option key.
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
 	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
+	lookup: dns.lookup  // custom dns lookup function for socket
+	family: null        // numbers 4 or 6 to limit dns resolution to ip v4 or v6
+	localAddress: null  // local interface to bind to for network connections
```
TimothyGu: Can you add commas after each property entry?

test/test.js
```diff
@@ -1531,6 +1533,58 @@ describe('node-fetch', () => {
...
+		const url = `${base}redirect/301`;
+		const recordedOptions = [];
+		const family = Symbol('family');
+		function lookupSpy(hostname, options, callback) {
+			recordedOptions.push(options);
+			return lookup(hostname, {}, callback);
+		}
+		return fetch(url, { lookup: lookupSpy, family }).then(() => {
+			expect(recordedOptions[0].family).to.equal(family);
+			expect(recordedOptions[1].family).to.equal(family);
```
TimothyGu: Maybe check the length of `recordedOptions` as well?

README.md
gr2m: I agree on adding the lookup option. 

I think the `family` option can be passed using an http agent though so we wouldn’t need to add another option for that? 

And I’m not sure what `localAddress` is needed for? Is it like bindAddress for servers? Why would a request library need it? The PR at https://github.com/bitinn/node-fetch/pull/256 does not have much info on it either. Maybe you could clarify it in the comment or add an explanation to the new node.js-only properties in the README for others who are puzzled by it?

bkw: While trying to find anything on how to specify the address family through an agent, I found an (admittedly old) [comment about the custom lookup only being used when not using an agent](https://github.com/nodejs/node/issues/4018#issuecomment-159600574)
Maybe we should investigate the interactions of these options a bit more. I'll set up some test cases tomorrow.

I can only guess that the use case for `localAddress` might be to influence routing by selecting the outgoing interface.

src/request.js
bkw: Done in 699fc3268f6cf60069d40f4fcad3ec4f812c685e, although I don't really agree too strongly here; we did not do this for `agent` either, and by passing in `dns.lookup` as the default we replicate the default value of a dependency, which might change in the future. Leaving it undefined when not specified would take a more neutral stance, imho. But it's not that important really, so I'm fine with the change if You are.

bkw:
Could someone with write access kindly clear the travis cache of the Node.js:8 FORMDATA=2.1.0 build? It is [failing](https://travis-ci.org/bitinn/node-fetch/jobs/393809846) [consistently](https://travis-ci.org/bitinn/node-fetch/jobs/393033610) without apparent reason while the same commit [builds fine in my account](https://travis-ci.org/bkw/node-fetch/builds/393809824) .

README.md
bkw: To clarify: The itch I'm scratching with this MR is really only the `lookup` option.
While implementing it I looked at other PRs and found that they touched the same code paths and would probably conflict with mine (and also not quite work as intended). So I folded them into this MR. In retrospect this probably was not as good an idea as I thought it was. I'll gladly reduce this back to just adding the lookup option and have it merged, instead of testing and reasoning about code by other people, which may not even be interested any longer.

gr2m:
> Could someone with write access kindly clear the travis cache

Done, and restarted the build

README.md
gr2m: > While trying to find anything on how to specify the address family through an agent, I found an (admittedly old) comment about the custom lookup only being used when not using an agent.

I had a similar experience when I researched the same thing for [`@octokit/rest`](https://github.com/octokit/rest.js). It used to have the request options `ca`, `proxy`, `family` and `rejectUnauthorized`. I think I wasn’t able to make the `family` option work myself with http agent, I only was able to create tests for the other three: https://github.com/octokit/rest.js/pull/783/commits/bfec75abfeb0154fa10be8b450fe072734c50bab. 

I still use an agent for the family option but I don’t actually know if it works:
https://github.com/octokit/rest.js/blob/7e5c8a501f25df1758e4bc2c162f6e307436062f/lib/get-request-agent.js#L20 ¯\\\_(ツ)\_/¯

So yeah, I don’t know. Nobody complained about it so far. Maybe in order to move things forward just implement the lookup option and leave out the other two? I’d not add them without having a clear use case that we agree on is within the scope of `node-fetch`

bkw:
I decided to drop the other patches and start a new PR with only `lookup`, although even that may not be necessary anymore, since the agent supports `lookup` as well (See #475). Please excuse the noise.

PR #475, opened 2018-06-19
support option `lookup`

bkw:
Predecessor of #474.

Please excuse the noise, see [discussion in #474](https://github.com/bitinn/node-fetch/pull/474/files/8629d027ace94290ae62859c80f33d1fc845c1eb#r196219236) about why `family` and `localAddress` are no longer part of this. Maybe #256 and #445 should be reopened.

@gr2m: Turns out, even `lookup` can be supplied via an agent. Shall we drop the whole thing? Passing in `lookup` via the agent is good enough for my use case, so feel free to just close.
OTOH, the whole path of agent options ending up in the socket options is really indirect and undocumented as far as I can see. Maybe we should at least mention it in the docs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,8 @@ The default values are shown after each option key.
 	timeout: 0,         // req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
-	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
+	agent: null,        // http(s).Agent instance, allows custom proxy, certificate etc.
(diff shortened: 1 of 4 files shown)


gr2m:
> Turns out, even lookup can be supplied via an agent. Shall we drop the whole thing?

Oh I’m glad you found that out, yeah I would suggest to drop it

> OTOH, the whole path of agent options ending up in the socket options is really indirect and undocumented as far as I can see. Maybe we should at least mention it in the docs?

Yes. How about we add documentation for it ourselves? Let’s at least add it to the comment in the options list

```
	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
```

And we could still merge the tests as a reference implementation, but using the agent option instead? That’s what I did for `@octokit/rest`, too

bkw:
Closing in favor of #476

PR #476, opened 2018-06-19
document and test using a custom dns lookup function through agent

bkw:
This is all that's left of #473, #474, and #475. Less code, less bugs. Yay!

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ The default values are shown after each option key.
 	timeout: 0,         // req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
-	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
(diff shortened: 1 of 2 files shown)


bkw:
Unrelated timeouts in some tests again.

Ironically, I'm pretty certain these could be improved by supplying a lookup function to an agent utilizing `dns.resolve` instead of `dns.lookup`, which was the point of the whole exercise in the first place... 😏

README.md
```diff
@@ -216,7 +216,7 @@ The default values are shown after each option key.
 	timeout: 0,         // req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
-	agent: null         // http(s).Agent instance, allows custom proxy, certificate etc.
+	agent: null         // http(s).Agent instance, allows custom proxy, certificate, lookup, family etc.
```
gr2m: Maybe only add `lookup`? I’m still not sure if family actually works with agent

bkw: I just added a test proving that it does indeed work.

PR #477, opened 2018-06-23
Headers get method returns combined value.

H1Gdev:
```js
let headers = new fetch.Headers();
headers.append('aaa', 'aaa');
headers.append('aaa', 'bbb');
headers.append('aaa', 'ccc');
let value = headers.get('aaa');
```
value is `aaa, bbb, ccc`.

```js
let headers = new fetch.Headers({
  'aaa': [ 'aaa', 'bbb', 'ccc' ]
});
let value = headers.get('aaa');
```
value is `aaa,bbb,ccc`.

```js
let headers = new fetch.Headers();
headers.set('aaa', ['aaa', 'bbb', 'ccc']);
let value = headers.get('aaa');
```
value is `aaa,bbb,ccc`.

**whatwg fetch spec**

```
3. Return the combined value given name and header list. 
```

I updated it to comply with this specification.

### DIFF
--- a/UPGRADE-GUIDE.md
+++ b/UPGRADE-GUIDE.md
@@ -41,7 +41,7 @@ spec-compliant. These changes are done in conjunction with GitHub's
 //////////////////////////////////////////////////////////////////////////////
 // `get()` now returns **all** headers, joined by a comma, instead of only the
 // first one. Its original behavior can be emulated using
-// `get().split(',')[0]`.
+// `get().split(', ')[0]`.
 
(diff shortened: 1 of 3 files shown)


H1Gdev:
@TimothyGu 

Thank you for your review.

Is not it necessary to adjust the first case to without space?

TimothyGu:
@H1Gdev No, because in the first case we are actually [combining](https://fetch.spec.whatwg.org/#concept-header-list-combine).

PR #478, opened 2018-06-24
Abort support patch

FrogTheFrog:
Hello,

I've made some minor changes to #437 in order to get signal events removed properly (I hope :)). Also, added some tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,7 @@ import Response from './response';
 import Headers, { createHeadersLenient } from './headers';
 import Request, {
 	removeAbortSignalCallback,
+	getAbortSignal,
 	getNodeRequestOptions
 } from './request';
 import FetchError from './fetch-error';
@@ -42,7 +43,8 @@ export default function fetch(url, opts) {
 	return new fetch.Promise((resolve, reject) => {
 		// build request object
 		const request = new Request(url, opts);
-		const signal = request.signal;
+		const signal = getAbortSignal(request);
+
 		if (signal.aborted) {
 			reject(new FetchError(`Fetch to ${request.url} has been aborted`, 'aborted'));
 			return;
@@ -146,7 +148,8 @@ export default function fetch(url, opts) {
 							agent: request.agent,
 							compress: request.compress,
 							method: request.method,
-							body: request.body
+							body: request.body,
+							signal
 						};
 
 						// HTTP-redirect fetch step 9
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,10 @@
 import Body, { writeToStream, getTotalBytes } from './body';
 import Response from './response';
 import Headers, { createHeadersLenient } from './headers';
-import Request, { getAbortSignal, getNodeRequestOptions } from './request';
+import Request, {
+	removeAbortSignalCallback,
+	getNodeRequestOptions
+} from './request';
 import FetchError from './fetch-error';
 
 const http = require('http');
@@ -39,7 +42,7 @@ export default function fetch(url, opts) {
 	return new fetch.Promise((resolve, reject) => {
 		// build request object
 		const request = new Request(url, opts);
-		const signal = getAbortSignal(request);
+		const signal = request.signal;
 		if (signal.aborted) {
 			reject(new FetchError(`Fetch to ${request.url} has been aborted`, 'aborted'));
 			return;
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -91,6 +100,7 @@ export default function fetch(url, opts) {
 
 		req.once('response', res => {
 			clearTimeout(reqTimeout);
+			removeSignalListeners();
```
FrogTheFrog: `removeSignalListeners()` is called on response and in `finallize()`. This should remove all signal events for `Request` object and `fetch` itself.

src/request.js
```diff
@@ -65,11 +61,11 @@ export default class Request {
...
 
-		if (init.signal != null) {
-			if (!isAbortSignal(init.signal)) {
+		if (init.signal !== undefined) {
+			if (init.signal !== null && !isAbortSignal(init.signal)) {
 				throw new TypeError('Provided signal must be an AbortSignal object');
 			}
-			signal = init.signal;
+			signal = init.signal === null ? undefined : init.signal;
 		}
```
FrogTheFrog: [whatwg](https://fetch.spec.whatwg.org/#request-class)(30) says that signal should be followed only if it's not `null`. I believe that when we provide `null`, we should no longer follow "parent's" signal. Or did I misinterpret it?

src/request.js
```diff
@@ -116,16 +114,17 @@ export default class Request {
 			redirect: init.redirect || input.redirect || 'follow',
 			headers,
 			parsedURL,
-			signal: abortController.signal
+			signal: abortController.signal,
+			onSignalAbort
 		};
```
FrogTheFrog: Storing `onSignalAbort` so that we could remove it later.

src/request.js
```diff
@@ -178,13 +177,31 @@ Object.defineProperties(Request.prototype, {
...
+}
+
+/**
+ * Removes the callback attached to a Request's signal "abort" event.
  *
  * @param   Request      A Request instance
- * @return  AbortSignal  request's signal
+ * @return  boolean      Indicates whether callback has been removed
  */
-export function getAbortSignal(request) {
```
FrogTheFrog: Since `signal` getter already exposes `signal`, I see no reason to have this method.

src/request.js
```diff
@@ -28,12 +28,17 @@ function isRequest(input) {
...
 
-// TODO: use toString check after https://github.com/mysticatea/abort-controller/pull/5
-function isAbortSignal(input) {
-	return (
-		typeof input === 'object' &&
-		input instanceof AbortSignal
-	);
+let isAbortSignal;
+if (typeof Symbol === "function" && typeof Symbol.toStringTag === "symbol") {
+	isAbortSignal = function(input) { return input.toString() === "[object AbortSignal]" };
```
FrogTheFrog: This fails for Node v4.

TimothyGu: This makes me think we should stop supporting Node.js v4.x if we want this feature.

src/request.js
TimothyGu: Eek, this is one of the areas the Web IDL spec is pretty unclear on. I think your handling is correct in this case, as I believe the spec writers wanted to distinguish null vs. undefined for exactly this purpose.

src/request.js
TimothyGu: We want to be safe against user overriding `signal` through `Object.defineProperty` or some other means.

TimothyGu:
This is great! I'll have to take another look at it, but thanks so much for spending the time to fix this.

src/index.js
TimothyGu: Hm, why are we removing the event listeners on response? We want the abort controller to work even after the initial response has been received… or am I missing something?

src/request.js
FrogTheFrog: Since the current Request has been resolved, we must remove it from signal's event listeners so it can be garbage collected. Unless I missed something, Response cannot be aborted. I did, however, forgot to propagate signal when redirecting.

src/index.js
FrogTheFrog: Since the current Request has been resolved, we must remove it from signal's event listeners so it can be garbage collected. Unless I missed something, Response cannot be aborted. I did, however, forgot to propagate signal when redirecting.

FrogTheFrog:
@TimothyGu I've addressed some of your concerns also did a little method renaming and reordering.

bitinn:
Hi @FrogTheFrog, we are working on getting AbortSignal support in fetch, if you got any feedback, let us know at #539, thx!

FrogTheFrog:
@bitinn Closing in favor of #539 :)

Issue #427, opened 2018-03-21
FetchObserver (for a single fetch)


jimmywarting:
Not implemented in the spec yet(?) But lets keep an open issue until it's final and implement it later

For the moment it looks to work something like this:

```js
fetch(url, {
  observe(observer) { 
    observer.onresponseprogress = e => console.log(e);
    observer.onrequestprogress = e => console.log(e);
    observer.onstatechange = n => console.log(observer.state)
  }
)
```

TimothyGu:
Is there a spec issue for this?

jimmywarting:
https://github.com/whatwg/fetch/issues/607

PR #480, opened 2018-06-30
Add onrequest option

akihikodaki:
`onrequest` option allows to intercept `http.ClientRequest`. It is especially
useful for HTTP Signatures or similar authentication mechanism which
require exact HTTP header representation.

This change is radical as you can see. Let me know if you prefer an alternative.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,8 +217,22 @@ The default values are shown after each option key.
 	compress: true,     // support gzip/deflate content encoding. false to disable
 	size: 0,            // maximum response body size in bytes. 0 to disable
 	agent: null         // http(s).Agent instance, allows custom proxy, certificate, lookup, family etc.
+	onrequest: null     // a callback to be called with `http.ClientRequest`.
 }
 ```
+
+###### Intercepting `http.ClientRequest` with `onrequest` callback
+
+`onrequest` callback will be called with underlying `http.ClientRequest` as
+the first argument immediately before flushing. It is assured that the following
+operations on `http.ClientRequest` can be performed:
+
+* Modifying `maxHeadersCount` property to inspect or change the maximum response headers count.
+* Calling `removeHeader` and `setHeader` method to modify any headers to be sent.
+* Calling `setNoDelay`, `setSocketKeepAlive`, and `setTimeout` method, expecting the documented behavior.
+
+Other operations with side effects on `http.ClientRequest` can cause an
+unexpected behavior.
 
 ##### Default Headers
 
--- a/src/index.js
+++ b/src/index.js
@@ -202,6 +202,10 @@ export default function fetch(url, opts) {
 			resolve(new Response(body, response_options));
 		});
 
+		if (request.onrequest) {
+			request.onrequest(req);
+		}
+
 		writeToStream(req, request);
 	});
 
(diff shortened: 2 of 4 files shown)


TimothyGu:
Can you specify more in detail what your use case for this is?

akihikodaki:
You need this change if you want to sign headers set or modified by `node-fetch`. `Content-Length` is an important header field to be signed, for example.

It also means you need to know whether a header field will be modified by `node-fetch` or not to sign it. Making such assumption may introduce incompatibility with future version of `node-fetch`. You do not make any assumptions for the headers if we have `onrequest` option and therefore it is *safer* for HTTP signature.

jimmywarting:
Would #427 (FetchObserver) solve your problem?
We might eventually have a spec standard for observing the states.

akihikodaki:
> Would #427 (FetchObserver) solve your problem?

I guess no. To sign HTTP request, you have to:
1. Intercept header fields to calculate signature (I guess FetchObserver does this)
2. Add HTTP signature header field (I guess FetchObserver doesn't do this)

So we have to consider a solution for the problem, maybe this change or something else.

jimmywarting:
I'm only speculating now. 
I have no idea if the requesting state should fired right before `http.request()` is called or if it's fired right after when the request have been made. I also don't know what will be available on the event and if it will be possible to change something before it fires

But if the requesting-state is fired right before `http.request` is made, maybe then it would be possible to make small adjustments to the request before it sets of

```js
/*
enum FetchState {
  // Pending states
  "requesting", "responding",

  // Final states
  "aborted", "errored", "complete"
};
*/

fetch(url, {
  observe(observer) { 
    observer.onstatechange = evt => {
      // begin requesting
      if (evt.state === 'requesting') {
        const request = evt.target
        // calculate signature
        const value = calculateSignature([...request.headers])
        // Add HTTP signature header before `http.request()` is called
        request.headers.set(key, value)
      }
    }
  }
)
```

akihikodaki:
@jimmywarting 
If you want a real use case, see https://github.com/miniorg/miniverse/blob/c6d758781c14da1c0293469258e3c99aa48ceab8/lib/transfer.js#L91. It is quite a straightforward:
```JavaScript
  const response = await fetch(repository, uri, {
    method: 'POST',
    body: JSON.stringify(activityStreams),
    headers: { 'Content-Type': 'application/ld+json; profile="https://www.w3.org/ns/activitystreams"' },
    size: -1,
    onrequest(request) {
      sign(request, {
        authorizationHeaderName: 'Signature',
        key: privateKeyPem,
        keyId
      });
    }
  });
```

`requesting`-state looks fine. The only concern is that it is impossible or difficult to implement asynchronous hook. For example, Node.js has asynchronous functions and what would happen if you want to use them to sign HTTP request? It was easy to add code to await them with dumb `onrequest` callback, but it is not for your suggestion.
To clarify, `onrequest` does not only have a feature to intercept incoming requests, but also has a feature to block while modifying requests. That differentiates `onrequest` from other events.

> Also what is preventing you from calculating the signature and setting the header before making the request?

See https://github.com/bitinn/node-fetch/pull/480#issuecomment-406912814.


> You need this change if you want to sign headers set or modified by `node-fetch`. `Content-Length` is an important header field to be signed, for example.

PR #496, opened 2018-07-26
Support separate timeouts for network timeout and body timeout

aakula-edmunds:
Support separate timeouts for network timeout and body timeout instead of using the same value

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -145,7 +145,7 @@ export default function fetch(url, opts) {
 				statusText: res.statusMessage,
 				headers: headers,
 				size: request.size,
-				timeout: request.timeout
+				timeout: opts && opts.bodyTimeout || request.timeout
 			};
 
 			// HTTP-network fetch step 12.1.1.3
--- a/test/test.js
+++ b/test/test.js
@@ -756,6 +756,20 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should allow separate custom timeout on response body different from request timeout', function() {
+		this.timeout(500);
+		const url = `${base}slow`;
+		const opts = {
+			bodyTimeout: 200,
+		};
+		return fetch(url, opts).then(res => {
+			expect(res.ok).to.be.true;
+			return expect(res.text()).to.eventually.be.rejected
+				.and.be.an.instanceOf(FetchError)
+				.and.have.property('type', 'body-timeout');
+		});
+	});
+
 	it('should clear internal timeout on fetch response', function (done) {
 		this.timeout(2000);
 		spawn('node', ['-e', `require('./')('${base}hello', { timeout: 10000 })`])


src/index.js
```diff
@@ -145,7 +145,7 @@ export default function fetch(url, opts) {
 				statusText: res.statusMessage,
 				headers: headers,
 				size: request.size,
-				timeout: request.timeout
+				timeout: opts && opts.bodyTimeout || request.timeout
```
jimmywarting: what differences dose this make?
The only thing you have done is making an alias for `opts.timeout` and added a test that probably already exist somewhere. 

Also I would discourage the use of such options and instead use the AbortController when it lands since timeout isn't part of the spec

For Example

# using abortController to control timeout for initial response header

```js
const controller = new AbortController()
const { signal } = controller
const timer = setTimeout(controller.abort, 5000)

fetch(url, { signal })
  .then(res => {
    // everything uploaded, starting to download something

    // Stop the timer from aborting something we are downloading
    clearTimeout(timer)
  })
```

# Or using abort controller with timeout for the hole upload and download

```js
const controller = new AbortController()
const { signal } = controller
const timer = setTimeout(controller.abort, 5000)

fetch(url, { signal })
  .then(res => {
    // everything uploaded, starting to download something
    // now waiting for the final response (leaving the timer be)
    return res.text()
  }).then(res => {
    // downloaded everything in time
  })
```


# Or starting a timeout if something you are downloading is very slow

```js
const controller = new AbortController()
const { signal } = controller

fetch(url, { signal })
  .then(res => {
    // everything uploaded, starting to download something
    // abort if it's going to slow
    const timer = setTimeout(controller.abort, 5000)
    // can use content-length to figure out roughly how long it shold be
    return res.text()
  }).then(res => {
    // downloaded everything in time
  })
```

jimmywarting:
Closing this since `AbortController` will be able to solve this problem and the fack that timeout is not an spec standard

blobcat:
This should not have been closed. As far as I can tell AbortController and signal are not supported by node-fetch. You can only use that with fetch in a browser environment, but in node you still need to use the timeout node-fetch extension if you don't want requests to go on indefinitely.

@jimmywarting, I think you missed the point of this PR. Currently node-fetch has a single timeout option, but that timeout is applied independently to two separate stages of the request. First it is applied to the connection and request, and then later it is applied again when the response is being read. The main issue with this is that the same timeout value may not be suitable for both stages.

For example, if you set your timeout to 50ms, and the endpoint you're calling is fast to respond, the timeout makes sense. However, if that fast endpoint returns a larger payload the 50ms timeout may be breached just in the response reading stage. This will result in a fetch that was actually completed but still failed.

I myself have run into this exact use case in node-fetch on multiple occasions, and while I understand the reasoning behind enforcing timeout for response body processing, I also feel like it should be treated separately. This PR solves that by introducing a separate, optional timeout to be specified for processing the response body. If you don't use it, then node-fetch would default to the old behavior. Exactly what I need and, from my perspective, doesn't go against standards any more than the timeout extensions already does.

jimmywarting:
> AbortController and signal are not supported by node-fetch.

They will be eventually. there is a PR open trying to adress this.
We want to be as spec compatible as possible.

PR #504, opened 2018-08-15
README 2.x update proposal... take 2

jkantr:
A continuation of the changes and discussion in #448 

This time without broken git history :)

The beginning of a general overhaul of docs to be easier to read, suggest best practices, fix some verbiage, and bring up to date with v2.x

Most suggestions from #448 have either been implemented or added as "comments" in the markdown (TODO 'links')

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ A light-weight module that brings `window.fetch` to Node.js
 - [API](#api)
     - [fetch(url[, options])](#fetchurl-options)
     - [Options](#options)
-    - [Class: Request](#class-request)Additional changes made as per discussion in #448
+    - [Class: Request](#class-request)
     - [Class: Response](#class-response)
     - [Class: Headers](#class-headers)
     - [Interface: Body](#interface-body)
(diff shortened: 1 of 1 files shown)


bitinn:
Thx @jkantr, I gave this a quick read and it looks good to me.

@jimmywarting @TimothyGu please give this a quick read when you can, I intend to merge it :)

bitinn:
Actually one quick thing:

Can you restructure this code block into a table (just like the default header table below).

I always think people overlooks `agent` options because they are very much hidden in docs, but it's the one option that workaround many problems.

https://github.com/bitinn/node-fetch/blob/2d373bcb77af75876407fdee64fc572e3a08c11a/README.md#options

jkantr:
@bitinn do you mean something like this? 
https://gist.github.com/jkantr/a8036d22b6aaec0626ae2340a1aaf1e2

Tables are pretty awkward to work with in `.md` unfortunately... maybe we could just have a separate sub header about what problems can be solved by using manual agents?

jkantr:
Colspan doesn't really work it seems (can't span the first col, only the ones after) in GFM so this was as close as I can get https://gist.github.com/jkantr/a8036d22b6aaec0626ae2340a1aaf1e2

It may be better just to stuff them in a table and explain each in its own `####` section after

bitinn:
@jkantr @jimmywarting @TimothyGu hey guys, last call, I am going to merge this PR, as is, please give OP a thumb up if you feel this is ok. And stop me if you see anything straight up wrong.

jimmywarting:
It was good.
You took all other feedback and consideration into account and forged a darn good documentation, like the hole ToC thing!

I didn't read carefully in the end just shimmed thought the rest. Will just say that all looks alright!
So I thought it became to much at the end... there where a lot of things packed into one readme file.
So here are my thoughts: Have a Readme that describes what the fetch api is (and isn't) and then use the github's wiki page

styfle:
> there where a lot of things packed into one readme file

The way we solve this problem with [marked](https://github.com/markedjs/marked) is to put the description and a simple getting started snippet in the README (this gets installed millions of times so we keep it small).

The README has a link to [GitHub pages](http://marked.js.org) which hosts the full documentation (stored in `/docs` in the repo).

PR #510, opened 2018-08-24
support for react-native

NikolaBorislavovHristov:


TimothyGu:
This is unfortunately unacceptable in its current form. We have already switched from `window` to `self` for Web Worker support, and I'm tired of seeing yet another similar change that technically doesn't work with Node.js' flavor of CommonJS, as the `this` value in a module is set to `exports`. How about something like

```js
const global = new Function('return this')();
```

and use `global` for the rest of the file?

jimmywarting:
> How about something like
> ```js
> const global = new Function('return this')();
> ```

Then someone might complain about Content Security Policy (CSP) - unsafe-eval

```js
new Function(code)
eval(code)
```

It's so fun to be a javascript developer this days... 😑

bitinn:
@jimmywarting to be fair, we were advocating for not putting any polyfill for browser fetch, but people keep asking, so there you go. This will never be quite right, like most things in universal javascript dream.

jimmywarting:
after reading [How to get the global object in JavaScript?](https://stackoverflow.com/questions/3277182/how-to-get-the-global-object-in-javascript)

I found out using Function and eval are the best way to go since it can escape whatever you declare above the script

So i liked **CoolAJ86** [answer](https://stackoverflow.com/a/6930376/1008999) the most

but didn't like the ending of using `global = window` if someone where to write `let window = {}` above

So **Remo H. Jansen** [answer](https://stackoverflow.com/a/33268326/1008999) where pretty good type checking but i revamped it to safer type checking with a fallback to `this`.

So this is what i ended up with at the end:

```js
var global = (function() {
  try {
    return Function('return this')() || (42, eval)('this')
  } catch(e) {
    // every global should have circular reference
    // used for checking if someone writes var window = {}; var self = {}
    return typeof window === "object" && window.window === window ? window : 
    typeof self === "object" && self.self === self ? self :
    typeof global === "object" && global.global === global ? global : this
  }
})()
```

puh, That's crazy!

NikolaBorislavovHristov:
Why not
```

**Comment by NikolaBorislavovHristov** [comment]
I've manage to workaround my problem with this statement
```javascript
global.self = global;
```

NikolaBorislavovHristov:
UPDATED: Using the same approach like   [setImmediate](https://github.com/YuzuJS/setImmediate/blob/master/setImmediate.js#L186)

NikolaBorislavovHristov:
Any update on this?
I don't know why the ci fails with node 4.

bitinn:
I haven't looked into this as I don't use react-native but can we somehow come up with some tests that validate what we are doing? `browser.js` has been changed many times, at this point I want to see confirmation that this doesn't somehow break webpack or other usages.

Also see cross-fetch which already tested react-native integration: https://github.com/lquixada/cross-fetch

bitinn:
Hi @NikolaBorislavovHristov, I am closing this in favor of #545, please let me know if it works, as I can't test every environment on my end or in CI.

PR #511, opened 2018-08-28
📝 Minor update to fix #412

### LINKED ISSUE #412: .textConverted needs to install the package encoding separately

motss:
See #412 for more details.


bitinn:
If you don't mind, I will let you know when we merge #504, then I will add this in.

motss:
@bitinn No problem. Just let me know when to update it to the latest.

bitinn:
readme was updated, but thx for your PR!

PR #515, opened 2018-09-10
Bind self.fetch to self

KSXGitHub:
To avoid "Illegal invocation" error


bitinn:
Hi @KSXGitHub, I am closing this in favor of #545, please let me know if it works, as I can't test every environment on my end or in CI.

PR #518, opened 2018-09-12
Fixed html 4 charset detection

jacktuck:
Old regex worked for`<meta http-equiv="Content-Type" content="text/html; charset=gb2312">`
but not `<meta content="text/html; charset=gb2312" http-equiv="Content-Type">`

old regex: https://regex101.com/r/y6hyRj/1
new regex: https://regex101.com/r/wb4AFp/1

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -305,11 +305,7 @@ function convertBody(buffer, headers) {
 
 	// html4
 	if (!res && str) {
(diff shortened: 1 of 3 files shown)


bitinn:
Thx for this, I am putting this on hold for now, as `textConverted()` is really a legacy API that's created to be backward compatible with node-fetch v1.

It's not a part of Fetch Spec and ideally we want to phase out this API and create a simple module that takes `res.body` and output exactly the same UTF-8 plain text.

If you ended up making such a module please let us know.

jimmywarting:
ok with close this?

Issue #517, opened 2018-09-11
implement Response.redirect


jimmywarting:
`Response.redirect` is missing

Need something like this:
```js
static redirect (location, status = 302) {
  return new Response('', { 
    headers: { location }, 
    status 
  })
}
```

https://fetch.spec.whatwg.org/#dom-response-redirect

xxczaki:
Closed by #718

PR #519, opened 2018-09-13
Add "response.redirected" property support

410stats:
As described in https://developer.mozilla.org/en-US/docs/Web/API/Response/redirected
If any redirection happened, it will be equal to true. Otherwise it will be equal to false.

There is two basic test is the commit to cover this property.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -145,7 +145,8 @@ export default function fetch(url, opts) {
 				statusText: res.statusMessage,
 				headers: headers,
 				size: request.size,
-				timeout: request.timeout
+				timeout: request.timeout,
+				redirected: request.redirected
 			};
 
 			// HTTP-network fetch step 12.1.1.3
--- a/src/request.js
+++ b/src/request.js
@@ -115,6 +115,13 @@ export default class Request {
 		return this[INTERNALS].redirect;
 	}
 
+	get redirected() {
+		if (this.counter > 0)
+			return true;
+		else
+			return false;
+	}
+
 	/**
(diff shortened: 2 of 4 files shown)


jimmywarting:
Mistakenly thought this to solve my recently added issue #517  😜

bitinn:
I think this is largely fine to end users, though it seems the spec has changed quite a lot, there is a URL List now which `redirected` getter is required to check. 

https://fetch.spec.whatwg.org/#concept-response-url-list
https://fetch.spec.whatwg.org/#dom-response-redirected

410stats:
> I think this is largely fine to end users, though it seems the spec has changed quite a lot, there is a URL List now which `redirected` getter is required to check.
> 
> https://fetch.spec.whatwg.org/#concept-response-url-list
> https://fetch.spec.whatwg.org/#dom-response-redirected

Thanks for the link !
To follow the specs, we should have in Response object an internal array containing URL of previous redirection (if i understood correctly, as i cannot find an example of implementation...).
It should be used also for the Response.url method that need to return the last element of this array.

So it's really doable, but this implies more internal modifications and thus more testing than this little commit. And Response.redirected is the only way to know that a redirection happened without handling redirection manually ...
For the moment i think we should push this and modify methods in a later commit when the internal URL list will be added to respect the specs 👍

jimmywarting:
Just needed this feature atm.

bitinn:
sorry for the late update, this is now available in v2.5.0, thx @410stats !

PR #520, opened 2018-09-15
Changed how URLSearchParams are handled 

jimmywarting:
I notice there where a problem reading the body of req/res constructed with URLsearchParams since that if condition where missing

https://github.com/bitinn/node-fetch/blob/1d4ab5a0de0df991f92f1f30efd5f4815bbcdb71/src/body.js#L176-L220

Posting it doe works when using fetch. Not sure how but it works 😕

Then i notice that responses didn't set default content-types so I moved the logic of sniffing the content-type into the Body mixin instead

---

So I chose not to not to add a if condition that would allow to read the content of URLSearchParams instead i just converted it to a buffer directly so i could solve 5a62bbf1598665bdf2a5110a51a996edb8e4a490 (mutation of URLSearchParam)

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -410,7 +410,7 @@ export function clone(instance) {
  *
  * @param   Mixed  instance  Response or Request instance
  */
-export function extractContentType(body) {
+function extractContentType(body) {
 	// istanbul ignore if: Currently, because of a guard in Request, body
 	// can never be null. Included here for completeness.
 	if (body === null) {
--- a/src/body.js
+++ b/src/body.js
@@ -26,7 +26,7 @@ const INTERNALS = Symbol('Body internals');
 export default function Body(body, {
 	size = 0,
 	timeout = 0
-} = {}, headers) {
+}, headers) {
 	if (body == null) {
 		// body is undefined or null
 		body = null;
--- a/src/body.js
+++ b/src/body.js
@@ -26,6 +26,9 @@ const INTERNALS = Symbol('Body internals');
 export default function Body(body, opts, headers) {
 	const { size = 0, timeout = 0 } = opts;
 
+	const haveType = headers.has('Content-Type');
+	let contentType = extractContentType(body);
+
 	if (body == null) {
 		// body is undefined or null
 		body = null;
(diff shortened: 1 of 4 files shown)


jimmywarting:
Also changed strings and Blob to buffer so less if checks can be used by consumeBody

jimmywarting:
It started out with that it wasn't possible at reading the body of something like this:
```js
new Response(new URLSearchParams("a=1")).text()
```
then it didn't set the correct content-type either (only Request did that)

what i have done to solve it was

- Converting the body input in [src/body.js#L26](https://github.com/bitinn/node-fetch/blob/1598b919c16ade23ddf9021c416169495bc5fbb3/src/body.js) directly into a buffer (for 4 reasons).
  - This removes the reference to the URLSearchParams. b/c it shouldn't be possible to change content of `Request` or the `Response` content by changing the content of URLSearchParams afterwards.
  - I also didn't want to create a extra if-condition in the `consumeBody` function to handle yet another data type, unifying most of the body to a buffer directly will decrease the complexity of the code
  - The Request class was first responsible for detecting the `Content-Type` (it called the `extractContentType` function which Response nerver did. this was probably a miss.) Now that is handle by both in a single place in the Body function)
  - 4th reason is b/c I wanted to solve the #210. That is if you call `req_or_res.body` you should get a stream and this is a step towards that solution (I think) (i tried to solve this in my code by monkey patching the body getter function but doing so broke `Content-Type` sniffing and that is what is putting me at a production halt) 
- b/c was changing it to a buffer it wasn't able to figure out the `Content-Type` of the input after it had been converted to a buffer. That is why i incorporated the `extractContentType` into the `Body` function so that it can do the `Content-Type` extraction at the same time. 
  - To do so i also needed to pass the Headers down to the `Body` function


This is more similar of what the fetch polyfill dose https://github.com/github/fetch/blob/08bc7423c9d26dd9f21905fee3b83660723e2d37/fetch.js#L208-L242 you can see that they also clone the URLSearchParams
```
    } else if (support.searchParams && URLSearchParams.prototype.isPrototypeOf(body)) {
      this._bodyText = body.toString()
```

jimmywarting:
ping @TimothyGu

jimmywarting:
What put the development on a halt? busy with other things now?

PR #521, opened 2018-09-18
Provide global polyfills

jaydenseric:
Often it is desirable to globally polyfill the `fetch` API in a Node.js server environment, particularly for SSR.

With this PR, you can easily polyfill `fetch`, `Headers`, `Request` and `Response` in global scope:

```js
require('node-fetch/polyfill')
```

Or, with `--experimental-modules`:

```js
import 'node-fetch/polyfill'
```

You can even polyfill a Node.js process via CLI:

```
node -r node-fetch/polyfill server.js
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,12 @@ We suggest you load the module via `require`, pending the stabalizing of es modu
 const fetch = require('node-fetch');
 ```
 
+To polyfill `fetch`, `Headers`, `Request` and `Response` in global scope:
+
+```js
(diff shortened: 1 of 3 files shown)


bitinn:
Thx for the PR but I believe this is better left for third-party module, where the author knows exactly what they want.

(we made mistake providing a `browser.js` and we regret it since, as people use it differently and sometimes in non-nodejs environment like react-native.)

PR #528, opened 2018-10-02
feat: Add unix socket support

mkg20001:
Adds support for requests over unix sockets, similar to how npm.im/request does it
Example: `fetch('http://unix:/var/lib/lxd/unix.socket:/')` does a `GET /` on `/var/lib/lxd/unix.socket`

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -149,7 +149,7 @@ Object.defineProperties(Request.prototype, {
  * @return  Object   The options object to be passed to http.request
  */
 export function getNodeRequestOptions(request) {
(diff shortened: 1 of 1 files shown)


bitinn:
Thx for the PR but we have no plan to support unix socket yet.

However, check if this approach allow you to use unix socket through passing in an Agent.

elmerbulthuis:
Would love to see unix socket upport! Also, why not?

Our particular use case is communication with the docker engine. We developed an openapi client that is using node-fetch. One of the clients we generate is for the docker engine. This engine usually listens to a unix socket.

PR #529, opened 2018-10-04
Remove useless call to `Error` constructor

lpinca:
The function call `Error()` is equivalent to the object creation
expression `new Error()` with the same arguments.

Refs: https://www.ecma-international.org/ecma-262/5.1/#sec-15.11.1


Richienb:
@lpinca The custom errors were recently modernised. If this problem still persists, please create another PR against the 3.x branch to address it.

PR #530, opened 2018-10-04
feat: Allow passing .cert and .key option for client side authentication

mkg20001:
See desc.
Note: More of a quick'n'dirty fix for https://npm.im/lxdn than a real pr


bitinn:
you should look into `https.Agent`, which allow you to specify custom cert and we do allow passing in agent.

mkg20001:
Ok, thanks

Issue #534, opened 2018-10-15
error: self is not defined after webpack build


rainmanhhh:
```ts
import fetch from 'node-fetch'

// todo data type
export async function callApi(data: ApiRequest) {
  console.log('call api with data:', data)
  const response = await fetch(config.apiUrl, {
    method: 'POST',
    headers: {
      'Content-Type': 'x-www-form-urlencoded'
    },
    body: 'data=' + encodeURIComponent(JSON.stringify(data))
  })
  const resp = await (response.json() as Promise<ApiResponse>)
  if (resp.status === Status.Fail) {
    console.error('ApiResponse status is failed:', resp)
    throw new Error('ApiResponse status is failed:' + resp.codeDesc)
  }
  return resp.responseContent
}
```

code above works well with tsc, but after webpack building, it reports:
```
ReferenceError: self is not defined
    at eval (webpack:///./node_modules/node-fetch/browser.js?:3:28)
    at Object../node_modules/node-fetch/browser.js (E:\workspace\wjw\dist\app.js:97:1)
    at __webpack_require__ (E:\workspace\wjw\dist\app.js:20:30)
    at eval (webpack:///./src/api.ts?:43:36)
    at Object../src/api.ts (E:\workspace\wjw\dist\app.js:164:1)
```

node version 10.12
module version:
    "typescript": "^3.1.3",
    "webpack": "^4.20.2",
    "webpack-cli": "^3.1.2"

  "dependencies": {
    "node-fetch": "^2.2.0",

jimmywarting:
why not use `window.fetch` instead of importing node-fetch?

rainmanhhh:
> why not use `window.fetch` instead of importing node-fetch?

it's a server-side app, not running on browser

bitinn:
Hi @rainmanhhh, please checkout #545 and see if it works.

PR #537, opened 2018-10-18
Polyfill self

brunolemos:
Fix support for environments that doesn't have a `self` global variable, including but not limited to react native.

Alternative implementation to #510
Might or might not help #534 

Closes #510

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,5 +1,15 @@
 "use strict";
 
+function getSelf() {
+  if (typeof self !== 'undefined') return self;
+
+  if (typeof this !== 'undefined') return this;
+  if (typeof global !== 'undefined') return global;
+  if (typeof window !== 'undefined') return window;
+  throw new Error('Failed to access global variable "self".');
+}
+
+;(function(self) {
 module.exports = exports = self.fetch;
 
 // Needed for TypeScript and Webpack.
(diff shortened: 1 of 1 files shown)


jimmywarting:
close this for the same reason as https://github.com/bitinn/node-fetch/pull/521#issuecomment-435822248?

> Thx for the PR but I believe this is better left for third-party module, where the author knows exactly what they want.
>
> (we made mistake providing a browser.js and we regret it since, as people use it differently and sometimes in non-nodejs environment like react-native.)

a solution for the author could be just to do this in top of their own code?

```js
this.self = this
```

bitinn:
@jimmywarting I would love to close them, BUT, since we have already made this mistake, it's now our job to make it better somehow.

I don't think we can reliably convince everyone to just use this:

```
this.self = this // or
global.self = global
```

So how about this?

```
var getGlobal = function () {
	// the only reliable means to get the global object is
	// `Function('return this')()`
	// However, this causes CSP violations in Chrome apps.
	if (typeof self !== 'undefined') { return self; }
	if (typeof window !== 'undefined') { return window; }
	if (typeof global !== 'undefined') { return global; }
	throw new Error('unable to locate global object');
}
```

Without automated CI test, I can't reliably say it works, but I think it does, and don't have crazy drawbacks.

ref: https://github.com/tc39/proposal-global

bitinn:
Hi @brunolemos, I am closing this in favor of #545, please let me know if it works, as I can't test every environment on my end or in CI.

PR #538, opened 2018-10-18
Add support for node-specific options to http(s).request

mvachhar:
This PR will allow users to specify nodeOptions to fetch for use directly in the http(s).request call.  This is especially helpful for https and client certificates, server validation etc.  Syntax is as follows:
```
fetch("http://example.org", { 
  nodeOptions: {
    setHost: false
});
```
For TLS client certificate authentication you could do 
```
fetch("https://example.org", {
  nodeOptions: {
    key: fs.readFileSync('test/fixtures/keys/agent2-key.pem'),
    cert: fs.readFileSync('test/fixtures/keys/agent2-cert.pem')
  }
}
```

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -36,6 +36,7 @@ function isRequest(input) {
 export default class Request {
 	constructor(input, init = {}) {
 		let parsedURL;
+		let nodeOptions = init.nodeOptions;
 
 		// normalize input
 		if (!isRequest(input)) {
@@ -51,6 +52,7 @@ export default class Request {
 			input = {};
 		} else {
 			parsedURL = parse_url(input.url);
+			nodeOptions = nodeOptions || input[INTERNALS].nodeOptions;
 		}
 
 		let method = init.method || input.method || 'GET';
@@ -85,7 +87,8 @@ export default class Request {
 			method,
 			redirect: init.redirect || input.redirect || 'follow',
 			headers,
-			parsedURL
+			parsedURL,
+			nodeOptions
 		};
 
 		// node-fetch-only options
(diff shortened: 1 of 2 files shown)


jimmywarting:
This kind of pr/feature request have usually been rejected for a couple of times (See #15 for example)

Reason being that we want to be as close to the spec as possible.

If you want to set a custom cert, you can do that with a custom Agent

```js
const agent = new https.Agent({
  key: fs.readFileSync(`${CERT_PATH}.key`),
  cert: fs.readFileSync(`${CERT_PATH}.crt`),
  rejectUnauthorized: false
})

fetch(url, { agent })
```

----

There have been a couple of request to expose the request or make node fetch modular somehow. even being more event driven. (#284) to customize/debug the underlying http

----

If we look into #427 (FetchObserver (for a single fetch))

then we might have some playground to do this kind of things.

```js
fetch(url, {
  observe(observer) { 
    observer.onstatechange = evt => {
      if (evt.state === 'requesting') {
        // do something with `evt` or `evt.target` and modify something
      }
    }
  }
)
```

mvachhar:
The agent works for me.  Thanks!

jimmywarting:
So it's okay to close this?

mvachhar:
Yeah, that should be fine.  Although, agent is non-standard as well right?
Since it allows manipulation of the connection, I think my pull request has
no use unless you really want to suppress the host header like I do in the
unit tests in this PR.  Am I missing a use case that this PR would enable
that agent would not with the current state of Node?

jimmywarting:
>  agent is non-standard as well right?

Yea, it has existed from almost the very first beginning

> Am I missing a use case that this PR would enable
that agent would not with the current state of Node?

Don't know

Issue #523, opened 2018-09-20
Should we deprecate timeout in favor of AbortController?


jimmywarting:
Timeout was a node-fetch only additions in it's early days and was never implemented in the spec. Now that there is a way to control it using the `AbortController` to be able to control when it should abort a request. One could control whether or not a timeout should affect the hole request and response or one or the other

what would you think about removing it and instead help out on creating a abort signal? 

(thinking maybe deprecate `timeout` in v2 or v3 - and remove it completely in v3 or v4)

To help out other on a migration path I where thinking maybe of some way to include a helper function that is optional and isn't included in the bundle so it makes it more modular.

some proposal:
 
```js
const timeout = require('node-fetch/timeout')
const signal = timeout(1000)
/*
  Same as doing
  const controller = new AbortController()
  const { signal } = controller
  setTimeout(controller.abort, 1000)
*/
fetch(url, { signal })
fetch(url, { signal: timeout(1000) })
```

Or if it returns a object then you can use spread to append it
```js
const timeout = require('node-fetch/timeout')
console.log( timeout(1000) ) // { signal: signal }

fetch(url, { headers, ...timeout(1000) })
```

Or returning a object that can help you start and stop the a singal to be able to control if the timeout should apply to only the upload part or the download part
```js
// what node-fetch/timeout would look like
const AbortController = require('./AbortController')

module.exports = function timeout(delay) {
  const controller = new AbortController()
  const { signal } = controller
  let timer = null

  return {
    signal,
    start() {
      clearTimeout(timer)
      timer = setTimeout(controller.abort, delay)
      return signal
    }
    stop() {
      clearTimeout(timer)
    }
  }
} 
```
```js
const timeout = require('node-fetch/timeout')
const timer = timeout(1000)

// for it's hole duration
fetch(url, { signal: timeout(1000).start() })

// just for upload
const timer = timeout(1000)
fetch(url, { signal: timer.start() })
  .then(res => {
    timer.stop()
  })


// just for download
const timer = timeout(1000)
fetch(url, { signal: timer.signal })
  .then(res => {
    timer.start()
  })
```

bitinn:
I think 80-90% people are happy with `timeout` as it is, so it depends on how easy to use our replacement will be (AbortController isn't the most intuitive thing), if it's indeed as simple as `signal: timeout(1000)`, sure, we might deprecated `timeout` in future.

jnields:
[...]

Issue #498, opened 2018-07-31
Throw SyntaxError in case of invalid JSON


TimothyGu:
Basically, revert #300 (#293). This change broke compatibility with browsers which throw a `SyntaxError`.

xxczaki:
Fixed in #700

PR #539, opened 2018-10-23
Add support for AbortSignal to cancel requests

jnields:
Whether or not you actually deprecate timeout should probably be a separate issue (#523) - but I think at the very least there should be support for `AbortSignal` here.

Let me know if you would need any changes to make this happen!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ import Response from './response';
 import Headers, { createHeadersLenient } from './headers';
 import Request, { getNodeRequestOptions } from './request';
 import FetchError from './fetch-error';
+import AbortError from './abort-error';
 
 // fix an issue where "PassThrough", "resolve" aren't a named export for node <10
 const PassThrough = Stream.PassThrough;
@@ -46,13 +47,40 @@ export default function fetch(url, opts) {
 		const options = getNodeRequestOptions(request);
 
 		const send = (options.protocol === 'https:' ? https : http).request;
+		const { signal } = request;
+		let response = null;
+
+		const abort = ()  => {
+			let error = new AbortError('The user aborted a request.');
+			reject(error);
+			if (request.body && request.body instanceof Stream.Readable) {
+				request.body.destroy(error);
+			}
+			if (!response || !response.body) return;
+			response.body.emit('error', error);
+		}
+
+		if (signal && signal.aborted) {
+			abort();
+			return;
+		}
+
+		const abortAndFinalize = () => {
+			abort();
+			finalize();
+		}
 
 		// send request
 		const req = send(options);
 		let reqTimeout;
 
+		if (signal) {
+			signal.addEventListener('abort', abortAndFinalize);
+		}
+
 		function finalize() {
 			req.abort();
+			if (signal) signal.removeEventListener('abort', abortAndFinalize);
 			clearTimeout(reqTimeout);
 		}
 
@@ -117,7 +145,8 @@ export default function fetch(url, opts) {
 							agent: request.agent,
 							compress: request.compress,
 							method: request.method,
-							body: request.body
+							body: request.body,
+							signal: request.signal,
 						};
 
 						// HTTP-redirect fetch step 9
@@ -142,7 +171,11 @@ export default function fetch(url, opts) {
 			}
 
 			// prepare response
+			res.once('end', () => {
+				if (signal) signal.removeEventListener('abort', abortAndFinalize);
+			});
 			let body = res.pipe(new PassThrough());
+
 			const response_options = {
 				url: request.url,
 				status: res.statusCode,
@@ -164,7 +197,8 @@ export default function fetch(url, opts) {
 			// 4. no content response (204)
 			// 5. content not modified response (304)
 			if (!request.compress || request.method === 'HEAD' || codings === null || res.statusCode === 204 || res.statusCode === 304) {
-				resolve(new Response(body, response_options));
+				response = new Response(body, response_options);
+				resolve(response);
 				return;
 			}
 
@@ -181,7 +215,8 @@ export default function fetch(url, opts) {
 			// for gzip
 			if (codings == 'gzip' || codings == 'x-gzip') {
 				body = body.pipe(zlib.createGunzip(zlibOptions));
-				resolve(new Response(body, response_options));
+				response = new Response(body, response_options);
+				resolve(response);
 				return;
 			}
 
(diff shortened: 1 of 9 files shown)


bitinn:
Thx a ton for this! `node-fetch` recent development has been slow due to:

- me not having enough free time or business reason to keep evolving `node-fetch` according to the spec.
- v2 was largely driven by Timothy but he is busy with nodejs and other standardization tasks.
- Fetch Spec is constantly being updated, but many of these updates are browser implementer oriented, which few users care.

This leads to why we are slow to accept PRs:

- To keep `node-fetch` spec compliance, one must thoroughly understand Fetch Spec, and few do when their usage is covered (me included, I haven't been keeping up).
- Timothy is the one that really cares about this, with him busy, PR reviews have virtually stopped.

I would like to see `Abort` support happen, I really do, as you can see there are 2 PRs on this topic already (#437, #478), so here are my questions:

- What were changed beyond those 2 PRs?
- How are your PR better than the other two PRs?

If you are committed to learning Fetch Spec and seeing this through, then I am sure many would like to help as `Abort` is one of the top requests.

jnields:
Hey, yeah! Happy to see this through!

The main motivation for this PR is that:

1. I don't believe an implementation of `AbortController` and `AbortSignal` should necessarily be bundled with this package. Instead it makes more sense to simply have this package attach to the API of a passed signal, and keep them separate.

2. The implementation of `AbortError` in those PRs is quite off from how we would expect it to work in the browser. It should more closely match the browser, e.g.,

```
fetch(url, { signal }).catch(err => if (err.name === 'AbortError') { /* aborted fetch */ })
```

Notably I still need to account for `new Request([Request])`, correctly assign signal to the request, correctly close the body if it's a ReadableStream, reject immediately if the `Request`'s signal already has an aborted flag.

This isn't too much, and I'm happy to edit it and make it as spec-compliant as is possible this weekend.

Some things might be missing that are in the other PRs, namely a type check for `signal`. I don't want to require that a global `AbortSignal` be provided so this is probably not possible ... I think this is OK though

bitinn:
@jnields

- I agree that given the API is still in flux, it would be better to not lock ourselves into a specific implementation of `AbortController`.
- I have looked at #437 #478 and I agree that we should return an `AbortError`, we made the same mistake wrapping json parsing error with `FetchError` (#498), we shouldn't fall into that trap again.
- In general, #478 covers a lot of edge cases of abort signal so be sure to pass those tests.
- Good jobs on the doc change. You could follow up with a usage example.
- I am less certain about deprecating `timeout` now, but if you can show an example updating `timeout` to abort `signal`, that would make convincing people easier.
- I think testing with [abort-controller](https://github.com/mysticatea/abort-controller) might be cleaner as we are not relying on the polyfill.

All in all, I look forward to your update!

jnields:
I believe this is to-spec and ready for review.

I don't think there's a clear way to test that the abort event listeners are being removed when using `abort-controller`.

Currently I am not doing any type check for `signal`. I can maybe add some "duck-typing" and check to see that `addEventListener` and `removeEventListener`, `aborted` are all `in` a provided `signal` and that they are of the right type if that seems necessary?

src/index.js
```diff
@@ -42,13 +43,40 @@ export default function fetch(url, opts) {
...
 
 		const send = (options.protocol === 'https:' ? https : http).request;
+		const { signal } = request;
+		let response = null;
+
+		const abort = ()  => {
+			let error = new AbortError('The user aborted a request.');
+			reject(error);
+			if (request.body && request.body instanceof Stream) {
+				request.body.destroy(error);
```
jnields: This is only available in node v8.0.0+

Maybe we can just emit an error event in node 4 & 6? Even though this is slightly off-spec

bitinn:
@jnields

> I don't think there's a clear way to test that the abort event listeners are being removed when using abort-controller.

Could you point to the test you are referring to, since we don't specify an AbortController implementation, I think the tests should work whether one use `abort-controller` or `abortcontroller-polyfill`: we should only rely on AbortController API in tests too.

> Currently I am not doing any type check for signal. I can maybe add some "duck-typing" and check to see that addEventListener and removeEventListener, aborted, and onabort are all in a provided signal and that they are of the right type if that seems necessary?

Perhaps we could just check if `AbortSignal` is the name of the object? I know it's not safe but these 2 packages should agree on this. To be extra sure you can check `addEventListener` and `removeEventListener` exists.

> This is only available in node v8.0.0+, Maybe we can just emit an error event in node 4 & 6? Even though this is slightly off-spec

I would just throw an error when `signal` is provided but node stream doesn't contain `.destroy`. So the user would know their environment is not supported.

Somewhere below this perhaps:

https://github.com/bitinn/node-fetch/blob/master/src/request.js#L160-L167

And I suggest we don't deprecate `timeout` yet, please remove the explicit deprecate warning from code, and change the readme from *deprecated* `timeout` to *encourage* use of `signal`. I will wait until node-fetch v3 to do that as we will drop older node support by then.

jnields:
OK, thank you! I implemented your feedback!

I'm using abortcontroller-polyfill for these tests:
https://github.com/bitinn/node-fetch/pull/539/commits/d926c9917f175c8631a15207da480b052b4971d4

It's not an entirely to-spec implementation, because it has a homemade EventTarget interface. The upside of this is that it exposes all attached events so I don't have to do something like spy on `removeEventListener`

bitinn:
@jnields I think this is largely fine now, I would like you to add one more test: confirm fetch indeed throw error when you try to pass it a valid `signal` on v4 and v6 platform.

I believe there is a fault in your current logic as you depend on `req.body` to be a stream in order to check v8 support, I believe it will fail when you don't pass it a stream body.

You should check that `destroy` is generally available for any readable stream (aka the stream interface support it), not the `req.body` stream.

bitinn:
Other than that, @jimmywarting @TimothyGu @jkantr, if you got time to review this, please do.

FrogTheFrog:
@jnields My only concern so far is that `requestOpts` does not inherit `signal`, thus newly constructed request is not abortable (or is it?). Also, could you port some of my tests (#478) to your PR? I could then close it :)

jnields:
@FrogTheFrog very good catch! You are correct, and I added in your tests to verify that it's now working

@bitinn there should be no reason to need to throw an error if the request body isn't a stream. Users should be able to still cancel requests in node v4 & v6 if the body is a string, FormData, Buffer, etc. I added in a test for cancelling requests with a string body, and it passes just fine on node v4 & v6

bitinn:
@jnields Thx for the quick update, and sorry I was confused by myself on when it should be throwing errors :)

If there are no significant feedback I will merge by the end of the week.

FrogTheFrog:
@bitinn New week, new day, waiting for you to merge this everyday! :D

bitinn:
This is now released as v2.3.0, thx all for pushing this to finish line!

PR #545, opened 2018-11-05
Workaround lack of self in react-native

bitinn:
By trying to guess available global.

- And no, we are not going to fix cases where `global`, `self` or `window` isn't actually global.

- And no, we are not going to check `this` is global, I know v8/jsc shell might need it, but it was not our goal to make `node-fetch` *just work* there.

People, you have to deal with your own mistakes here.

ref: https://github.com/tc39/proposal-global

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Changelog
 
 # 2.x release
 
+## master
+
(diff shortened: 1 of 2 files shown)


NikolaBorislavovHristov:
That looks great

bitinn:
This is now released as v2.3.0, thx all for inputs!

PR #547, opened 2018-11-08
Clone urlsearchparams to avoid mutation

jimmywarting:
...and set correct content-type in response constructor

followup on #520

### DIFF
--- a/src/response.js
+++ b/src/response.js
@@ -23,7 +23,7 @@ export default class Response {
 		Body.call(this, body, opts);
 
 		const status = opts.status || 200;
-    const headers = new Headers(opts.headers)
+		const headers = new Headers(opts.headers)
 
 		if (body != null && !headers.has('Content-Type')) {
 			const contentType = extractContentType(body);
--- a/src/response.js
+++ b/src/response.js
@@ -7,7 +7,7 @@
 
 import { STATUS_CODES } from 'http';
 import Headers from './headers.js';
-import Body, { clone } from './body';
+import Body, { clone, extractContentType } from './body';
 
 const INTERNALS = Symbol('Response internals');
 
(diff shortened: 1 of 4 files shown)


bitinn:
@jimmywarting I think this is now fine, could you fix up the indent for match with current format? Also double check the coverage for `extractContentType` is still 100% now that you modified it slightly.

(use `npm run report`)

jimmywarting:
extractContentType seems fine.

report:
```
----------------|----------|----------|----------|----------|-------------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------|----------|----------|----------|----------|-------------------|
All files       |    95.41 |    92.42 |      100 |    95.35 |                   |
 body.js        |     88.4 |    82.24 |      100 |    88.07 |... 33,463,508,509 |
 fetch-error.js |      100 |      100 |      100 |      100 |                   |
 headers.js     |      100 |      100 |      100 |      100 |                   |
 index.js       |      100 |      100 |      100 |      100 |                   |
 request.js     |      100 |      100 |      100 |      100 |                   |
 response.js    |      100 |      100 |      100 |      100 |                   |
----------------|----------|----------|----------|----------|-------------------|
```

src/response.js
```diff
@@ -23,12 +23,20 @@ export default class Response {
 		Body.call(this, body, opts);
 
 		const status = opts.status || 200;
+    const headers = new Headers(opts.headers)
```
bitinn: you probably missed this line.

bitinn:
Alright I have merged it, but I suspect `res.body` will change in future as it's not spec compliant to leave it as a buffer (same goes for other type of body input I guess, basically #210)

https://developer.mozilla.org/en-US/docs/Web/API/Body/body

jimmywarting:
> EDIT: actually, it might be ok to leave them as a buffer internally, as long as res.body getter is guaranteed to be a stream

That was my idea too. I think I will solve #210 by making two PR, one where i unify everything to a buffer to remove redundant code. and one after that would make it easier to create a stream when all the internal is either a buffer or a stream

bitinn:
@jimmywarting sounds good

PR #558, opened 2018-12-06
Pass localAddress to node request options

Legogris:
This is the `localAddress` part of #256. Happy for suggestions on how to expose this in order to make it testable.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -123,6 +123,7 @@ export default class Request {
 			input.compress : true;
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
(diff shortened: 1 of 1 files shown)


bitinn:
What you are looking for is actually already doable but a little hidden:

- https://github.com/bitinn/node-fetch/blob/master/test/test.js#L1874-L1902
- https://nodejs.org/api/http.html#http_agent_getname_options

bitinn:
I think the nodejs doc will be updated in near future too.

https://github.com/nodejs/node/issues/24098

UrielCh:
It's looks like this change is not present on master branch.

PR #562, opened 2018-12-29
Headers should not accept empty field name

kevinkassimo:
Related: https://github.com/github/fetch/pull/684 , https://github.com/denoland/deno/pull/1427

From https://tools.ietf.org/html/rfc7230#section-3.2,
```
header-field   = field-name ":" OWS field-value OWS

field-name     = token
```
and
```
token = 1*tchar
```

In Chrome, appending a header with empty field name throws and error:
screen shot 2018-12-29 at 12 28 53 am


bitinn:
cheers

PR #563, opened 2019-01-06
Add custom highWaterMark option to clone method

### LINKED ISSUE #386: When users clone(), automatically create streams with custom highWaterMark
Instead of changing [default highWaterMark](https://nodejs.org/api/stream.html#stream_implementing_a_readable_stream) of `res.body` (see #151), perhaps setting them during `clone()` would be a better idea. AFAIK, users mostly run into backpressure when they clone() and consume a large response while leaving the other stream idle.

This will likely solve many trivial issues.

nikdo:
Fixes #386.

I've spent enormous time investigating the conditions when backpressure occurs, yet I still have a lot of unanswered questions. Although I hate uncertainties, the formula I introduced is a verified hypothesis. I'm well prepared and open to any discussions about this topic. :)

I would appreciate any feedback on the code, especially tests.

### DIFF
--- a/CLONE-HIGHWATERMARK.md
+++ b/CLONE-HIGHWATERMARK.md
@@ -1,0 +1,109 @@
+# Increasing clone highWaterMark
+
+## The Problem
+
+When using `res.clone` method it might happen that you want process either the original body or the cloned body first:
+
+```js
+fetch(url)
+  .then(res => cache.put(url, res.clone()))
+  .then(res => res.json())
+  ...
+```
+
+The original response waits for the cloned response to be completed. That means the whole response is buffered in a memory during the first `than` statement. With big response sizes that might lead to consuming too much of the precious server resources.
+
+To keep app allocated memory low, Node.js provides [`highWaterMark` limit of stream internal buffer][hwm]. It defaults to 16kB for all streams but can be overridden explicitly.
+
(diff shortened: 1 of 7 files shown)


bitinn:
Many thx for this PR, I will look into as soon as I am available, it’s a much needed improvement.

If anyone on the node-fetch team want to comment on this please feel free to. It’s not a part of the Fetch spec but browsers definitely have their ways of handling large streams.

(From my phone)

bitinn:
Hey @nikdo, I looked into it and while your documentation is very good. I am thinking about 2 things:

- Personally, I don't see a strong reason to expose `clone(highWaterMark)` option. The reason being, it's not spec-compliant, hence not isomorphic. It seems like people facing this clone problem are using `node-fetch` in an isomorphic sense, which is why they run into `highWaterMark` issue, because they expect nodejs to have the same limit as browsers.

- I would prefer 2 other solutions:

  (a) Can we pass such value via Fetch option, which means at least the browsers can ignore it and not cause a problem?

  (b) Better yet, can we simply provide a helper library that takes a node-fetch response and clone them in a way that allow for large response? And still provide the same API like text() and json()? Basically a wrapper that's isomorphic.

gutenye:
What's the status of this PR? What can I do to help advance it? It's a pretty serious problem considering a lot of users using `node-fetch` in an isomorphic sense. Yet no fix in almost 3 years.

xxczaki:
@gutenye We are currently in process of preparing the `v3` release of `node-fetch` and this issue is in our [roadmap](https://github.com/bitinn/node-fetch/issues/668).

I've just submitted a [pull request](https://github.com/bitinn/node-fetch/pull/671), that fixes this problem, but in a [preferred way](https://github.com/bitinn/node-fetch/pull/563#issuecomment-454669575) :smile:

nikdo:
Let's close this then.

PR #564, opened 2019-01-10
Add code to support TLS client certificates in 1.x branch; bump to v1.7.4

ateamlunchbox:
I need to be able to include TLS info with the request, so added a few lines to copy the info from the init object into the Request when it is created.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -61,6 +61,13 @@ function Request(input, init) {
 	this.port = url_parsed.port;
 	this.path = url_parsed.path;
 	this.auth = url_parsed.auth;
+
+	// TLS certificate info
+	this.ca = init.ca
+	this.cert = init.cert
+	this.checkServerIdentity = init.checkServerIdentity
+	this.key = init.key
+	this.servername = init.servername
 }
 
 Request.prototype = Object.create(Body.prototype);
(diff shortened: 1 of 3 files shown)


ateamlunchbox:
Investigating why tests failed now.

ateamlunchbox:
Apologies as I'm new to node.  Fixing the failing test now.

jimmywarting:
this not something achievable with a custom Agent?

ateamlunchbox:
I am using [apollo-link-http](https://www.npmjs.com/package/apollo-link-http), and that uses node-fetch < 2.0.  I passed my client key/cert and custom CA cert in to apollo-link-http as TLS configuration options as part of the config for fetch, but the TLS info wasn't being copied down into the Request object.  I updated the code so that if TLS configuration information is defined in the initializing object, it will be copied into the new Request being created.

Again, I'm new to node, so if there is a different way of accomplishing my goal without updating this library I would love to entertain any suggestions.

bitinn:
@ateamlunchbox I haven't used `apollo-link-http`, but are you saying somehow when you pass a `agent` to `createHttpLink`, it's not passing all options?

Ref:
- https://github.com/bitinn/node-fetch/pull/564
- https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-http/README.md

AFAIK, you shouldn't have any problem doing that with node-fetch 1.x directly, so it's likely something `apollo-link-http` can change?

PR #572, opened 2019-01-25
fix #571 (OnRedirect Hook) dist the hook for redirects

### LINKED ISSUE #571: Cross-protocol redirect reuses same agent, but for instance http.Agent doesn't support followed https request
The same agent is reused on each followed redirect, but often the redirect crosses the protocol boundaries,  and it can cause an exception, when the agent can't handle the new protocol.

Simple example:

```
let agent = new http.Agent({ keepAlive: true });
let response = await fetch('http://google.com', { agent });
```

> TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"

The reason is - the `node-fetch` follows the redirect to `https`, but still uses `http` agent. [index.js#L151](https://github.com/bitinn/node-fetch/blob/e996bdab73baf996cf2dbf25643c8fe2698c3249/src/index.js#L151)
As a consumer, we don't always know if and where the request will be redirected, so there should be some way to prevent agent forwarding on protocol boundaries, or some hooks should be implemented to give the consumer the ability to set another agent.

tenbits:
Adds the hook on redirects, so that the consumer can change some options or url. 
In particular, the #571 issue can be solved by the developer, by passing the appropriate agent.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -109,7 +109,7 @@ export default function fetch(url, opts) {
 				const location = headers.get('Location');
 
 				// HTTP fetch step 5.3
-				const locationURL = location === null ? null : resolve_url(request.url, location);
+				let locationURL = location === null ? null : resolve_url(request.url, location);
 
 				// HTTP fetch step 5.5
 				switch (request.redirect) {
@@ -144,7 +144,7 @@ export default function fetch(url, opts) {
 
 						// HTTP-redirect fetch step 6 (counter increment)
 						// Create a new Request object.
-						const requestOpts = {
+						let requestOpts = {
 							headers: new Headers(request.headers),
 							follow: request.follow,
 							counter: request.counter + 1,
@@ -153,6 +153,7 @@ export default function fetch(url, opts) {
 							method: request.method,
 							body: request.body,
 							signal: request.signal,
+							onRedirect: request.onRedirect
 						};
 
 						// HTTP-redirect fetch step 9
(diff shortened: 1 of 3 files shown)


jimmywarting:
I believe you are jumping into a deep ocean without knowing if there is any sharks around. 
(not knowing the full history of node-fetch)

- This has been requested a few times before (By building some kind of powerful plugin/event stuff #284. To be able to build HAR, cookie jar, debugging etc)
- Reason for not existing yet is b/c it is not in the spec (we want to follow the spec as much as possible that means we sometimes need to reject some ideas or PR)
- this can already be solved without any modification to fetch by using `{ redirect: 'manual' }` doe it means you have to do more work by yourself
- The closest thing to doing something like this in the web is to look at what service worker are doing.<br>A service worker can act as a MITM and do things like `respondWith` where you can change the request before it gets sent out as well as debugging. doe we also have [fetchObserver](https://github.com/bitinn/node-fetch/issues/427) but that won't allow you to modify the request. This is something i maybe would hope to see in node-fetch one day

tenbits:
Sure, I was just trying the lib, and bamp - the exception. Means, you can't use an `agent` and redirects feature at once. As at any time you could get the redirect to other protocol, which the agent does not support - you'll get the exception. And you can't do nothing about this. Sure, you can do all redirects manually - while the reason again: the redirects with an agent are broken. You can say `Do it yourself` about anything in the library - usually it is the last resort solution. I don't know, what was already suggested, buy I see what's the current state is. `onRedirect` could be the first step to some kind `hook-in` system. I understand you want to stick with the `Fetch API`, but we are talking about the `agent` option, which is already not part of it's consumer api. 

See also how [axios:http:L82](https://github.com/axios/axios/blob/75c8b3f146aaa8a71f7dca0263686fb1799f8f31/lib/adapters/http.js#L82) is handling the issue with different agents - they just have two options. But one agent in options, without any control - is the deadend.

Cheers, Alex

PR #573, opened 2019-01-28
Expose isRedirect method

vinaybedre:
Exposed isRedirect method


bitinn:
This was a legacy api not meant to be used and isn't a part of the spec, we will deprecate it in next release.

You can however write your own helper function as redirect status are defined here:

https://fetch.spec.whatwg.org/#redirect-status

PR #575, opened 2019-02-02
Provide `error.code` for AbortError

szmarczak:


src/abort-error.js
```diff
@@ -13,8 +13,9 @@
 export default function AbortError(message) {
```
szmarczak: Shouldn't this be `class AbortError extends Error` ?

szmarczak: Ah. Old Node support... What do you think about dropping everything below 8?

PR #586, opened 2019-02-23
Allow explicitly-constructed Responses to use custom promises.

### LINKED ISSUE #585: Can't use a custom Promise when explicitly constructing a Response
`node-fetch` claims to provide the means to use a custom promise implementation, and indeed does when you invoke its `fetch` function - however, if you explicitly construct a `Response` (i.e., `new Response`), its `.text()`, `.json()`, etc methods will use the global `Promise` constructor. It'd be really nice if there was a way to have a `Response` with a non-native promise implementation, for tests where you're faking timers and don't want to involve the native event loop at all.

quasicomputational:
Fixes #585.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ const Bluebird = require('bluebird');
 fetch.Promise = Bluebird;
 ```
 
+There's a `Promise` property on `Response` as well, which you should set beforehand if you are constructing one of those explicitly with a non-native Promise library.
+
 ## Common Usage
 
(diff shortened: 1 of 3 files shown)


bitinn:
It's likely when we move onto node-fetch 3.x, custom promise support will be dropped as nodejs' promise implementation is now universal and fast.

Even micro benchmark show the general overhead is minimal, not to mention when you do actual work with promise.

http://bluebirdjs.com/docs/benchmarks.html

PR #589, opened 2019-03-12
Expose Blob

Errorname:
It would be useful to access the Blob polyfill you use.
This would allow us to manually create Blobs to pass to fetch.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,6 +19,7 @@ import Headers, { createHeadersLenient } from './headers';
 import Request, { getNodeRequestOptions } from './request';
 import FetchError from './fetch-error';
 import AbortError from './abort-error';
+import Blob from './blob'
 
 // fix an issue where "PassThrough", "resolve" aren't a named export for node <10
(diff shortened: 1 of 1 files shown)


bitinn:
our Blob implementation is now published as a separate package, we will integration it to node-fetch in future release.

https://github.com/bitinn/fetch-blob

jimmywarting:
What when through your mind when you decided to call it `fetch-blob` and not something like `node-blob`? 😂

Anyway, how would you get a blob then?

```js
fetch.blob = require(blob)

// or 
var impl = require(blob)
var res = await fetch(url) 
await res.blob(impl) 
```

or dose node-fetch rely on fetch-blob?

bitinn:
node-blob already exists, don’t want to use named scope.

Probably want to go with a plug-in / mixin route, as it is a part of Timothy’s old goal. I haven’t decided yet.

(From my phone)

jimmywarting:
ok, only the feature will tell

PR #590, opened 2019-03-19
Improve performance on buffer concatenation

zbinlin:
Buffer.concat with a total length argument will ignore recalculate buffers
length in Buffer.concat, it will improve performance.


jimmywarting:
closed via bee2ad8

zbinlin:
>>> I wonder why we have a getter for type and size instead of just doing this.size and this.type...
This is my coding habit.

PR #596, opened 2019-03-28
fix: Update host if there's a redirection

### LINKED ISSUE #570: Redirect-following error (HTTP -> HTTPS + naked domain -> www subdomain)
Steps to reproduce (tested with Node 8.12.0 and 11.7.0):

```
git clone https://github.com/buu700/simple-reverse-proxy
cd simple-reverse-proxy
npm install
npm start &
curl -H 'Host: cyph.com' http://localhost/wp-content/uploads/2018/10/auditconclusion.txt
```

Expected result: Prints content of https://www.cyph.com/wp-content/uploads/2018/10/auditconclusion.txt

Actual result: `FetchError: maximum redirect reached at: https://www.cyph.com/wp-content/uploads/2018/10/auditconclusion.txt`

Note that it works as expected when changing either the Host header to www.cyph.com or the origin to https://localhost (or both); it's only when redirecting from http://cyph.com -> https://cyph.com -> https://www.cyph.com that the failure occurs.

abruzzihraig:
Closes #570

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -142,8 +142,8 @@ export default function fetch(url, opts) {
 							return;
 						}
 
-            // Update host due to redirection
-            request.headers.set('host', [locationURL.host]);
+						// Update host due to redirection
+						request.headers.set('host', [locationURL.host]);
 
 						// HTTP-redirect fetch step 6 (counter increment)
 						// Create a new Request object.
--- a/src/index.js
+++ b/src/index.js
@@ -142,6 +142,9 @@ export default function fetch(url, opts) {
 							return;
 						}
 
+            // Update host due to redirection
+            request.headers.set('host', [locationURL.host]);
+
 						// HTTP-redirect fetch step 6 (counter increment)
 						// Create a new Request object.
 						const requestOpts = {


keepcosmos:
I have same issue. please merge this.

Richienb:
We probably need to unit test this.

bitinn:
There is one thing I do think could be improved though: our error message is misleading in this case, we might throw a `FetchError` here just for mismatching `Host` and `Location` header.

abruzzihraig:
Sorry guys, I tried to remember what case I'd had when I was dealing with this issue. I just remember it was a very popular domain, and the redirect flow was like http://google.com -> http://www.google.com -> https://www.google.com or something. But I cannot reproduce it anymore even based on fetch@2.3.0.

The fix I've made is not that safe and reliable. It was more like a patch for solving my urgent need at that moment. I didn't pay much time into the whole context to ensure the change is reasonable.

I agree with @bitinn and hope you guys can make a better decision to simply ignore this PR or improve it by other ways. Thanks.

Issue #392, opened 2018-01-20
Expose Blob


modulesio:
Does it make sense to expose `Blob` [here](https://github.com/bitinn/node-fetch/blob/v2.0.0-alpha.9/src/index.js#L200)?

Since `Response` is exported, and `Response` supports `Blob`, not exporting it hinders usability of that.

It seems if someone wanted to they could grab the `Blob` `.constructor` anyway, so hiding it doesn't add any security/encapsulation.

TimothyGu:
Right now I don't want to expose the `Blob` class, as it's very possible that Node.js will add a `Blob`/`File` class as well that can be shared for all modules. I'm only providing the `res.blob()` API to allow for direct usage of the `Blob`'s properties, and to use this `Blob` as the body of another fetch.

modulesio:
Makes sense. If node added Blob I suppose node-fetch would just use that as a drop in.

It's just weird right now since Response effectively does not support constructed Blobs without hacks.

I'll fork for now, but we'll probably end up in the same place. Thanks!

mattlubner:
Not exporting Blob from node-fetch poses a slight problem with testing assertions. We can't do `myBlob instanceof Blob` because we can't access the `Blob` constructor used to create a particular blob.

Could we reconsider exporting the Blob implementation from node-fetch? It would be fine if done in a clunky way, like `export TESTONLY_Blob`, to emphasize that it should not be relied on in production code.

I'm doing `expect(qrBlob.constructor.name).toBe('Blob');` for now, but it's not as strong of an assertion, so not ideal.

jimmywarting:
> @Gozala
>
>I'm using node-fetch with https://github.com/RangerMauve/webrun/ and running into problem because `Blob` class can't be referenced nor there is a way to extract bytes from it. Please consider exposing so it could be used.

Gozala:
@jimmywarting if you use web compatibly libraries they assume `Blob` is available and often attempt to instantiate it, problem with that is they either fail or instantiate different `Blob` implementation which than can't be used with this library because it's unable to recognize it. Essentially this means this library is unable to interop with any other library that uses `Blob`s because implementations are different & incompatible.

Exposing `Blob` would make it possible to use it across libraries and avoid the interop issues.

Gozala:
@modulesio Did you end up forking ? And if so is it maintained / published on npm ?

jimmywarting:
A solution for now:

[...]

PR #601, opened 2019-04-05
Export Blob & BUFFER for dependenants

### LINKED ISSUE #600: Please export Blob class and BUFFER symbol
Hi,

I'm using node-fetch with https://github.com/RangerMauve/webrun/ and running into problem because `Blob` class can't be referenced nor there is a way to extract bytes from it. Please consider exposing so it could be used.

Gozala:
Fix <s>#600</s> #392


bitinn:
our Blob implementation is now published as a separate package, we will integration it to node-fetch in future release.

https://github.com/bitinn/fetch-blob

Issue #199, opened 2016-11-23
Implement body.formData


jimmywarting:
The spec implements a formData method to turn a request/response body back into a FormData

bitinn:
Node.js doesn't have FormData, I just don't think we can do that without explicit dependency on `form-data`, and `form-data` isn't explicitly isomorphic either.

TimothyGu:
When the user is already using node-fetch, then he/she is probably already past the isomorphic layer – as node-fetch itself is not isomorphic. Regarding the dependency on form-data, I don't consider it to be that great of an issue; but we could let the user specify the constructor as an option if we were to care about the size of this installation.

jimmywarting:
Something that just pop-out of my mind is handing over just a instance to wish we can append the data into...  May not be the *right* thing to do...

```js
var FormData = require('form-data')
new Response(...).formData(new FormData)
```

One think i don't like so much about form-data today is that's not fully implemented. Lacks many feature of reading and manipulate the form. Without this it would be no point in even implementing body.formData since you would not have any way of changing/reading it... But hopefully they will resolve this issue: https://github.com/form-data/form-data/issues/124 someday - kind of lost hope since it have taken them a long time now...

I did implement a better fully spec'ed FormData polyfill [here](https://github.com/jimmywarting/FormData). But it was targeted for browser (not for node - since it depends on Blob & File)

octet-stream:
@jimmywarting Hi! If you still need a FormData implementation, may be you check out [this one](https://github.com/octet-stream/form-data)? I've implemented more than just an .append method.

jimmywarting:
Never seen it before, have only used https://www.npmjs.com/package/form-data so far.
Just another implementation of FormData. It's nice that the rest of the methods are implemented like the iterator and all that. But maybe not certain node-fetch can work together with that.

The sad part is that there is no standard spec of how it should be read/handeld. 
But the feature might look brighter: https://github.com/w3c/FileAPI/issues/40
We will maybe get `Symbol.stream` or something like that. There have been numerus of talk about  making FormData streamable and having a unknown content-length

jimmywarting:
btw, i did this for the fetch polyfill: https://github.com/github/fetch/issues/571

The reason i would like to see this implemented on the server is so that i can have one less dependency :P

Express.js app:
```js
app.post('/user', async (req, res) => {
  const fd = new FormData()
  await new Response(req).formData(fd)

  fd.get('username')
})
```

PR #603, opened 2019-04-09
Add formdata-node support

octet-stream:
This pull request brings support of the [formdata-node](https://github.com/octet-stream/form-data) package in request body.

Notes on the PR:
- Differences between form-data and formdata-node: https://github.com/octet-stream/form-data/issues/1
- Content-Length is not supported for formdata-node because the package only have asynchronous method for that: [`FormData#getComputedLength()`](https://github.com/octet-stream/form-data#getcomputedlength---promisenumber)
- formdata-node requires at least [Node 8](https://github.com/octet-stream/form-data/blob/a663fd21f3c13a600186e6c4d742404cc5f46ccd/package.json#L16), but tests seem to work on Node 6 as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,25 @@ fetch('https://httpbin.org/post', options)
     .then(json => console.log(json));
 ```
 
+node-fetch support [formdata-node](https://github.com/octet-stream/form-data) as an alternative:
+
+```js
+import FormData from 'formdata-node';
+
+const form = new FormData();
+
+form.set('greeting', 'Hello, world!');
+
+const options = {
+    method: 'POST',
+    body: form
+};
+
+fetch('https://httpbin.org/post', options)
+    .then(res => res.json())
+    .then(json => console.log(json));
+```
+
 #### Request cancellation with AbortSignal
 
 > NOTE: You may only cancel streamed requests on Node >= v8.0.0
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "codecov": "^3.0.0",
     "cross-env": "^5.1.3",
     "form-data": "^2.3.1",
+    "formdata-node": "^1.5.2",
     "is-builtin-module": "^1.0.0",
     "mocha": "^5.0.0",
     "nyc": "11.9.0",
--- a/src/body.js
+++ b/src/body.js
@@ -5,7 +5,7 @@
  * Body interface provides common methods for Request and Response
  */
 
-import Stream from 'stream';
+import Stream, { Readable } from 'stream';
 
 import Blob, { BUFFER } from './blob.js';
 import FetchError from './fetch-error.js';
@@ -50,6 +50,9 @@ export default function Body(body, {
 		body = Buffer.from(body.buffer, body.byteOffset, body.byteLength);
 	} else if (body instanceof Stream) {
 		// body is stream
+	} else if (body.stream instanceof Readable && typeof body.boundary === 'string') {
+		// body is an instance of formdata-node
+		body = body.stream
 	} else {
 		// none of the above
 		// coerce to string then buffer
@@ -422,6 +425,8 @@ export function extractContentType(body) {
 	} else if (typeof body.getBoundary === 'function') {
 		// detect form data input from form-data module
 		return `multipart/form-data;boundary=${body.getBoundary()}`;
+	} else if (body.stream instanceof Readable && typeof body.boundary === "string") {
+		return `multipart/form-data;boundary=${body.boundary}`;
 	} else if (body instanceof Stream) {
 		// body is stream
 		// can't really do much about this
--- a/src/body.js
+++ b/src/body.js
@@ -5,7 +5,7 @@
  * Body interface provides common methods for Request and Response
  */
 
-import Stream, { Readable } from 'stream';
+import Stream from 'stream';
 
 import Blob, { BUFFER } from './blob.js';
 import FetchError from './fetch-error.js';
@@ -50,7 +50,7 @@ export default function Body(body, {
 		body = Buffer.from(body.buffer, body.byteOffset, body.byteLength);
 	} else if (body instanceof Stream) {
 		// body is stream
-	} else if (body.stream instanceof Readable && typeof body.boundary === 'string') {
+	} else if (body.stream instanceof Stream.Readable && typeof body.boundary === 'string') {
 		// body is an instance of formdata-node
 		body = body.stream
 	} else {
@@ -425,7 +425,7 @@ export function extractContentType(body) {
 	} else if (typeof body.getBoundary === 'function') {
 		// detect form data input from form-data module
 		return `multipart/form-data;boundary=${body.getBoundary()}`;
-	} else if (body.stream instanceof Readable && typeof body.boundary === 'string') {
+	} else if (body.stream instanceof Stream.Readable && typeof body.boundary === 'string') {
 		return `multipart/form-data;boundary=${body.boundary}`;
 	} else if (body instanceof Stream) {
 		// body is stream
--- a/src/body.js
+++ b/src/body.js
@@ -425,7 +425,7 @@ export function extractContentType(body) {
 	} else if (typeof body.getBoundary === 'function') {
 		// detect form data input from form-data module
 		return `multipart/form-data;boundary=${body.getBoundary()}`;
-	} else if (body.stream instanceof Readable && typeof body.boundary === "string") {
+	} else if (body.stream instanceof Readable && typeof body.boundary === 'string') {
 		return `multipart/form-data;boundary=${body.boundary}`;
 	} else if (body instanceof Stream) {
 		// body is stream
--- a/test/test.js
+++ b/test/test.js
@@ -7,6 +7,7 @@ import chaiString from 'chai-string';
 import then from 'promise';
 import resumer from 'resumer';
 import FormData from 'form-data';
+import FormDataNode from "formdata-node";
 import stringToArrayBuffer from 'string-to-arraybuffer';
 import URLSearchParams_Polyfill from 'url-search-params';
 import { URL } from 'whatwg-url';
@@ -1346,6 +1347,28 @@ describe('node-fetch', () => {
 			expect(res.body).to.equal('a=1');
 		});
 	});
+
+  it('should support formdata-node as POST body', function() {
+    const form = new FormDataNode();
+
+    form.set('field', "some text");
+    form.set('file', fs.createReadStream(path.join(__dirname, 'dummy.txt')));
+
+    const url = `${base}multipart`;
+    const opts = {
+      method: 'POST',
+      body: form
+    };
+
+    return fetch(url, opts)
+      .then(res => res.json())
+      .then(res => {
+        expect(res.method).to.equal('POST');
+        expect(res.headers['content-type']).to.equal(`multipart/form-data;boundary=${form.boundary}`);
+        expect(res.body).to.contain('field=');
+        expect(res.body).to.contain('file=');
+      });
+  });
 
 	it('should allow POST request with object body', function() {
 		const url = `${base}inspect`;


jimmywarting:
If we should support any kind of formdata we should begin to implement our own way iterating over all key/values and generate the boundary/body ourself so we can support any custom formdata implementation. 

I was considering making my [FormData](https://github.com/jimmywarting/FormData/) implementation able to run on node too (it's all synchronous). But it uses Blob so much that it's impractical to implement in Node.

However i think that `formdata-node` version looks nicer then `form-data` imo since it has all formdata methods and discarded the hole http request stuff. But i think it has too many other none speced functions that shouldn't be there like: 
- isFormData 
   - you should check this yourself wither with toString/symbol.toStringTag or instance of
- get boundary
- get headers 
- stream
- getComputedLength 
- inspect 
  - you should use nodes [custom inspect symbol](https://nodejs.org/api/util.html#util_util_inspect_custom) instead.
- asyncIterator

Also i can see that formdata-node breaks some test that my browser polyfilled handles correctly.

I'm also a bit concerned about the constructor to, maybe in the feature FormData constructor will accept simular things like the URLSearchParams but it can also include files

I get that they some methods are necessary like some async methods, getting the size. but i think they should be wrapped in a Blob/File like object. that has the size and filename information along with it

Now how would you read the blob without a FileReader you may ask? well, the w3c/FileAPI are implementing new methods to the Blob prototype, namely `blob.arrayBuffer()`, `blob.text()` both of which returns a promise, and also `blob.stream()` chrome are currently implementing it now as a experimental feature

That will be the key features to how we would know how to read the hole content of a custom FormData implementation including files. (files that have not been read into the memory)

So a good formdata implementation would 
 1) take the string path and read the size synchronously
 2) create a File like object with name/size and make a custom arrayBuffer(), text() and stream() implementation

for other stream you would have to create a File like object with a given size too

jimmywarting:
Also can't help but thinking that you have been a bit influenced by my version cuz they looks a bit similar :)

octet-stream:
No, I'm pretty sure I haven't been influenced by your implementation. Originally `formdata-node` was created because `form-data` does not cover FormData API (as I mentioned in first issue of the package). But yeah, I now about your implementation. And I tried it few years ago in one of my project, because Safari was like `form-data` package – with the only `.append()` method available.

> Also i can see that formdata-node breaks some test that my browser polyfilled 
 version handles correctly.

Can you tell me bore of these cases? Maybe I missed something.

> If we should support any kind of formdata we should begin to implement our own way iterating over all key/values and generate the boundary/body ourself so we can support any custom formdata implementation. But form-data currently don't have that functionality 😞

That's may be a good point, but it means that another HTTP client might have to implement that too if they want to support FormData implementations same way. I guess.
And it also means that you have to make yet another partial implementation! :D You know, that part where you read data from `form-data`.

And I don't think is that having those additional methods is something bad, since they're here only to take alternative way to read the data from FormData instance. Like `FormData#_blob` in your implementation, I guess.

bitinn:
Thx for the PR, this is definitely something we should look into on 3.x pre-release. Being modular and allow third-party to provide API implementation is one of our goal.

(To be honest I am a quite busy with my C# project, if @jimmywarting @octet-stream you would like to merge this and help move forward 3.x pre-release, I am happy to help)

Richienb:
Since this PR is intended for version 3.x, I'll switch the target branch accordingly.

octet-stream:
Okay. But this need to be implemented other way.

I almost done with the next major version of my lib which should be more compliant with the spec and will be used in the PR.
But I'm not sure where should I release it since I want to drop 8.x node because of its upcoming [end-of-life](https://github.com/nodejs/release)

Richienb:
@octet-stream That's fine. node-fetch v3.x now requires NodeJS 10.x or later. Just update this PR when you're ready.

octet-stream:
@Richienb Hey, I think we should probably stick with the form-data encoding algorithm as @jimmywarting mentioned [here](https://github.com/octet-stream/form-data/issues/5#issuecomment-486743554) instead of use current solution, since it will just add a support of my package.
formdata-node 2.x is here, so I think it could be implemented now.

Richienb:
@octet-stream There's some linting errors: https://travis-ci.org/bitinn/node-fetch/jobs/630430547

octet-stream:
Sure, I will fix that during work on the feature.

Linter doesn't say anything unless I publish a new commit so that's why these errors happens. I think it would be nice something like husky and lint-staged to inform developers before they publish new pieces of code. Or maybe even prettier.

Richienb:
@octet-stream We'll have to see what @bitinn thinks.

For now, use `yarn lint`.

octet-stream:
Okay, I have added FormDataStream class which supposed to go through form-data instance and push its all fields to Readable stream that can be then used in request.
Not sure, but maybe I need to add some more tests. For now it seem to work fine.

xxczaki:
@octet-stream Is this PR ready for review & eventual merge (after resolving the merge conflicts of course) or you are still working on it?

octet-stream:
The implementation itself is done – fetch can encode FormData-ish objects going through its entries, but I'm not sure how could I bring {Request,Response}#formData() method. Not sure if it necessary. But you can take a look anyway.

tinovyatkin:
@octet-stream Do you mind to rebase over current master? It will be great to see this feature in upcoming 3.x.

octet-stream:
@tinovyatkin Now it matches the master branch. Could you change the target branch then?

Richienb:
Done!

src/request.js
```diff
@@ -217,7 +217,9 @@ export function getNodeRequestOptions(request) {
 	}
 
 	// HTTP-network-or-cache fetch step 2.11
-	if (!headers.has('User-Agent')) {
+	if (headers.get('User-Agent') === 'null') {
+		headers.delete('User-Agent');
+	} else if (!headers.has('User-Agent') && headers.get('User-Agent') !== 'null') {
 		headers.set('User-Agent', 'node-fetch (+https://github.com/node-fetch/node-fetch)');
 	}
```
tinovyatkin: @Richienb Is this commit somehow related to the rest of this PR?
I didn't see the team consensus at #818, so, surprising to see such unrelated change inside commits here...

test/main.js
```diff
@@ -1430,6 +1432,30 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should support formdata-node as POST body', async () => {
+		const form = new FormDataNode();
+
+		const filename = path.join('test/utils/dummy.txt');
```
tinovyatkin, line 1438:
```suggestion
		const filename = path.join('test', 'utils', 'dummy.txt');
```

test/utils/server.js
```diff
@@ -302,6 +302,14 @@ export default class TestServer {
...
 		}
 
+		if (p === '/error/premature') {
+			res.writeHead(200, {'content-length': 50});
+			res.write('foo');
+			setTimeout(() => {
+				res.destroy();
+			}, 100);
+		}
+
```
tinovyatkin, line 312: Is this somehow related to this PR or just a rebase problem?

octet-stream: Yeah, this is not a part of the PR

src/request.js
octet-stream: This probably could be another rebase problem.
https://github.com/node-fetch/node-fetch/blob/bea4a7cb732c54965544bc53324a9bb90fd1004e/src/request.js#L224-L227

src/utils/form-data-stream.js
```diff
@@ -0,0 +1,61 @@
...
+			this.push(Buffer.isBuffer(value) ? value : Buffer.from(String(value)));
+		};
+
+		const onRejected = err => this.emit('error', err);
+
+		this._curr.next().then(onFulfilled).catch(onRejected);
+	}
+}
+
+export default FormDataStream;
```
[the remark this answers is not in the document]
tinovyatkin: Yes, just bump `engines` a little more to Node 10.17 👍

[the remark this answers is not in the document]
octet-stream: I will try this out. Perhaps I also will adopt this for the formdata-node package 👍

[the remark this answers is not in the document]
jimmywarting: fyi, we have been thinking about allowing iterator in the core before and doing "iterator to stream" ourself to allow for both whatwg and node streams (to try and solve https://github.com/node-fetch/node-fetch/issues/647, at least for the Request). (since both are asyncIterable) and yields uint8arrays (i [proposed](https://github.com/whatwg/fetch/issues/809) iterator as body to the spec and it later got closed by an ["iterator to stream"](https://github.com/whatwg/streams/issues/1018) solution instead)

so the form-data encoding could potential just be a iterator as it is (and converted later down the line)

```js
fetch(url, {
  body: asyncIterator
})
```

this would make the node-fetch more "loose" and not so strict.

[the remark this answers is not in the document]
octet-stream: I did similar thing for my FormData implementation: it can take both node and whatwg stream as field's value.
Also, replaced FormDataStream with formDataIterator in the PR. It seem to work fine.

src/utils/boundary.js
```diff
@@ -0,0 +1,13 @@
...
+import {customAlphabet} from 'nanoid';
+
+const alpha = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
+
+/**
+ * @api private
+ */
+const boundary = customAlphabet(alpha, 22);
+
+export default boundary;
```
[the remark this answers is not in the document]
octet-stream: I have replaced `nanoid` with `Math.random().toString(32).slice(2)`.

octet-stream:
I have no idea why does the tests fails on Windows.
Also XO falls to even though I have set minimal version to 10.17

jimmywarting:
@octet-stream updated your comment
it dose not close #199, your PR is about encoding FormData bodies to make a request.
my issue #199 is about decoding a payload back into a FormData :  `res.body.formData()`

octet-stream:
It sure does not. My comment was edited by somebody, according its to history. But I'll think on it, if I can implement it.

jimmywarting:
I edit it 😉

octet-stream:
No, before you did. I never meant to add `body.formData()` in the PR.

jimmywarting:
Started to have some concerns about the calculation of content-length... Math.random isn't always the same length... also think the calculation belongs more to node-fetch than any formdata implementation also...

based on my formdata-polyfill (for browser) i came up with this calculation:

```js
fd = new FormData()
fd.append('a', 'ä')
fd.append('b', new Blob(['aä']), 'skärmdump')
console.log('native size', fd._blob().size)

/**
 * Calculates the content-length of FormData + boundary
 * 
 * @param  {FormData} fd       The formdata to send
 * @param  {string}   boundary the boundary to seperate fields with
 * @return {number}            the total byte size to upload
 */
function calculate(fd, boundary) {
  // cross platform version:
  var calc = txt => new TextEncoder().encode(txt).length
  
  // node version:
  // var calc = txt => Buffer.byteLength(txt)
  
  const boundarySize = calc(boundary)
  let size = 4 + boundarySize

  for (const [name, value] of fd) {
    size += boundarySize + calc(name) + (value instanceof Blob
      ? 78 + calc(value.name) + (calc(value.type) || 24) + value.size
      : 49 + calc(value))
  }

  return size
}

console.log('calculated', calculate(fd, '----formdata-polyfill-'))
```

octet-stream:
> also think the calculation belongs more to node-fetch than any formdata implementation also...

Of course! If you're talking about `FormData#getComputedLength()` method exposed by my library (which was as well mentioned in the PR's description), this method was meant to use at my initial implementations, but since we made our own data form-data encoding algorithm, we can't use it and so I'll add it to the PR.

> Started to have some concerns about the calculation of content-length... Math.random isn't always the same length...

So we will stick you your proposal then? If the length is matter, it should return a string with the same length for each call.

jimmywarting:
Starting to think your formdata-node is more spec compatible and doing things more correct, form-data isn't even iterable. Beginning to think yours is the recommended package for dealing with formdata. and want to change the readme docs

Only thing i wish was for is some sort of `File.from(path)` utility function
```js
fd = new FormData()
fd.append('item', File.fromSync(path))
fd.append('item', await File.from(path))
```

src/utils/form-data.js
```diff
@@ -43,3 +49,32 @@ export default async function * formDataIterator(form, boundary) {
...
+			length += Buffer.byteLength(value);
+		}
+
+		length += carriageLength;
+	}
+
+	// The FormData is empty?
+	if (length === 0) {
+		return length;
+	}
```
jimmywarting, line 82: Even if it is empty it should still append the boundary...
Skärmavbild 2020-06-05 kl  01 15 14

from native chrome...

octet-stream: So what should I return here? A boundary's length?

octet-stream: I took this code from here: https://github.com/octet-stream/form-data/blob/4e5c214fbbdeb3474cc5a3cc19a72e372f1ced1c/src/lib/FormData.mjs#L354-L356

jimmywarting:
> Of course! If you're talking about FormData#getComputedLength() method exposed by my library (which was as well mentioned in the PR's description), this method was meant to use at my initial implementations, but since we made our own data form-data encoding algorithm, we can't use it and so I'll add it to the PR.

👍 

> So we will stick you your proposal then? If the length is matter, it should return a string with the same length for each call.

the length don't matter as long as the "get computed length" takes the boundary size into consideration and uses that same boundary for iterating over

src/utils/form-data.js
jimmywarting: I'm pretty sure https://github.com/node-fetch/node-fetch/pull/603#issuecomment-638985881 calculates the size correct and dose it faster without the need to use some sort of getHeader function. guess it should be `--${boundary}--` size (4 + boundary.size)

octet-stream: > I'm pretty sure #603 (comment) calculates the size correct and dose it faster without the need to use some sort of getHeader function. guess it should be --${boundary}-- size (4 + boundary.size)

I'm just using the same function for formDataIterator and getFormDataLength, so its in the name of DRY principle. Also, these literals looks meaningless when I'm reading your code. And I don't thinл is there some really necessary performance difference to avoid of getHeader.

PR #608, opened 2019-04-16
support reading blob with stream

### LINKED ISSUE #604: Maybe add new read methods to blob.js?
w3c/FileAPI and chrome are making blob.arrayBuffer() blob.text() and blob.stream() happen

https://chromestatus.com/feature/6206980857266176
https://pr-preview.s3.amazonaws.com/w3c/FileAPI/117/795750f...42f8a45.html
[w3c/FileAPI#117](https://bit.ly/2U7Vpfl)

should we implement them too, now? or should we hold off for a while?

jimmywarting:
- added missing `;`
- added missing blob.toString()
- added missing blob.stream() - this returns a node stream.

closes #604

### DIFF
--- a/src/blob.js
+++ b/src/blob.js
@@ -1,6 +1,8 @@
 // Based on https://github.com/tmpvar/jsdom/blob/aa85b2abf07766ff7bf5c1f6daafb3726f2f2db5/lib/jsdom/living/blob.js
 // (MIT licensed)
 
+import { Readable } from 'stream';
+
 export const BUFFER = Symbol('buffer');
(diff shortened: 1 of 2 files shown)


bitinn:
lgtm, but we should track the issue of separating blob / supporting other blob implementation somewhere.

jimmywarting:
> lgtm, but we should track the issue of separating blob / supporting other blob implementation somewhere.

ok, but we can merge this to the current blob.js implementation anyway and then maybe lift it to some other module

PR #610, opened 2019-04-23
Main File Fix

FHGDev:
Set the main file to "src/index" instead of "lib/index", considering the "lib" folder doesn't exist


bitinn:
see: https://github.com/bitinn/node-fetch/blob/master/rollup.config.js

PR #613, opened 2019-04-25
Support making request with any blob that have stream() method

jimmywarting:
This makes node-fetch a bit less dependent on our own Blob implementation.
Now it can construct a Request of any blob like object that has `blob.stream()`

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -37,9 +37,10 @@ export default function Body(body, {
 	} else if (isURLSearchParams(body)) {
 		// body is a URLSearchParams
 		body = Buffer.from(body.toString());
-	} else if (body instanceof Blob) {
+	} else if (isBlob(body)) {
 		// body is blob
-		body = body[BUFFER];
+		this.blobSize = body.size;
+		body = body.stream();
 	} else if (Buffer.isBuffer(body)) {
 		// body is Buffer
 	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
@@ -356,6 +357,24 @@ function isURLSearchParams(obj) {
 }
 
 /**
+ * Check if `obj` is a W3C `Blob` object (which `File` inherits from)
+ * @param  {*} obj
+ * @return {boolean}
+ */
+function isBlob(obj) {
+		return typeof obj === 'object' &&
+				typeof obj.arrayBuffer === 'function' &&
+				typeof obj.slice === 'function' &&
+				typeof obj.text === 'function' &&
+				typeof obj.type === 'string' &&
+				typeof obj.stream === 'function' &&
+				typeof obj.constructor === 'function' &&
+				typeof obj.constructor.name === 'string' &&
+				/^(Blob|File)$/.test(obj.constructor.name) &&
+				/^(Blob|File)$/.test(obj[Symbol.toStringTag])
+}
+
+/**
  * Clone body given Res/Req instance
  *
  * @param   Mixed  instance  Response or Request instance
@@ -460,8 +479,7 @@ export function getTotalBytes(instance) {
 		return null;
 	} else {
 		// body is stream
-		// can't really do much about this
-		return null;
+		return instance.blobSize || null;
 	}
 }
 
--- a/src/body.js
+++ b/src/body.js
@@ -39,8 +39,6 @@ export default function Body(body, {
 		body = Buffer.from(body.toString());
 	} else if (isBlob(body)) {
 		// body is blob
-		this.blobSize = body.size;
-		body = body.stream();
 	} else if (Buffer.isBuffer(body)) {
 		// body is Buffer
 	} else if (Object.prototype.toString.call(body) === '[object ArrayBuffer]') {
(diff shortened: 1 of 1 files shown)


jimmywarting:
was the invalid header thing something that came along with the newer version of node?

bitinn:
yeah I need to fix that issue (on the other hand, if you got time, PR welcomed)

The issue is likely with `test/server.js`, it's blocking another PR too.

https://github.com/bitinn/node-fetch/pull/598#issuecomment-486572612

bitinn:
Regarding this PR: I think we should just check that incoming `Blob` implementation provides API that we will use, not every features of `Blob`

jimmywarting:
```js
  return typeof (obj || {}).stream === 'function' &&
         /^(Blob|File)$/.test(obj[Symbol.toStringTag])
```

Good enough?

bitinn:
@jimmywarting 

- I think `size`, `type`, `constructor.name` checks are needed.
- Should we call `stream` when `arrayBuffer` is available? Whichever we use, check it too.
- The rest seems overkill.

jimmywarting:
> Should we call stream when arrayBuffer is available? 

I think so.

Blob's has a internal [get stream](https://pr-preview.s3.amazonaws.com/w3c/FileAPI/pull/117.html#blob-get-stream)

Both the fetch and FileReader spec says if you want to read/post some blob you should initiate the internal get stream method. even when you use `fr.readAs[Arraybuffer|text|DataURL]` and also even the `blob[text|arrayBuffer|stream]` should initiate the "get internal stream" method

it would also mean you would handle memory more efficient and you can start sending chunks of data already to your request. 

It would be possible to see someone (maybe even Node themself) doing this eventually:
(Very roof sketch with error/missing parts)
```js
const internalMethods = new WeakMap()

class CustomBlob {
  constructor (...args) {
     // something
  }

  get size () {
    return internalMethods.get(this).size
  }

  stream () {
    return internalMethods.get(this).getStream()
  }

  async text () { return consumeStream(this) }
  async arrayBuffer () { return consumeStream(this) }

  // create a new blob but use the same `getStream()` method
  // and just change the start/end and type
  slice (start, end, type = same) {
    const internal = internalMethods.get(this)
    const internal = Object.assign({}, internal.option, {
      start: start,
      end: end,
      size: newSize
    })

    const blob = new Blob([], { type: type })
    internalMethods.set(blob, internal)
    return blob
  }
}

fs.getBlob = function (path, options = {}) {
  const stats = fs.statSync('2gb of pi first numbers.txt')
  const size = stats.size
  const type = getType(path)
  const blob = new Blob([], { type: type })

  // Set the internal get stream method
  internalMethods.set(blob, {
    size: size,
    options: options,
    getStream () {
      return fs.createReadStream(path, options)
    }
  })

  return blob
}
```

This is something i hope @octet-stream would implement in his [formdata-node](https://github.com/octet-stream/form-data) when appending files to FormData

bitinn:
@jimmywarting

> it would also mean you would handle memory more efficient and you can start sending chunks of data already to your request.

- In theory, yes.
- In practice, I am wondering how likely people choose `Blob.stream` over plain `stream`.

Either way, it appears using `Blob.stream` doesn't make our implementation harder, so I have no issue with it.

Will check into test coverage issue when we fixes the node 12 problem.

bitinn:
@jimmywarting  Just wondering, do you plan on adding some test cases and move this forward?

jimmywarting:
There are some existing test for Blobs already, I have just changed the way it reads them...
Any particular test you would like me to write?

bitinn:
@jimmywarting make sense, are there any third-party Blob implementation we would like to test?

bitinn:
Another question:

After merging this PR, Do you think we should publish this `Blob` implementation as separate package so that #589 and #392 is solved?

(Also remind me to update the readme a bit to tell people we now accepts third-party Blob.)

jimmywarting:
> Are there any third-party Blob implementation we would like to test just to be sure?

Not that i know of that supports the stream.

> Do you think we should publish this Blob implementation as separate package

I do think it should be a separate package, But not to happy about making yet another npm-package that dose the same thing with a twist. maybe make a PR to any of this?
- [node-blob](https://www.npmjs.com/package/node-blob)
- [blob.js](https://www.npmjs.com/package/blob.js) 
- [blob-polyfill](https://www.npmjs.com/package/blob-polyfill)
- [node-file-api](https://github.com/node-file-api)

I have helped @eligrey integrated the new read methods on his blob implementation that blob.js and blob-polyfill is based on but they use a older version

bitinn:
@jimmywarting seems like the right candidate, because it's actually just the same implementation without our recent update.

https://github.com/shaneharris/node-blob/blob/master/index.js

bitinn:
Also how does `blob.js` work in node? It seems to take a WHATWG stream?

jimmywarting:
> Also how does blob.js work in node? It seems to take a WHATWG stream?

yep, It tries to use WHATWG streams if it exist.
it also assigns blob to the global scope. in that since it's not modular. with it you get FileReader, Blob And File. 

I have not tested blob.js in node yet.

bitinn:
I don’t think it will work at all (also likely not their design goal)

(From my phone)

bitinn:
rebased and merge in #629

PR #615, opened 2019-04-26
pass custom timeout to subsequent requests

### LINKED ISSUE #612: Pass request timeout on redirect
It seems we forgot to pass `request.timeout` on redirect:

https://github.com/bitinn/node-fetch/blob/master/src/index.js#L147-L156

And for some reasons, we never got around to merge this PR:

https://github.com/bitinn/node-fetch/pull/452

Even though we have AbortSignal support, should still fix it for `timeout` users.

bitinn:
fix #612

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -153,6 +153,7 @@ export default function fetch(url, opts) {
 							method: request.method,
 							body: request.body,
 							signal: request.signal,
+							timeout: request.timeout
 						};
 
 						// HTTP-redirect fetch step 9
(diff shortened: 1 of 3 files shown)


gdoron:
@bitinn Though one can argue it's by design, if I set a timeout of 30 seconds for my request to a.com.
a redirects me to b after 10s, b to c after 10 seconds and so on, it can go forever, while the user asked for 30s timeout.

I think you need to take the delta of the remaining time not the original timeout, no?
@bitinn

bitinn:
@gdoron `timeout` option was an extension to the Fetch spec, before `AbortSignal` was introduced. To precisely control timeout and abort at precise timing, use it instead. (see readme for details).

PR #621, opened 2019-04-28
build: start testing against node.js 12

JustinBeckwith:
This adds Node.js 12.0.0 to the Travis testing matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
   - "6"
   - "8"
   - "10"
+  - "12"
   - "node"
 env:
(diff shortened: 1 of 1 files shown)


bitinn:
Note to self: test matrix is too large today, will merge when we release 3.x and drop node 4.x (likely node 6.x too as end-of-life has past.)

Richienb:
At the moment, the `node` keyword in the `.travis.yml` file depicts NodeJS 12.x so no changes are required until the 22nd of October 2019 when NodeJS 13.x drops.

PR #622, opened 2019-04-28
chore(deps): address deprecated url-search-params package

JustinBeckwith:
While doing an `npm install`, I saw that the `url-search-params` package is deprecated, and needed to be replaced.  See https://www.npmjs.com/package/url-search-params for more details!

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   },
   "homepage": "https://github.com/bitinn/node-fetch",
   "devDependencies": {
+    "@ungap/url-search-params": "^0.1.2",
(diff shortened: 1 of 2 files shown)


bitinn:
Thx, will merge on next release.

jimmywarting:
Wonder if we can't use the [built in](https://nodejs.org/api/url.html#url_class_urlsearchparams)

bitinn:
@jimmywarting this: https://github.com/bitinn/node-fetch/issues/245#issuecomment-375944880

PR #623, opened 2019-04-28
build: disable generation of package-lock since it is not used

JustinBeckwith:
With `package-lock.json` ignored, it's better to just tell `npm` not to generate it at all.  This removes the need for developers on the package to keep deleting `package-lock.json` and re-installing to get the latest patch versions of transitive deps.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


bitinn:
> This removes the need for developers on the package to keep deleting `package-lock.json` and re-installing to get the latest patch versions of transitive deps.

I am not quite sure what you meant here.

JustinBeckwith:
Sure!  Let me explain :)  If you keep a `package-lock.json` in source control, every time someone modifies the dependency tree there *could* be an update to the lock file.  Any change to that file would be included in the commit.

When you ignore that file, it means that this can happen:
- I clone the repo, run `npm install`
- It generates a lock file
- I wait a few weeks
- I `git pull` to update my branch, no lock file updates cause it's ignored
- I run `npm install`

At this stage, I am going to have tons of old packages sitting in my node_modules folder, because they match the last known state of the lock file.  Unless I explicitly delete the lock file, and re-run `npm install`, I will have old patch versions that match the semverity of the request version, but are out of date.  

If `npm install` is instructed to generate no lock file at all - this issue goes away :)

bitinn:
Thx for the explanation, I will add this soon.

PR #625, opened 2019-04-28
test: enable --throw-deprecation for tests

JustinBeckwith:
Enabling `--throw-deprecation` for the tests ensures that no deprecated methods are being used in nodejs.  I'm trying to track down a node.js 12 property deprecation, and figured I may as well do this while I was in here :)


bitinn:
Thx, will merge this on next release.

Issue #588, opened 2019-03-11
HTTP and HTTPS agent


normanrz:
I am trying to use keep-alive agent with node-fetch as described in #423, but I have a problem with HTTPS redirects. Initially, I call a HTTP resource which returns a redirect to a HTTPS resource. In the default `node-fetch` this is not a problem, because the agent is set based on the current request. When overriding the agent there is no way of having per-request agents.

I would propose an API that allows a function instead of a fixed agent:
```javascript
fetch(url, { agent: (protocol) => protocol === 'https' ? httpsAgent : httpAgent })
```

bitinn:
Yes this is a known limit, but since `agent` is non-standard I doubt there will be much work on this, however, if you are lucky enough to be using nodejs 11.7.0+, globalAgent is now replaceable, meaning you can control this precisely:

- https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V11.md#2019-01-17-version-1170-current-bridgear
- https://github.com/nodejs/node/pull/25170

normanrz:
Thanks for pointing to that change in node 11.

bitinn:
this issue is addressed in v2.6.0, `options.agent` now accepts a function, see readme:

https://github.com/bitinn/node-fetch#custom-agent

PR #632, opened 2019-05-01
Issue #588: allow a function supplying the agent to support in follow…

edgraaff:
This PR addresses issue #588. When the user wants to provide an `agent` option, the user has to decide (prior making the call to `fetch`) to provide a http or https agent. If request is for http and makes it redirect to https (or vice versa), the provided `agent` is incompatible and breaks further processing.
Based on the proposal, `agent` can now be a function which takes `parsedURL` as an argument.

A typical use would be:

```js

const options = {
  agent: url => url.protocol === 'https:' ? httpsAgent : httpAgent
};
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ The default values are shown after each option key.
     timeout: 0,         // req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies). Signal is recommended instead.
     compress: true,     // support gzip/deflate content encoding. false to disable
     size: 0,            // maximum response body size in bytes. 0 to disable
-    agent: null         // http(s).Agent instance, allows custom proxy, certificate, dns lookup etc.
+    agent: null         // http(s).Agent instance (or function providing one), allows custom proxy, certificate, dns lookup etc.
 }
 ```
 
--- a/src/request.js
+++ b/src/request.js
@@ -230,7 +230,12 @@ export function getNodeRequestOptions(request) {
 		headers.set('Accept-Encoding', 'gzip,deflate');
 	}
 
-	if (!headers.has('Connection') && !request.agent) {
+	let agent = request.agent;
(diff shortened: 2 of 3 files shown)


bitinn:
This looks good to me, solves issues related to url-based agent creation (not just http/https change during redirect, but more advanced management of connection pools too.)

@jimmywarting what's your take on this?

jimmywarting:
Lgtm

jimmywarting:
I would almost want to get rid of the option to provide a agent and only have one way of solving it...
```diff
- agent: null   // http(s).Agent instance (or function providing one), allows custom proxy, certificate, dns lookup etc.
+ agent: null   // function that can return an Agent that allows for custom proxy, certificate, dns lookup etc.
```

Think it's better to have one way of solving things instead of having to deal with every scenario

bitinn:
> Aside from that I would almost want to get rid of the option to provide a agent and only have a callback function

That's a breaking change and I think largely an unnecessary break (saves only 2 lines).

PR welcomed to fix doc though. I prefer adding another paragraph below the code block.

jimmywarting:
> That's a breaking change and I think largely an unnecessary

Yea, i know and that would be bad. Agree that it's unnecessary. 

> PR welcomed to fix doc though. I prefer adding another paragraph below the code block.

or just an example block using agent? didn't see any in the readme...

---

I'm just a bit frustrated ATM on another project that tries to deal with a lot of unnecessary overhead instead of just accepting one form of input. kinda: is stream? is it a buffer? canvas? string? promise? object? blob? FileList? filePath? url? then do this or that and it magically unwrap itself sometimes sync and sometimes async

Now when i mention it, it kind sounds like our `Body.js` constructor 😅 
but it's not node-fetch i'm lashing out on.

PR #637, opened 2019-05-14
Return null instead of default size for Streams

ghmeier:
[This commit](https://github.com/bitinn/node-fetch/commit/0fc414c2a88e897fd941c06734993a1d9a2747e7) introduced a functional regression when sending
readable streams as the body of a fetch request. If the `size` option is
passed into `fetch` and a `Stream` body is present, the `Content-Type`
header is still set to whatever `size` was even though streams don't
have a consistent method for retrieving the size prior to reading them.

To fix this, we can just revert to the previous implementation of
`getTotalBytes`. That change doesn't appear related to the rest of the
commit.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -48,7 +48,7 @@ import FetchErrorOrig from '../src/fetch-error.js';
 import HeadersOrig, { createHeadersLenient } from '../src/headers.js';
 import RequestOrig from '../src/request.js';
 import ResponseOrig from '../src/response.js';
-import Body from '../src/body.js';
+import Body, { getTotalBytes } from '../src/body.js';
 import Blob from '../src/blob.js';
 import zlib from "zlib";
 
@@ -1323,6 +1323,27 @@ describe('node-fetch', () => {
 		const opts = {
 			method: 'POST',
 			body
+		};
+		return fetch(url, opts).then(res => {
+			return res.json();
+		}).then(res => {
+			expect(res.method).to.equal('POST');
+			expect(res.body).to.equal('a=1');
+			expect(res.headers['transfer-encoding']).to.equal('chunked');
+			expect(res.headers['content-type']).to.be.undefined;
+			expect(res.headers['content-length']).to.be.undefined;
+		});
+	});
+
+	it('should resolve POST requests with readable stream as body when size is passed as a parameter', function() {
+		let body = resumer().queue('a=1').end();
+		body = body.pipe(new stream.PassThrough());
+
+		const url = `${base}inspect`;
+		const opts = {
+			method: 'POST',
+			body,
+			size: 2048
 		};
 		return fetch(url, opts).then(res => {
 			return res.json();
@@ -2631,6 +2652,15 @@ describe('Request', function () {
 		return req.text().then(result => {
 			expect(result).to.equal('a=1');
 		});
+	});
+});
+
+describe('Body', function () {
+	it('should return null for a stream\'s total bytes, ignoring size', () => {
+		const fetchBody = new Body(fs.createReadStream(path.join(__dirname, 'dummy.txt')), {
+			size: 2048
+		});
+		expect(getTotalBytes(fetchBody)).to.equal(null);
 	});
 });
 
--- a/src/body.js
+++ b/src/body.js
@@ -484,7 +484,8 @@ export function getTotalBytes(instance) {
 		return null;
 	} else {
 		// body is stream
-		return instance.size || null;
+		// Since there's no standard way if getting a stream's size before reading it, return nothing.
+		return null;
 	}
 }
 


test/test.js
```diff
@@ -2634,6 +2655,15 @@ describe('Request', function () {
 	});
 });
 
+describe('Body', function () {
+	it('should return null for a stream\'s total bytes, ignoring size', () => {
+		const fetchBody = new Body(fs.createReadStream(path.join(__dirname, 'dummy.txt')), {
+			size: 2048
```
wearhere: What do you think about throwing an error if `size` is set when `Body` is constructed with a stream?

If the library previously just ignored `size` (can't tell), then I guess we might as well continue ignoring it, just wondering if `size` is at all useful here and if ignoring it might mask any other errors.

wearhere: If we did enforce such an error I actually might suggest doing so at the level of `fetch`, i.e. when the `body` and `size` options conflict.

ghmeier: The library previously ignored size so I'd switch back to that as a patch and a more opinionated API design could be implemented later. 

Based on my reading of the spec, it doesn't appear that [`Content-Length` is necessary](https://fetch.spec.whatwg.org/#concept-read-all-bytes-from-readablestream) when reading from a stream. Additionally, the http/https request functions default to `Transfer-Encoding: chunked` which doesn't require a `Content-Length`.

bitinn:
Hmm, this make senses, I guess it's slipped into commit without actually being tested (which is weird as I didn't see a missing coverage).

I will merge this unless @jimmywarting has further comment.

test/test.js
bitinn: For this particular test, if the purpose is to test `getTotalBytes` then I think a new `Body` section isn't needed, you can put it under `node-fetch` section.

jimmywarting:
Haven't looked closely on the pr, issue and what it tries to solve.

A bit busy right now but will come back and review it again later.

But i agree with @ghmeier 
Think it should not be possible to set the size using
```js
fetch(url, {
  size: 123
})
// or 
new Request(url {
  size: 123
})
```

That is not how fetch is spec:ed. you should be more explicit and set content-length header your self

> Based on my reading of the spec, it doesn't appear that Content-Length is necessary when reading from a stream. Additionally, the http/https request functions default to Transfer-Encoding: chunked which doesn't require a Content-Length.

...Right, Isn't that a new addition? don't think there is any way as of as right now to upload something on the web that has a unknown length, it was the new addition of fetch + writable stream that changed it to allow chunked transfer.

```js
const rs = new ReadableStream()
new Request(url, { body: rs })
```

bitinn:
I think using `size` from `options` is simply a bug, and it does not make any sense to me.

ghmeier:
Cool! @bitinn I figured it was something that slipped in there and wouldn't be an obvious test case. Thanks for the quick responses and happy to help 👍

jimmywarting:
How do you feel about instead removing `.size` all together in v3?

jimmywarting:
I have never used `.size`

but when i read the docs now it says

>    size: 0,            // maximum response body size in bytes. 0 to disable

To me that tells me it has something to do with the response size limit. And that it has nothing to do with setting Content-Length when making a request with a stream. It was confusing at first...

ghmeier:
@jimmywarting at first thought I'm on board with removing the `size` parameter altogether although we'd want to evaluate if [this code](https://github.com/ghmeier/node-fetch/blob/master/src/body.js#L249) is useful? From what I can see the `fetch` spec doesn't mention a `size` parameter, so it might be safe to remove.

bitinn:
> To me that tells me it has something to do with the response size limit.

Yes, which why `instance.size` makes no sense

bitinn:
@ghmeier if you don't might, could you remove the tests, and just leave that one line of change. Otherwise I would suggest you just test `getTotalBytes` directly, without fetch at all, it simply shouldn't return a length for stream.

jimmywarting:
Should have been named something like `responseLimit` if anything.

I'm only speculating right now how you would do it in the web browser and if we could maybe do something similar to how you would actually solve it using the spec. (So that the core fetch api don't need to deal with size (responseLimit) at all)

```js
// Util function
function responseLimit(ctrl, size) {
  return response => {
    const length = res.headers.get('content-length')
	
    if (length) {
      if (Number(length) > size) {
        ctrl.abort()
      }
    } else {
      // content-length is unknown, and is a chunked transfer response
      // clone response or tee the response.body.
      // read chunks and then do ctrl.abort if it exceeds
    }

    return response
  }
}


var ctrl = new AbortController()

fetch('https://httpbin.org/bytes/20000', { signal: ctrl.signal })
  .then(limit(ctrl, 3000))
  .then(res => {
    // ...
  })
```

bitinn:
v2.6.0 contains fix to this issue. Thx @ghmeier for reporting.

jimmywarting:
Just remembered. An alternativ solution is to simply use range header

```js
fetch(url, {
  headers: {
    'range': 'bytes=0-1023'
  }
})
```

PR #641, opened 2019-05-20
Added size option to fileType example of Buffer

michailw:
file-type package does not require to read whole file to guess mime type. We can put size option to the request, so example has better quality for readers.


jimmywarting:
also possible to solve it using range request header... just saying

PR #645, opened 2019-05-26
docs: Add more complete stream download example

jdleesmiller:
This PR is based on https://github.com/bitinn/node-fetch/issues/375#issuecomment-495953540 and https://github.com/bitinn/node-fetch/issues/375#issuecomment-496010966, which suggested opening an issue. I thought it might be easier to just open this PR.

The current stream example is correct but leaves out error handling and doesn't wait for the download to complete, which I think most users would probably want to do. There is a new helper function in node 10+, [stream.pipeline](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback) that makes it not too much harder to do both.

I've adjusted the async/await example in https://github.com/bitinn/node-fetch/issues/375#issuecomment-495953540 to fit the style of the other examples.

It's also worth noting that if you run it, the `Octocat.png` image comes back as 404 now, so we might want to update it, but I thought I should keep that out of scope here.


jdleesmiller:
I should also mention that https://github.com/bitinn/node-fetch/pull/463 added the required stream error handling without using `stream.pipeline`, but it seems that it got lost somewhere between then and now, specifically as part of the readme reorganisation in https://github.com/bitinn/node-fetch/pull/504.

Richienb:
This seems to just document a different way to do streaming.

jdleesmiller:
> different way to do streaming

That is perhaps true... but I would argue it is also a more modern, safe and correct way of doing streaming, which is probably closer to what users will want to actually do.

jdleesmiller:
I should also add that this PR was the result of a request from @gr2m in https://github.com/bitinn/node-fetch/issues/375#issuecomment-496010966 --- quite a few people seem confused by the current example. This updated example was an attempt to head off that confusion.

Richienb:
Oh yup.

xxczaki:
Thank you @jdleesmiller!

Issue #643, opened 2019-05-21
Plans to drop babel in v3


bitinn:
I have been thinking about this lately, once we dropped node v4 and v6 support in 3.x release, do we still need `babel`?

My gut feeling is node v8+ are modern enough that all of our code are covered, so there is no value in transpiling (and I want to change back to CommonJS syntax)

https://node.green

If possible I want to drop our build step (`rollup`) altogether.

jimmywarting:
I'm on board with that too.

gr2m:
I guess a build step would still be needed as modules are not the same in Node & web.

I ended up using [`@pika/pack`](https://github.com/pikapkg/builders), which is using rollup under the hood. I see it as a polyfill until we can have the same modules work in both, node & browser.

I wrote about my setup here: https://gist.github.com/gr2m/5dd2ec6e3a04c3dc79f0cbb46decbe46

dmitriz:
@bitin
> My gut feeling is node v8+ are modern enough that all of our code are covered, so there is no value in transpiling (and I want to change back to CommonJS syntax)

The PR https://github.com/bitinn/node-fetch/pull/649 is meant as the first step in this direction. It deals with single exports. 

My question is now, what should be the strategy for multiple exports?
A possible way could be to avoid them altogether, 
to prevent the modules from doing "too much".
Would that be desired or is some other strategy preferred?

TimothyGu:
I’m very much onboard with removing Babel. The time has come for that. I still prefer to keep our source in ES module form though. It’s much easier to transform from ESM to CommonJS style than vice versa, and with the dust settling on Node.js’ support for ESM more people would want to use it, rather than relying on CJS-ES interop.

bitinn:
@TimothyGu I hear you, but the status quo seems to be that many bundling tools are better at transforming from CommonJS to ESM? We have had a lot of problem with people trying to package up node-fetch for AWS using Webpack.

(My guess is most of them are confused about how Webpack/ESM work, but my anecdote observation remains that Webpack seem to just work with CommonJS?)

I love the named export from ESM, for sure, just that it seems to cause some troubles for users at the moment.

TimothyGu:
> many bundling tools are better at transforming from CommonJS to ESM

Well I meant like actually transforming the source, rather than providing interoperability. But if that's the difficulty people are having, then I'm all for it :)

dmitriz:
@TimothyGu 
[...]

PR #640, opened 2019-05-18
[WIP] Update dependencies

gr2m:
Follow up https://github.com/bitinn/node-fetch/issues/626#issuecomment-493311447

The tests are passing, but the build fails. I don’t use babel or rollup myself, any idea what the error could mean?

```
$ npm run build

> node-fetch@2.6.0 build /node-fetch
> cross-env BABEL_ENV=rollup rollup -c


src/index.js → lib/index.js, lib/index.es.js, lib/index.mjs...
[object Object]
    at Object.isRestProperty (/node-fetch/node_modules/@babel/types/lib/validators/generated/index.js:4288:11)
    at DestructuringTransformer.pushObjectPattern (/node-fetch/node_modules/babel-plugin-transform-es2015-destructuring/lib/index.js:197:15)
    at DestructuringTransformer.push (/node-fetch/node_modules/babel-plugin-transform-es2015-destructuring/lib/index.js:108:14)
    at DestructuringTransformer.init (/node-fetch/node_modules/babel-plugin-transform-es2015-destructuring/lib/index.js:317:12)
    at PluginPass.VariableDeclaration (/node-fetch/node_modules/babel-plugin-transform-es2015-destructuring/lib/index.js:468:27)
    at newFn (/node-fetch/node_modules/@babel/traverse/lib/visitors.js:193:21)
    at NodePath._call (/node-fetch/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/node-fetch/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/node-fetch/node_modules/@babel/traverse/lib/path/context.js:88:12)
    at TraversalContext.visitQueue (/node-fetch/node_modules/@babel/traverse/lib/context.js:118:16)

# ... more of the same

(!) Plugin at position 2 plugin: The transformBundle hook used by plugin at position 2 is deprecated. The renderChunk hook should be used instead.
created lib/index.js, lib/index.es.js, lib/index.mjs in 1.2s
```

PR #649, opened 2019-06-07
Changing to CommonJS -- simple cases

dmitriz:
As suggested in #643 
Single export cases are treated.

### DIFF
--- a/src/abort-error.js
+++ b/src/abort-error.js
@@ -10,7 +10,7 @@
  * @param   String      message      Error message for human
  * @return  AbortError
  */
-export default function AbortError(message) {
+function AbortError(message) {
 	Error.call(this, message);
 
 	this.type = 'aborted';
@@ -23,3 +23,5 @@ export default function AbortError(message) {
 AbortError.prototype = Object.create(Error.prototype);
 AbortError.prototype.constructor = AbortError;
 AbortError.prototype.name = 'AbortError';
+
+module.exports = AbortError;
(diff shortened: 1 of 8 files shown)


bitinn:
This is a good direction I might consider, but can't merge at the moment due to various implications.

Also changes will have to be made on v3 branch. Master branch is v2 only for now.

dmitriz:
> This is a good direction I might consider, but can't merge at the moment due to various implications.

Sounds mysterious. I've thought the project was looking for new contributions... did I misunderstand?

> Also changes will have to be made on v3 branch. Master branch is v2 only for now.

Any reasoning behind? The change is not breaking, not even a change in functionality.

bitinn:
> Sounds mysterious.

TL;DR: I can't merge it as the actual export isn't a part of the tests. I can't verify this to work in the wild. (which users can import using named or commonjs).

And if the goal is to remove `babel`, then there is more work to check it actually works and provide a migration plan for cases that doesn't.

I do agree 2.x was doing too much and hence making our work more difficult.

bitinn:
Here is my suggestion to move forward:

- Make the same PR against `v3` branch.
- Try to test different ways to import `node-fetch` if you can.
- After we merged this and #640, we can publish v3 alpha and try to iterate further on this.

(I would love to only deal with commonjs export, but v2 is not the place to do it.)

Richienb:
In the new [roadmap](https://github.com/bitinn/node-fetch/issues/668), we are using replacing Babel and Rollup with Microbundle.

PR #650, opened 2019-06-07
Make browser version a UMD

tmcw:
node-fetch has a browser build, which is great! and it supports 2 of the 3 typical require strategies - browser global and CommonJS.

This PR adds the third - AMD, for loaders like d3-require, so that modules that rely on the UMD build of node-fetch can happily load it.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,23 +1,26 @@
 "use strict";
 
-// ref: https://github.com/tc39/proposal-global
-var getGlobal = function () {
-	// the only reliable means to get the global object is
-	// `Function('return this')()`
-	// However, this causes CSP violations in Chrome apps.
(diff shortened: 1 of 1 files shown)


bitinn:
Thx, but we don't plan to make any further change to our browser export.

In fact, the browser export could very well become deprecated in v3 release.

Do not rely on node-fetch to export browser target, this is not the goal of node-fetch.

In fact, we outlined a few suggested package for isomorphic usage, but of course there are many more on npm if you are willing to search further.

tmcw:
Okay, I'll try to fix the issue downstream with bundle configuration or switching to an explicitly isomorphic package. Thanks for the explanation!

PR #652, opened 2019-06-12
Fix 'can only call Window.fetch on instances of Window' error in Safari.

ricmoo:
In Safari, without binding fetch to the global, I get the error:

```
Unhandled Promise Rejection: TypeError: Can only call Window.fetch on instances of Window
```

Screen Shot 2019-06-12 at 1 04 54 AM

It works fine in Chrome. By binding the function to the global, it works in Safari and Chrome.

Thanks.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -13,7 +13,7 @@ var getGlobal = function () {
 
 var global = getGlobal();
 
-module.exports = exports = global.fetch;
+module.exports = exports = global.fetch.bind(global);
 
 // Needed for TypeScript and Webpack.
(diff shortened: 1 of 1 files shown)


bitinn:
Do not use `node-fetch`'s browser export, it will be deprecated in v3.

Use one these isomorphic wrapper package instead.

ricmoo:
@bitinn Thanks! :)

ricmoo:
Confirmed `cross-fetch` works perfectly. Thanks again for the awesome library. :)

PR #670, opened 2019-09-08
Fix stream piping

### LINKED ISSUE #309: missing response stream error events
The response stream is setup as a series of pipes, but node-fetch only listens for errors on the end of the pipe. Pipes don't send error events along the way. We've seen expected end of JSON input errors which are in fact probably caused by ECONNRESET's which are not surfacing because of this issue.

xxczaki:
Fixes #309 

This PR replaces all `stream.pipe()` with [`pump`](https://www.npmjs.com/package/pump).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "node": ">=8.0.0"
   },
   "scripts": {
-    "build": "microbundle --external http,https,stream,zlib --name fetch --target node --format es,cjs",
+    "build": "microbundle --external http,https,stream,zlib,pump --name fetch --target node --format es,cjs",
     "prepare": "npm run build",
     "test": "cross-env BABEL_ENV=test mocha --require @babel/register --throw-deprecation test/test.js",
     "report": "cross-env BABEL_ENV=coverage nyc --reporter lcov --reporter text mocha -R spec test/test.js",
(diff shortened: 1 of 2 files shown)


bitinn:
lgtm, would be perfect if we can come up with a test case.

bitinn:
actually I think we need more than just using `pump`, without adding a callback we aren't really rejecting upstream errors. (haven't used `pump` myself, need a test to confirm this is working at expected.)

xxczaki:
I don't really know how to test this, looking at [this](https://github.com/bitinn/node-fetch/issues/309#issuecomment-314293531) comment we might not even be able to :cry:

xxczaki:
@Richienb I think we should depend on it.

xxczaki:
@bitinn I think we are safe to merge this, after the changes I made in fb00b38.

xxczaki:
I accidentally published a branch called `pika-pack` (where I was testing usage of [@pika/pack](https://github.com/pikapkg/pack)) and immediately deleted it. 

Unfortunately because of that Travis build is broken and I'm not sure how to fix that :cry:

Richienb:
@xxczaki The best way to fix the Travis CI build is to trigger a build on a known good branch to restore the status.

xxczaki:
I think it's time to merge this :smile:

PR #673, opened 2019-09-13
fix: Better object checks

### LINKED ISSUE #667: Do not use constructor.name to check object
`isAbortSignal` and `isBlob` use `constructor.name` only to check.

https://github.com/bitinn/node-fetch/blob/bf8b4e8db350ec76dbb9236620f774fcc21b8c12/src/request.js#L36-L43

https://github.com/bitinn/node-fetch/blob/95286f52bb866283bc69521a04efe1de37b26a33/src/body.js#L364-L378

However, `constructor.name`  can be rewrited. So, for example, webpack4 with production mode will break `constructor.name` of AbortSignal that will be rewritten to `"s"`.

image

So, `Expected signal to be an instanceof AbortSignal` Error was occured (ref: https://github.com/yumetodo/vscode-google-photos-uploader/issues/11 )

`isURLSearchParams` is adopting good implementation (ref: https://github.com/bitinn/node-fetch/issues/296#issuecomment-307598143, https://github.com/bitinn/node-fetch/pull/297/commits/302992d68e75748a7781c52e868b4fde7925325f ).  `Object.prototype.toString()` will call `Symbol.toStringTag` propaty which is not writable(see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag))

https://github.com/bitinn/node-fetch/blob/95286f52bb866283bc69521a04efe1de37b26a33/src/body.js#L346-L362

Of course, I know that there are two workarounds for webpack4:

1. Set `optimization.minimize` to `false` (ref: https://github.com/webpack/webpack/issues/6841#issuecomment-376129163)
2. Specify minimizer manually and set `keep_fnames` to `true` (ref: https://github.com/CartoDB/carto-vl/pull/1066)

However, rely on `constructor.name` is a bad idea, I think.

Richienb:
Closes #667.

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -14,8 +14,8 @@ const NAME = Symbol.toStringTag;
  * @return {boolean}
  */
 export function isURLSearchParams(obj) {
-    return (
-        typeof obj === 'object' &&
+	return (
+		typeof obj === 'object' &&
         typeof obj.append === 'function' &&
         typeof obj.delete === 'function' &&
         typeof obj.get === 'function' &&
@@ -24,7 +24,7 @@ export function isURLSearchParams(obj) {
         typeof obj.set === 'function' &&
         typeof obj.sort === 'function' &&
         obj[NAME] === 'URLSearchParams'
-    );
+	);
 }
 
 /**
@@ -34,14 +34,14 @@ export function isURLSearchParams(obj) {
  * @return {boolean}
  */
 export function isBlob(obj) {
-    return (
-        typeof obj === 'object' &&
+	return (
+		typeof obj === 'object' &&
         typeof obj.arrayBuffer === 'function' &&
         typeof obj.type === 'string' &&
         typeof obj.stream === 'function' &&
         typeof obj.constructor === 'function' &&
         /^(Blob|File)$/.test(obj[NAME])
-    );
+	);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Richienb:
@bitinn In that case, do we need `isBlob` to continue using duck-type checks?
`URLSearchParams` is not built into node v8 so we can either convert this to duck-type checks also or include the polyfill.

src/utils/is.js
```diff
@@ -0,0 +1,61 @@
...
+export function isBlob(obj) {
+	return (
+		typeof obj === 'object' &&
+        typeof obj.arrayBuffer === 'function' &&
+        typeof obj.type === 'string' &&
+        typeof obj.stream === 'function' &&
+        typeof obj.constructor === 'function' &&
+        typeof obj.constructor.name === 'string' &&
+        /^(Blob|File)$/.test(obj[NAME])
+	);
```
jimmywarting: > drop constructor.name in favor of toStringTag in v3 release where applicable.

can we use symbol here instead?

Richienb: It turns out the check after `constructor.name === 'string'` would indirectly also test for a string so I've removed the extraneous check.

Richienb:
We've got enough reviews - time to merge this.

PR #674, opened 2019-09-16
check for global.fetch before binding it

bbosiljcic:
We ran into some issues where this was executed in browsers that didn't support fetch (e.g. IE11)


bitinn:
AFAIK we will be dropping `browser` export in v3 release, so this is a band-aid patch for v2 only, if there is going to be another release of v2.

The correct solution would be to use some of these suggested modules that export correctly given contexts.

Otherwise LGTM

Issue #370, opened 2017-11-22
Logging and/or FetchEvent


mfulton26:
I'd like to be able to log every request and response to the console (and/or possibly to a file).

Is FetchEvent supported in any way so that I can hook into node-fetch?

jimmywarting:
I tried to mimic FetchEvent as much as possible for nodejs
with this package: https://www.npmjs.com/package/fetch-event
Stop maintain it...

Tried to build CacheStorage that exist in Service Workers also:
https://www.npmjs.com/package/fetch-event

jimmywarting:
I would like to bring this up again.

The [FetchEvent](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent) is something standardized and follows a specification, not only can you modify the request and the response, you can also [respondWith](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith) with something from a cache without every making a request. It can be helpful for mocking responses, logging each request (#787) before it's execute and hooking up a cookie store (jar)

Then we also got the [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) that could also potentially be useful to others.

But before implementing it i guess we need some developer interest. 

I would have preferred FetchEvent over this PR #677 & #480

jimmywarting:
some keypoints:

- Can we create a little min service worker with http.createServer that can dispatch fetch events?
  - can you then use `event.respondWidth` to send back a response to the client?
- Just using fetch to make request:
  - how would the request dispatch an event back to the developer?
  - what if two modules uses node-fetch to make request, dose both modules listen to all requests that is going on or should you instantiate some new class or something?

I created this [fetch-server](https://github.com/jimmywarting/fetch-server) some long time ago in a attempt to create some sort of service worker in node. looking back at it now feels a bit... old
some things shouldn't even be in there either...
It's also able to handle multipart/form-data request.
But the hole project got a bit abandoned
node-fetch had some bugs i wanted to get fixed before i could proceed with it.

then there is also the https://www.npmjs.com/package/fetch-cachestorage and https://www.npmjs.com/package/fetch-event which is even more outdated than my fetch-server.

just to borrow some ideas.

PR #677, opened 2019-09-21
feat: Add basic middleware functionality

Richienb:
This PR adds basic middleware functionality. I'm not sure how to test this.


bitinn:
Tests are needed to demonstrate why this feature is needed. I need to know the use case for this to be honest.

Richienb:
> Tests are needed to demonstrate why this feature is needed. I need to know the use case for this to be honest.

@bitinn There's a use case explained here: https://github.com/bitinn/node-fetch/issues/668#issuecomment-533532776

jimmywarting:
Before doing something like this i would want to start on crating a constructor to get a new fetch instance that don't conflict with others `new FetchInstance()`

gdoron:
Also useful for using automatic follow redirect but in the end get all the redirect that there were in the middle.
That's the use-case for me.

JefferyHus:
This is more of `hooks` and not a `middle-ware` design. To be honest I don't see a reason why would someone need such things when it comes to sending `requests`, then comes the `after` hook which is imo is useless, `fetch` returns a promise so you already can execute things after fulfilling your request.

LinusU:
I agree with @JefferyHus, and would like to add that I think that global hooks like this are quite messy. If someone uses this to e.g. parse errors returned from an api, it can easy interfere with any other responses being returned from other parts of the code...

Also, should we really add features that aren't present in the official `fetch` specification? 🤔

xxczaki:
@li

xxczaki:
> Also, should we really add features that aren't present in the official fetch specification? 🤔

@LinusU The main reason why we want to introduce a middleware functionality/hooks is to allow users to set up things like caching/cookies, which are in the spec, but can't be added to node-fetch in an unopinionated way.

Richienb:
We can allow the consumer to specify separate hooks as an option like from [got](https://www.npmjs.com/package/got#hooks).

LinusU:
@xxczaki thanks for the clarification, I still think that it shouldn't be global though. I think that `got`s approach is quite good!

xxczaki:
@LinusU Yeah, I also really like `got` hooks and I think we can take some inspiration from them.

lastthyme (approved):

jimmywarting:
This PR have been idle for to long (created 2019) and never merged and perhaps outdated as well?. with some consideration i think FetchEvent #370 is better suited - so closing this...

PR #678, opened 2019-09-23
refactor: Add fetch-blob

Richienb:
> Depends on https://github.com/bitinn/fetch-blob/pull/3.


bitinn:
I have no idea what the original code's "prevent copying" comment mean, hopefully it's a non-issue?

Otherwise LGTM, if all tests are green.

PR #682, opened 2019-10-07
fix: Properly parse meta tag when parameters are reversed

dsuket:
`convertBody`function work fine when html has this meta tag.
`<meta http-equiv="content-type" content="text/html;charset=shift_jis">`

However, if the `http-equiv` attribute is reversed like this, it will not work.
`<meta content="text/html; charset=Shift_JIS" http-equiv="Content-Type" />`

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -307,7 +307,7 @@ function convertBody(buffer, headers) {
 	if (!res && str) {
 		res = /<meta[\s]+?http-equiv=(['"])content-type\1[\s]+?content=(['"])(.+?)\2/i.exec(str);
 		if (!res) {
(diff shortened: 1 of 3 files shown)


Richienb:
In v3, we are going to be replacing the regex system with an actual parser in [Richienb/fetch-charset-detection/src/index.ts#L53](https://github.com/Richienb/fetch-charset-detection/blob/32e2216aad68c3c482913130ac5bb3e8f5ce86f1/src/index.ts#L53) but we can still use this as a band-aid patch if we do eventually make a final release for v2.

I'm adding some unit tests now.

dsuket:
@Richienb  Thank you for your reply.
Added test. I'm sorry if it's misplaced.

PR #684, opened 2019-10-09
Create requestOpts only if it's needed

gdoron:
Regarding the comments of "step 6\9", not sure where it's documented, so I decided not to touch them, let me know if you want me to change something.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -61,7 +61,7 @@ export default function fetch(url, opts) {
 		const { signal } = request;
 		let response = null;
 
-		const abort = ()  => {
+		const abort = () => {
 			let error = new AbortError('The user aborted a request.');
 			reject(error);
 			if (request.body && request.body instanceof Stream.Readable) {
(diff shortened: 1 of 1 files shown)


gdoron:
@xxczaki 
Thank you for devoting your time to review.

How can a change that has absolutely no side effects besides performance and memory improvements (albeit small) be not spec-compliant?

xxczaki:
@gdoron Sorry for the confusion, what I was trying to say was that this PR changes the order of steps of an HTTP-redirect fetch defined in [the specification](https://fetch.spec.whatwg.org/#http-redirect-fetch) (step 9 would occur before step 6). 

So in order to merge this we would need to explain why we are doing so, preferably using comments. 

I'm aware of performance & memory usage improvements and because of that (contrary to what I said earlier) I would like to merge this :smile:

gdoron:
@xxczaki is there something you want me to add\change to this MR?

PR #688, opened 2019-10-30
v3 docs & @pika/pack

xxczaki:
## Docs

- **README.md**
	- [x] Format example code to use `xo` style
	- [x] Update `1.x` and `2.x` references
	- [x] Check Node.js references, so that they are up-to-date with Node.js >=10
	- [ ] Document new features
	  - [x] `highWaterMark` option
	  - [x] Bundled TypeScript types
	  - [ ] UTF8-encoded urls (wait until we replace `url.parse()` with WHATWG `new URL()`)
	  - [ ] Data URI support
	- [x] Update acknowledgements

- **UPGRADE-GUIDE.md**
	- [x] Format example code to use `xo` style
	- [x] Add `3.x` upgrade guide

- **LIMITS.md**
	- [x] Update according to `3.x` changes

- **ERROR-HANDLING.md**
	- [x] Format example code to use `xo` style
	- [x] Update according to `3.x` changes (if needed)

- **CHANGELOG.md**
	- [ ] Update according to `3.x` changes (wait until v3 roadmap is fully finished)


bitinn:
Lovely so far, could also test run these examples on runkit to make sure they are valid.

https://runkit.com/npm/node-fetch

bitinn:
Also don't forget about the backer and discord badge.

Write a sentence or 2 to convince people to support this project also won't hurt :)

README.md
```diff
@@ -431,6 +464,20 @@ const options = {
 };
 ```
 
+<a id="custom-highWaterMark"></a>
+
+#### Custom highWaterMark
+
+Stream on Node.js have a smaller internal buffer size (16Kb, aka `highWaterMark`) from client-side browsers (>1Mb, not consistent across browsers). Because of that, when you are writing an isomorphic app and using `res.clone()`, it will hang with large response in Node. 
+
+Since `3.x` you are able to modify the `highWaterMark` option:
```
bitinn: My advise is don't suggest people use `highWaterMark` as the primary solution, and instead tell them the correct way is to resolve both original response and cloned response using `Promise.All` (it's isomorphic and you no longer need to worry about `highWaterMark`), preferably with an example.

Richienb:
Turns out, Runkit [supports custom code](https://runkit.com/docs/resources#pages) to show on our [package page](https://npm.runkit.com/node-fetch).

For now, I'll set it to the recommended importing code from the README but we could also add an example code file.

xxczaki:
@Richienb That's a cool feature! I think it would be the best to add a file with basic, but most common examples, like so:

```js

// Plain text or HTML
fetch('https://github.com/')
    .then(res => res.text())
    .then(body => console.log(body));

// JSON
fetch('https://api.github.com/users/github')
    .then(res => res.json())
    .then(json => console.log(json));

// Simple Post
fetch('https://httpbin.org/post', {method: 'POST', body: 'a=1'})
    .then(res => res.json())
    .then(json => console.log(json));

// Post with JSON
const body = {a: 1};
 
fetch('https://httpbin.org/post', {
        method: 'post',
        body:    JSON.stringify(body),
        headers: {'Content-Type': 'application/json'}
    })
    .then(res => res.json())
    .then(json => console.log(json));
```

xxczaki:
I created the upgrade guide for v3.x

bitinn:
@xxczaki I have read the upgrade guide, suggest adding the following:

- We have also dropped the `browser` export in package.json, and ask people to use package like `cross-fetch`, this is a breaking change to note.
- We have also dropped `res.textConverted`, and ask people to use `fetch-charset-detection`, this is also a breaking change.
- Blob implementation is now `fetch-blob` and hence is exposed, unlikely previously, where Blob type is only internal and not exported.
- We now handles UTF-8 URL properly due to WHATWG-compliant URL support in nodejs.

Also, put the breaking changes in its own section, and enhancements in another.

Also, probably better to write another doc, instead of overwriting the existing v1->v2 upgrade doc, it's worth preserving it for future reference :)

xxczaki:
@bitinn Done :smile: I also added an "Upgrading" section to readme.

xxczaki:
I moved the *.md files to the `docs` folder (except from README.md of cource). I also created the `v3-LIMITS.md` file with updated differences. Let me know if I missed anything :smile:

cc @bitinn @Richienb

Richienb:
All reviewers have now approved the PR. Time to merge!

PR #689, opened 2019-11-02
Replace microbundle with @pika/pack

xxczaki:
There are a couple problems with [microbundle](https://github.com/developit/microbundle):
- it has 70 issues open
- it doesn't work well when named and default exports are used together

A while ago, @gr2m suggested using [@pika/pack](https://github.com/pikapkg/pack). Just like microbundle, it handles bundling without (almost) any configuration. It also includes some really cool plugins, for example [@pika/plugin-build-types](https://github.com/pikapkg/builders/tree/master/packages/plugin-build-types/), which automatically generates TypeScript types from JavaScript (though we are not using it, as we already have type definitions :smile:). @pika/pack fixes the issue with exports as well.


gr2m:
I'm very happy with Pika. Especially it's https://github.com/apps/pika-ci app (npm-installable and web-importable builds for _every_ pull request!) and CDN integration is amazing.

Richienb:
@xxczaki Since this is **node**-fetch, we don't actually need `@pika/plugin-build-web`. I'll take that out.

xxczaki:
Ok, I think we are ready to merge this :smile:

PR #694, opened 2019-11-10
refactor: Replace `encoding` with `fetch-charset-detection`.

### LINKED ISSUE #554: charset detection not working
The URL https://www.aksam.com.tr/guncel/baskan-erdogan-48inci-muhtarlar-toplantisinda-konusuyor/haber-795519 has:
```
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-9" />
<meta http-equiv="Content-Type" content="text/html; charset=windows-1254" />
```
However `res.textConverted()` returns 
```<p>Cumhurba�kan� Recep Tayyip Erdo�an, Be�tepe'de 48'inci Muhtarlar Toplant�s�'nda konu�tu.</p>    <p><strong><em>Erdo�an'�n a��klamalar�ndan sat�r ba�lar� ��yle:</em></strong></p>```

I see that in body.js `convertBody` the charset is supposed to be detected.  I think the reason is that `Content-Type` is uppercase, but the regex in body.js only matches lowercase.  The preview `str` should be lowercased.

### PR DESCRIPTION
**Pull request by Richienb**
This PR replaces the `encoding` optional dependency with `fetch-charset-detection`.

Fixes #554.

cc @bitinn @xxczaki

### DIFF
--- a/docs/v3-UPGRADE-GUIDE.md
+++ b/docs/v3-UPGRADE-GUIDE.md
@@ -37,7 +37,7 @@ If you want charset encoding detection, please use the [fetch-charset-detection]
 
 ```js
 const fetch = require("node-fetch");
-const {convertBody} = require("fetch-charset-detection");
+const convertBody = require("fetch-charset-detection");
 
 fetch("https://somewebsite.com").then(res => {
 	const text = convertBody(res.buffer(), res.headers);
--- a/docs/v3-UPGRADE-GUIDE.md
+++ b/docs/v3-UPGRADE-GUIDE.md
@@ -33,7 +33,16 @@ Prior to v3.x, we included a `browser` field in the package.json file. Since nod
 
 ## Dropped the `res.textConverted()` function
 
-If you want charset encoding detection, please use [fetch-charset-detection] package.
+If you want charset encoding detection, please use the [fetch-charset-detection] package ([documentation][fetch-charset-detection-docs]).
+
+```js
+const fetch = require("node-fetch");
+const {convertBody} = require("fetch-charset-detection");
+
+fetch("https://somewebsite.com").then(res => {
+	const text = convertBody(res.buffer(), res.headers);
+});
+```
 
 # Enhancements
 
@@ -66,6 +75,7 @@ Since v3.x you no longer need to install `@types/node-fetch` package in order to
 [LTS plan]: https://github.com/nodejs/LTS#lts-plan
 [cross-fetch]: https://github.com/lquixada/cross-fetch
 [fetch-charset-detection]: https://github.com/Richienb/fetch-charset-detection
+[fetch-charset-detection-docs]: https://richienb.github.io/fetch-charset-detection/globals.html#convertbody
 [fetch-blob]: https://github.com/bitinn/fetch-blob#readme
 [whatwg-nodejs-url]: https://nodejs.org/api/url.html#url_the_whatwg_url_api
 [changelog]: CHANGELOG.md
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
 		"codecov": "^3.6.1",
 		"cross-env": "^6.0.3",
 		"form-data": "^2.5.1",
+		"iconv-lite": "^0.5.0",
 		"mocha": "^6.2.2",
 		"nyc": "^14.1.1",
 		"parted": "^0.1.1",
@@ -67,7 +68,7 @@
 		"utf8": "^3.0.0"
 	},
 	"optionalDependencies": {
-		"encoding": "^0.1.12"
+		"fetch-charset-detection": "^0.2.0"
 	},
 	"@pika/pack": {
 		"pipeline": [
--- a/package.json
+++ b/package.json
@@ -1,160 +1,160 @@
 {
-  "name": "node-fetch",
-  "version": "2.6.0",
-  "description": "A light-weight module that brings window.fetch to node.js",
-  "main": "dist/index.js",
-  "module": "dist/index.mjs",
-  "types": "types/index.d.ts",
-  "files": [
-    "src/**/*",
-    "dist/**/*",
-    "types/**/*.d.ts"
-  ],
-  "engines": {
-    "node": ">=10.0.0"
-  },
-  "scripts": {
-    "build": "pika-pack --out dist/",
-    "prepare": "npm run build",
-    "test": "cross-env BABEL_ENV=test mocha --require @babel/register --throw-deprecation test/test.js",
-    "report": "cross-env BABEL_ENV=coverage nyc --reporter lcov --reporter text mocha -R spec test/test.js",
-    "coverage": "cross-env BABEL_ENV=coverage nyc --reporter json --reporter text mocha -R spec test/test.js && codecov -f coverage/coverage-final.json",
-    "lint": "xo"
-  },
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/bitinn/node-fetch.git"
-  },
-  "keywords": [
-    "fetch",
-    "http",
-    "promise"
-  ],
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by bitinn** [comment]
Also the diff view on GitHub is weird, I wouldn't expect large change on `package.json`, also `writeToStream` is somehow removed?

**Comment by bitinn** [comment]
Also note the charset detection is not a part of the Fetch Spec, which is why we have been trying to deprecate it. In particular, I introduced them in v1 and I regret it ever since.

**Comment by bitinn** [comment]
There is no reason is hook charset detection into our code neither, by doing encoding conversion you essentially need to cache the whole response in memory. Get them through `res.body` is basically the same.

Also this is why I asked about documentation, we are introducing a breaking API changes, people who use `node-fetch` as web crawler will need to use the new package, so documentation is needed to guide them on the upgrade.

(I am writing these comments for future reader's reference, because obviously people will hate breaking changes)

**Comment by Richienb** [comment]
> Also this is why I asked about documentation, we are introducing a breaking API changes, people who use `node-fetch` as web crawler will need to use the new package, so documentation is needed to guide them on the upgrade.

Charset encoding detection is already described in the [v3 migration guide](https://github.com/bitinn/node-fetch/pull/688/files#diff-a253ce11d719b97799fdf32b898a1a87R36) and the API is documented in the [documentation](https://richienb.github.io/fetch-charset-detection).

**Comment by Richienb** [comment]
@bitinn Just for clarification, which functions were we supposed to migrate to `fetch-charset-detection`?

**Comment by bitinn** [comment]
Public API-wise it's really just `textConverted`, I do not believe we expose anything else. But we should remove code that's added only in support for `textConverted`, like `convertBody`.

I am not aware of any other dependency on `encoding`.

**Comment by bitinn** [comment]
Given v2 code I would say `fetch-charset-detection` could expect a `buffer` input.

https://github.com/bitinn/node-fetch/blob/master/src/body.js#L151

**Comment by bitinn** [comment]
Now that I have looked at `fetch-charset-detection`, I think it's doing a lot more than my original expectation. Sorry to not clarify early. I would have keep `fetch-charset-detection` minimal (and it's should be because it doesn't depends on anything else, just a `buffer` input, and a `content-type` header)

**Comment by bitinn** [comment]
In short, I would just ask the users to pass the `res.buffer` and `content-type` header. `fetch-charset-detection` need not to know `node-fetch`, to be honest. It can be used for any other response that's compatible.

**Comment by Richienb** [comment]
@bitinn It's ready for another re-review.

File: package.json [hunk 1]
```diff
@@ -67,7 +68,7 @@
 		"utf8": "^3.0.0"
 	},
 	"optionalDependencies": {
-		"encoding": "^0.1.12"
+		"fetch-charset-detection": "^0.2.0"
```
bitinn: I don't know if I can be more clear, but no, we don't want dependency on fetch-charset-detection at all, not even optional

src/body.js
```diff
@@ -11,9 +11,9 @@ import Blob from 'fetch-blob';
 import FetchError from './errors/fetch-error';
 import {isBlob, isURLSearchParams, isArrayBuffer, isAbortError} from './utils/is';
 
-let convert;
+let convertBody;
 try {
-	convert = require('encoding').convert;
+	convertBody = require('fetch-charset-detection').convertBody;
```
bitinn: which means, we are removing these code, not replacing them with the new code

src/body.js
```diff
@@ -144,6 +144,10 @@ Body.prototype = {
 	 * @return  Promise
 	 */
 	textConverted() {
+		if (!convertBody) {
```
bitinn: which means, we no longer has this API, we dropped it.

the users will have to rewrite their code from `res.textConverted()` to something like

```
var convertBody = require('fetch-charset-detection');

var buffer = res.buffer();
var length = res.headers.get('content-length');
var text = convertBody(buffer, length);
```

Richienb:
Oh. I am very stupid. I'll be back for another review.

bitinn:
@Richienb it's my bad, I should have been more clear on this. I thought it should be clear while the situation is more nuanced than my expectation.

src/body.js
```diff
@@ -433,34 +343,28 @@ export function extractContentType(body) {
...
+	// Detect form data input from form-data module
 	if (body && typeof body.getLengthSync === 'function') {
-		// Detect form data input from form-data module
-		if (body._lengthRetrievers && body._lengthRetrievers.length == 0 || // 1.x
-			body.hasKnownLength && body.hasKnownLength()) { // 2.x
-			return body.getLengthSync();
-		}
-
-		return null;
+		return body.hasKnownLength && body.hasKnownLength() ? body.getLengthSync() : null;
```
bitinn: This introduced a breaking change: form-data v1 support is dropped here, please document in changelog and upgrade guide

docs/v3-UPGRADE-GUIDE.md
```diff
@@ -33,7 +33,16 @@ Prior to v3.x, we included a `browser` field in the package.json file. Since nod
 
 ## Dropped the `res.textConverted()` function
 
-If you want charset encoding detection, please use [fetch-charset-detection] package.
+If you want charset encoding detection, please use the [fetch-charset-detection] package ([documentation][fetch-charset-detection-docs]).
+
+```js
+const fetch = require("node-fetch");
+const {convertBody} = require("fetch-charset-detection");
```
bitinn: should it just be `const convertBody = require("fetch-charset-detection");` ?

we only expose one function.

Richienb: The original problem was that using `export default` for ECMAScript would break when using CommonJS so I changed it to a named export to work consistently since using `module.export` would also indirectly break Typedoc.

I've now figured out that I can just export them using both methods so Typedoc, ECMAScript and CommonJS all work.

src/body.js
Richienb: @bitinn It looks like the only thing that is a breaking change is that form-data v2 [drops support for NodeJS v0.10](https://github.com/form-data/form-data/releases/tag/v2.0.0). The first item in the changelog already documents the minimum version requirement.

bitinn:
BUT: our tests keep failing, can we fix that first? Is it pending on some PR?

Richienb:
@bitinn The failing build is related to `Replace the calls to url.parse and url.resolve with calls to new URL to use the new UTF-8 awareness and remove the utf8 package.` in the v3 roadmap and therefore does not relate to this PR specifically.

bitinn:
@Richienb it's also failing in a test for data uri support.

`should accept data uri of plain text`

https://travis-ci.org/bitinn/node-fetch/jobs/613306983

bitinn:
But I guess due to linting issues no PR to `v3` will pass? Or was the linting test optional?

bitinn:
I would say we could have a test for using `fetch-charset-detection` in our test suite? It's not strictly necessary for unit testing, but nice to confirm it works *integration-wise*.

Richienb:
@bitinn 

> it's also failing in a test for data uri support.

That's not related to this PR.

> But I guess due to linting issues no PR to `v3` will pass? Or was the linting test optional?

It looks like VSCode autoformatted the code incorrectly. I'll run `yarn lint --fix` and push the changes.

> I would say we could have a test for using `fetch-charset-detection` in our test suite? It's not strictly necessary for unit testing, but nice to confirm it works _integration-wise_.

https://github.com/Richienb/fetch-charset-detection/blob/master/test.js

Richienb:
@bitinn It looks like the build for the PR may be bugged. However, the build for this [branch](https://travis-ci.org/bitinn/node-fetch/builds/613306957) isn't.

bitinn:
then it's good to go.

Richienb:
All reviewers approve. Time to merge!

PR #699, opened 2019-11-24
Use req.setTimeout for timeout

kennethaasan:
This PR is using the native Node way to set a timeout on a request:
https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback

I've been seeing a lot of weird _socket hang up_ errors when using _node_fetch_ on requests that are reaching the timeout I set. I'm using an agent with keep alive on. I'm wondering if this might be the reason.

Similar issue in axios:
https://github.com/axios/axios/pull/1752
https://github.com/axios/axios/blob/master/lib/adapters/http.js#L250-L261

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -83,7 +83,6 @@ export default function fetch(url, opts) {
 
 		// send request
 		const req = send(options);
-		let reqTimeout;
 
 		if (signal) {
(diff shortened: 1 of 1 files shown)


xxczaki:
Closed in favor of #719

Issue #549, opened 2018-11-13
Document types of fetch errors more thoroughly  


bitinn:
There used to be only one type of operational error from node-fetch: `FetchError`.

This is no longer the case as I can count at least six: `Error`, `NodeError`, `TypeError`, `SyntaxError`, `AbortError` and `FetchError`. All of them can be triggered in one way or another through invalid response.

We would like to unify them in some ways but it's (a) a breaking change, (b) sometimes not standard compliant (see #498).

So for those we can't wrap with `FetchError` in next major release, the doc needs to be updated:

https://github.com/bitinn/node-fetch/blob/master/ERROR-HANDLING.md

PR #700, opened 2019-11-29
fix: Throw SyntaxError instead of FetchError in case of invalid JSON

### LINKED ISSUE #498: Throw SyntaxError in case of invalid JSON

Richienb:
Closes #498.


bitinn:
also see #549, but it should be done in a separate PR.

PR #703, opened 2019-12-13
fix: Proper data uri to buffer conversion

Richienb:
Turns out, our implementation of data URI to buffer conversion was quite flawed. It always assumed base64 encoding (which wasn't the case when it came to the failing test) and tried to retrieve the non-existent `Buffer.body` property.

I was going to refactor the logic but I realised [`data-uri-to-buffer`](https://www.npmjs.com/package/data-uri-to-buffer) already did **exactly** what we wanted.

\
\
oh and the build is passing WOOO 🎉

cc @bitinn @xxczaki

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
 		"xo": "^0.25.3"
 	},
 	"dependencies": {
+		"data-uri-to-buffer": "^3.0.0",
 		"fetch-blob": "^1.0.4",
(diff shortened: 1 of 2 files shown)


Richienb:
@bitinn @xxczaki

Richienb:
We got enough approvals. Time to merge!

PR #705, opened 2019-12-18
This patch fixes a problem where not all prototype methods are copied from the Body via the mixin method due to a failure to properly detect properties in the target.

ekkis:
This patch fixes a problem where not all prototype methods are copied from the Body via the mixin method due to a failure to properly detect properties in the target.  The current code uses the `in` operator, which may return properties lower down the inheritance chain, thus causing them to fail the copy.  The new code properly calls the .hasOwnProperty() method to make the determination

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -165,7 +165,7 @@ Object.defineProperties(Body.prototype, {
 Body.mixIn = function (proto) {
 	for (const name of Object.getOwnPropertyNames(Body.prototype)) {
 		// istanbul ignore else: future proof
-		if (!(name in proto)) {
+		if (!proto.hasOwnProperty(name)) {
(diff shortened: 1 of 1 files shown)


Richienb:
Automatic branch conversion doesn't seem to work here. Please recreate this PR on the 3.x branch.

ekkis:
can we not apply the patch to the master branch directly?

Richienb:
@ekkis Version 3 is in development. Version 2 is no longer receiving updates.

ekkis:
ah.  I see.  I've recreated it in the correct branch.  how long do you think before this makes it to the main branch?

Richienb:
Moved to #706

Richienb:
@ekkis Probably on the first alpha or stable release of 3.x

PR #706, opened 2019-12-18
Flawed property existence test

ekkis:
This patch fixes a problem where not all prototype methods are copied from the Body via the mixin method due to a failure to properly detect properties in the target. The current code uses the in operator, which may return properties lower down the inheritance chain, thus causing them to fail the copy. The new code properly calls the .hasOwnProperty() method to make the determination


Richienb:
https://travis-ci.org/bitinn/node-fetch/jobs/626583990#L254

PR #707, opened 2019-12-27
do not set ContentLength to NaN

yaacovCR:
node-fetch should take advantage of @naugtur NaN workaround for the form-data package, see https://github.com/form-data/form-data/pull/397#issuecomment-471976669.

form-data package implementation of FormData allows for only two situations: streams of user pre-specified "knownLength" or streams with defined length retrievers (file streams with ends defined, files with sizes on disk, http requests with a content-length), but no concept of a regular old stream with an unknownable length.

naugtur suggests prespecifying the knownLength of such a stream as NaN, which will flow into the calculations and the eventual result of FormData.prototype.getLengthSync.

naugtur notes that this will cause the request package to not set the ContentLength header, and could even be considered a FormData feature.

node-fetch should support this "feature" as well.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -212,7 +212,8 @@ export function getNodeRequestOptions(request) {
 	}
 	if (request.body != null) {
 		const totalBytes = getTotalBytes(request);
-		if (typeof totalBytes === 'number') {
+		// set Content-Length if totalBytes is a number (that is not NaN)
+		if (typeof totalBytes === 'number' && totalBytes === totalBytes) {
 			contentLengthValue = String(totalBytes);
 		}
 	}
--- a/src/request.js
+++ b/src/request.js
@@ -213,7 +213,7 @@ export function getNodeRequestOptions(request) {
 	if (request.body != null) {
 		const totalBytes = getTotalBytes(request);
 		// set Content-Length if totalBytes is a number (that is not NaN)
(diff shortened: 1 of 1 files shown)


yaacovCR:
Related: https://github.com/form-data/form-data/issues/399

naugtur:
*I'm famous!* :smile: 

Yes, the NaN hack seems the only way to push the "unknowable length" through the function calls down to form-data so that both form-data and request work correctly.

I must add one thing: `a === a` is not the easiest to understand NaN check. 
I'd suggest this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

yaacovCR:
So famous!!!

Agree, isNaN is better. Thoughts from @bitinn ?

Richienb:
Please open PRs against the 3.x branch.

yaacovCR:
On further reflection, I think the underlying "right" thing is for FormData to not rely on use of NaN, but just to support streams of unknowable length out of the box.

See further discussion (hopefully!) at https://github.com/form-data/form-data/issues/394.

Although I guess still appropriate for node-fetch to check NaN, too. I will open a PR against the 3.0.0 branch as well.

Richienb:
Moved to https://github.com/bitinn/node-fetch/pull/709.

PR #709, opened 2019-12-31
Fix: Do not set ContentLength to NaN

yaacovCR:
As requested within #707, PR for v3 branch.

The underlying "right" thing may be for users of FormData to not have to rely on NaN, but for FormData to support streams of unknowable length out of the box, but it can't hurt for node-fetch to check for NaN in the meantime.

See further discussion (hopefully!) at form-data/form-data#394.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -206,7 +206,7 @@ export function getNodeRequestOptions(request) {
 
 	if (request.body != null) {
 		const totalBytes = getTotalBytes(request);
-		// set Content-Length if totalBytes is a number (that is not NaN)
+		// Set Content-Length if totalBytes is a number (that is not NaN)
 		if (typeof totalBytes === 'number' && !Number.isNaN(totalBytes)) {
 			contentLengthValue = String(totalBytes);
 		}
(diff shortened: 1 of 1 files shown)


naugtur:
The problem is you're using form-data via request.

AFAIR form-data itself is accepting any falsy value passed in the right way, but for request to not add another incorrect content lenght it had to be NaN

A change in request would be necessary

yaacovCR:
Via node-fetch, which calls getLengthSync and gets incorrect content length.

Why shouldn't FormData just figure out that some streams have unknowable lengths and start returning null or undefined from getLengthSync. I achieved that in a class extending FormData at the link but could be integrated within base FormData package.

naugtur:
@yaacovCR I don't remember exactly anymore, but there was one more issue with getting request and form-data to cooperate correctly in that case. 
If lenght if the stream is unknown you not only need to avoid setting content lenght on the multipart section but also the header on the request and there was something wrong with the assumptions there as well.  
But I don't remember enough details to tell whether fixing that function would solve both

yaacovCR:
https://github.com/request/request/issues/1402
Not hanging with my contrived graphql example for node-fetch. You do have to set the content Type.

PR #719, opened 2020-01-12
fix: Use req.setTimeout for timeout

xxczaki:
Original PR by @kennethaasan: #699 

Created against the `3.x` branch.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -86,7 +86,6 @@ export default function fetch(url, opts) {
 
 		// Send request
 		const req = send(options);
-		let reqTimeout;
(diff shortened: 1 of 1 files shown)


jimmywarting:
FYI

https://github.com/node-fetch/node-fetch/issues/523#issuecomment-438158003

> now that v2.3.0 is released with AbortSignal support, I think we opened up a path to deprecate timeout in v3.0, where we would likely deprecated support for node < 8.

...haven't been active for a while on node-fetch but i thought i would check back on how things are going

PR #746, opened 2020-03-13
Fix package.json

bitinn:


bitinn:
Actually, my PR is also wrong. We should be publishing the `dist` folder, but I don't think we are. I need to look at it more.

bitinn:
Current `pika` build doesn't work, even if I install the `dist` folder locally, it still throws this:

```
$ node test.js
internal/modules/cjs/loader.js:613
    throw err;
    ^

Error: Cannot find module './body.js'
Require stack:
- /Users/df/git/node-fetch/dist/dist-node/index.js
```

Maybe it's due to my lack of understanding of this tool, but this is not working as I imagine. @Richienb @xxczaki

Issue #749, opened 2020-03-15
v3 incorrectly strips out GET params [?]


dzek69:
Example code, use Node 13.2+:

`index.mjs`
```javascript
import express from "express";
import f from "node-fetch";
const fetch = f.default; // Native ESM

const app = express();
app.get("*", (req, res) => {
    console.log("srv", req.url)
    res.send("yo");
});
app.listen(5001);

(async () => {
    const res = await fetch("http://127.0.0.1:5001/api/getStuff?a=tra").then(r => r.text())
    console.log(res);
})().catch(e => {
    console.error(e);
});
```

With `3.0.0-beta.4` version of `node-fetch` you'll see just `srv /api/getStuff` in the console
With `2.6.0` you will see `srv /api/getStuff?a=tra` - as expected

So it looks like somehow GET query params gets stripped out.

KhafraDev:
Can't reproduce on Windows 10/Node v13.9.0 with your example code. Tried it a bit and then modified the script to run a couple thousand times. ->
```js
(node:7816) ExperimentalWarning: The ESM module loader is experimental.
srv /api/getStuff?a=tra
yo
```

dzek69:
I have prepared a repository:
https://github.com/dzek69/node-fetch-bug-demo

Use `yarn` to keep the same versions as me.
Run `yarn start` to test on your local node version
Run `yarn docker` to run in docker with node 13.10.1
Run `yarn docker:12` to run in docker with node 12

I have also built and published images on docker hub:
dzek69/node-fetch-bug:12 -- node-fetch v3 + node 12
dzek69/node-fetch-bug:13 -- node-fetch  + node 13
dzek69/node-fetch-bug:v2 -- node-fetch  + node 13 -- this works correctly, because it's v2 version

In every environment this behaves the same for me (that's why I've build docker images so you can see it too if this somehow runs correctly on your machine)

If I run docker images on my Linux machine it behave the same as when run from windows:
```
dzek@srv:~$ docker run dzek69/node-fetch-bug:12
Unable to find image 'dzek69/node-fetch-bug:12' locally
12: Pulling from dzek69/node-fetch-bug
c9b1b535fdd9: Pull complete
6ad7c27326e7: Pull complete
968d8b4948da: Pull complete
a1ec837a6331: Pull complete
f462b809bd5c: Pull complete
bc393c91984a: Pull complete
[...]

PR #759, opened 2020-03-22
fix: Ensure search parameters are included in URL path

terminal69:
#749 can confirm that this is indeed a bug, introduced at commit 0930d6dedc08fa2adde9949a1a36d98179902582
path should be pathname + search


terminal69:
> Please make sure tests are passed ;)

Please see #760

terminal69:
rebased to latest commit on master and tests passed

PR #761, opened 2020-03-25
fix: Import URL and URLSearchParams in typings

maxdumas:
This change adds an explicit import for the URL and URLSearchParams typings from the Node standard library `'url'` module.

This was needed to prevent the TypeScript compiler from issuing errors when compiling against targets where URL and URLSearchParams are not ambiently available.


maxdumas:
Note that [tests already appear to be failing on the upstream](https://travis-ci.com/github/node-fetch/node-fetch/jobs/298128208). This change doesn't introduce any new test failures AFAICT.

xxczaki:
Thank you!

PR #767, opened 2020-04-08
fix: requestOptions should conform to the http(s).request interface

### LINKED ISSUE #749: v3 incorrectly strips out GET params [?]

jstewmon:
The requestOptions being returned did not conform to the options
interface expected by node's request function. Specifically, the path
option must include the pathname and search properties from a URL
instance, host is ignored when hostname is present, and the following
are not recognized: hash, search, query, href

fixes #749

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -260,15 +260,10 @@ export function getNodeRequestOptions(request) {
 
 	// manually spread the URL object instead of spread syntax
 	const requestOptions = {
-		path: parsedURL.pathname,
-		pathname: parsedURL.pathname,
+		path: `${parsedURL.pathname}${parsedURL.search}`,
 		hostname: parsedURL.hostname,
 		protocol: parsedURL.protocol,
 		port: parsedURL.port,
-		hash: parsedURL.hash,
-		search: parsedURL.search,
-		query: parsedURL.query,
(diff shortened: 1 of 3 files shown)


jstewmon:
🤦‍♂ Sorry, I totally missed the reference to #759 when I scrolled through #749. Please feel free to close. I'll leave this here just in case, since it looks like the conversation may have stalled on #759.

xxczaki:
Closed, as #759 got merged ;) Thank you anyway.

PR #768, opened 2020-04-08
Settle `consumeBody` promise when the response closes prematurely

jstewmon:
👋 Hi, the newly added test demonstrates that the `consumeBody` promise would never settle if the body stream was closed prematurely by the server.

---------

### fix: consumeBody promise should settle when the response closes prematurely

The stream.finished function is used consolidate the terminal cases.
The writable option must be set to false, since the body stream is
a Duplex stream, but our terminal cases only apply to the Readable side.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -5,7 +5,7 @@
  * Body interface provides common methods for Request and Response
  */
 
-import Stream, {PassThrough} from 'stream';
+import Stream, {finished, PassThrough} from 'stream';
 import {types} from 'util';
 
 import Blob from 'fetch-blob';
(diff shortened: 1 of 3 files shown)


jimmywarting:
> There is a single timeout option which applies to both the receiving of
response headers and the receiving of the entire response body. Once
the response is received, the socket timeout must be cleared to allow
the timeout to apply to the receiving of the entire response body.

That is a changing/breaking behavior for ppl using timeout already. in fact, timeout is discouraged and was something we added long ago in the beginning before [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) where even a thing

Using a abort-singal is the way to go now and is more cross node/web compatible. You have control over when it should timeout (request, response or both)...

I wish you broke the PR up to two... I'm more prone to accept the consumeBody than a timeout change.

jstewmon:
Hi @jimmywarting 👋 , I moved the timeout fix to #770 - I'll ref your comment above and respond over there, so we can keep the conversation focused.

jstewmon:
The build is "working" - the failure is the flaky `should allow custom timeout on response body` test that is addressed in #770

jstewmon:
👋 hi, wanted to see if there’s any more feedback on this PR? The bug demonstrated by the test is one I have to workaround by not using node-fetch in a particular case. I also think it is the likely culprit of the linked async iterator issue.

tinovyatkin:
@jstewmon Can you please fix merge conflicts on this?

jstewmon:
@tinovyatkin rebased and ready to go 🙂

PR #770, opened 2020-04-10
fix: Disambiguate timeout behavior for response headers and body

jstewmon:
There is a single timeout option which applies to both the receiving of
response headers and the receiving of the entire response body. Once
the response is received, the socket timeout must be cleared to allow
the timeout to apply to the receiving of the entire response body.

Without clearing the socket timeout, if the nearing the idle timeout
when the Promise for the body is created, then the request's timeout
handler will abort the request, emitting ERR_STREAM_PREMATURE_CLOSE in
the body Promise.

By clearing the socket timeout once the response headers are received,
the timeout for the entire body can be started when the body is awaited.
Since the request will no longer be aborted by the socket timeout,
destroy is called on the body to prevent it continuing to emit data
events.

--------

Picked from #768

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -200,7 +200,9 @@ function consumeBody() {
 		if (this.timeout) {
 			resTimeout = setTimeout(() => {
 				abort = true;
-				reject(new FetchError(`Response timeout while trying to fetch ${this.url} (over ${this.timeout}ms)`, 'body-timeout'));
+				const err = new FetchError(`Response timeout while trying to fetch ${this.url} (over ${this.timeout}ms)`, 'body-timeout');
+				reject(err);
+				body.destroy(err);
 			}, this.timeout);
 		}
 
--- a/src/index.js
+++ b/src/index.js
@@ -117,6 +117,7 @@ export default function fetch(url, options_) {
 		});
 
 		request_.on('response', res => {
+			request_.setTimeout(0);
 			const headers = createHeadersLenient(res.headers);
 
 			// HTTP fetch step 5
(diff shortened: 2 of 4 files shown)


jstewmon:
Responding to @jimmywarting https://github.com/node-fetch/node-fetch/pull/768#issuecomment-612234113

> That is a changing/breaking behavior for ppl using timeout already. in fact, timeout is discouraged and was something we added long ago in the beginning before AbortController where even a thing

> Using a abort-singal is the way to go now and is more cross node/web compatible. You have control over when it should timeout (request, response or both)...

Let's set aside the desire to remove timeout, since it is still here, and I believe this is a useful bug fix as long as timeout remains.

I don't think this is a breaking change - can you elaborate on why you see it that way?

I'm sure you noticed that test `should allow custom timeout on response body` is flaky - that's what is being fixed here. I left that test unchanged to demonstrate that it would reliably pass now, and I added a new test to demonstrate why.

Previously, if `timeout` was specified and the response headers were received before the timeout, the `body` Promise may have rejected with a `ERR_STREAM_PREMATURE_CLOSE` due to the `abort` call from the timeout handler that was added to the request. I considered this a bug because a timeout led to a rejection which was not recognizable as such.

Further, the timer that is started when the body is awaited covers all data events emitted by the stream, which is to say the entire body must be received before the timer is due. The socket timeout is an _idle_ timeout, which is reset on each `data` event. So, the change that I have made only affects use cases where there was a significant delay between the response being received and the body being awaited. The impact on such cases is that when the body is awaited, the promise will no longer reject with `ERR_STREAM_PREMATURE_CLOSE` if the socket was idle between the response being received and the body being awaited.

So, I would say the existing behavior was "undefined" for this edge case, and now it is defined in a way that is consistent with how the API has long been described to work.

jstewmon:
Thanks for the reviews @tinovyatkin and @NotMoni 🙏  I rebased to resolve the merge conflict.

PR #777, opened 2020-04-13
fix question mark stripped from url when no params are given (#776)

### LINKED ISSUE #776: v3 incorrectly strips out question mark from URL when no query params after question mark
This bug is somehow related to #749 

When I want to get an url like that: "https://example.com/resource?" the path sent to the server should include a question mark, even I specified no parameters. v3 node-fetch strips question mark out. Similar issue for python library, where question if this should be included or not has already been discussed is here: https://github.com/psf/requests/issues/2912, especially see this: https://github.com/psf/requests/issues/2912#issuecomment-162345429 - tl;dr spec defines that question mark should be included in this case

This works as expected in node-fetch v2 and of course works as expected in browser "real" `fetch`

Demo code for node:
```
const fetch = require("node-fetch").default;

(async () => {
    const ny = await fetch("https://s.o7o.pl/cUNtAD?").then(r => r.text());
    console.log("is result ok?", ny.includes("s.o7o.pl"));
})().catch(e => console.error("main", e));
```

Demo for "real" fetch in browser:
- head to https://s.o7o.pl/ (to avoid CORS issues)
- open dev tools console
- paste code from node demo, but strip out `require` line

Explaination of demo:
s.o7o.pl is a private use url shortener, under URL: `https://s.o7o.pl/cUNtAD` there is redirection to npmjs.com page, but when question mark is added to url then html page with target url is shown instead (similar to bit.ly behavior, but they use "+" sign). Some other APIs may rely on that.

PR incoming.

dzek69:
note: `new URL()` will return both `search` and `hash` with their starting character (`?` or `#`) but not when these parameters are empty, including when only their starting characters are given in url

however `href` keeps them both and this fix relies on that

`hrefWithoutHash` is technically incorrect when on that edge case with hash, but that doesn't matter here so I haven't wrote a code that doesn't matter in the final result

Fixes #776


src/utils/get-search.js
```diff
@@ -0,0 +1,12 @@
+export function getSearch(parsedURL) {
```
jstewmon, line 1: I'm not sure a helper function is needed for this. It can be written concisely without any function calls or allocations like so:
```js
u.search || u.href[u.href.length - u.hash.length - 1] === '?' ? '?' : ''
```

dzek69: I can do that, but I usually avoid "magic looking code" (or just unreadable at first sight) even if I personally like that. Decision is up to node-fetch authors.

dzek69: Oh, and your code will fail, it should look for `?` with `includes`

jstewmon: Hmm... What input will cause it to fail? Using `includes` will perform a linear search from the beginning of the string. Given that we know the input must have been valid to have been parsed by URL, we should be able to safely check the end of the `href`, minus the `hash`.

dzek69: It will fail for http://example.com/?# but if you include some hash check too it should be okay to test only the last character I think

jstewmon: Ah, good spotting 😄 I suppose unit tests would be useful for covering such edge cases. How about this?

```js
function getSearch(u) {
  if (u.search) {
    return u.search;
  }
  const lastOffset = u.href.length - 1;
  const hash = u.hash || u.href[lastOffset] === '#' ? '#' : '';
  return u.href[lastOffset - hash.length] === '?' ? '?' : '';
}
```

dzek69: Seems good to me. I have added an unit test for the search string already, so updating my solution with yours should pass the tests

I don't know if just the hash alone causes issues anywhere, as it is not send to the server anyway

dzek69: I have pushed an update with your suggestion (and fixed missing parentheses). I have replaced `u` with `parsedURL` for readability (build job should minify the code, seems like builds are kind of work in progress so I guess this will come later).

Thank you for your suggestions

src/utils/get-search.js
```diff
@@ -0,0 +1,9 @@
...
+	if (parsedURL.search) {
+		return parsedURL.search;
```
xxczaki, line 3: This would need to be tested

src/utils/get-search.js
```diff
@@ -0,0 +1,9 @@
...
+	}
+
+	const lastOffset = parsedURL.href.length - 1;
+	const hash = parsedURL.hash || (parsedURL.href[lastOffset] === '#' ? '#' : '');
```
xxczaki, line 7: And this

xxczaki:
See #782

Issue #668, opened 2019-09-07
v3 Roadmap


xxczaki:
**Note: v3 beta is now available on npm under the `next` tag**

---

# v3 Roadmap

## General stuff

- [x] **BREAKING:** Drop support for Node.js 4 and Node.js 6 (#489, #624)
- [x] Update devDependencies (related to #489)
- [x] When users clone(), automatically create streams with custom highWaterMark (#386, PR open: #671)
- [x] Bundle TypeScript types (Support for TypeScript without need of installing `@types/node-fetch` package) (#669)
- [x] Replace Rollup with Microbundle (#643)
- [x] Introduce linting to the project (#303)
- [x] Simplify Travis CI build matrix (related to #431) - Dropping support for Node.js 4 & 6 will help
- [x] Response.statusText should not default (#578)
- [x] Use w3c defined message for AbortError (#657)
- [x] URL is not encoded as UTF-8 (#245)
- Deprecate timeout in favor of AbortController (#523)
- [x] Data URI support (#659)
- [x] Drop the optional dependency on encoding and move charset detection into a separate package. (In progress: [Richienb/fetch-charset-detection](https://github.com/Richienb/fetch-charset-detection) (PR Open: #694))
- [x] Drop existing blob implementation code and use fetch-blob as dependency instead (https://github.com/bitinn/node-fetch/issues/668#issuecomment-529175627)
- [x] Fix stream piping (#309, PR open: #670)
- [x] Fix stream resolving (#642, #665)
- [x] Do not use constructor.name to check object (#667, PR open: #673)
- [x] Properly forward request errors while piping response stream
- [x] Convert `Content-Encoding` to lowercase (#661, PR open: #672)
- [x] More tests for data URI (https://github.com/bitinn/node-fetch/issues/668#issuecomment-529175096)
- [ ] Drop credentials before redirection (#274)
- [x] Modernise the code behind FetchError and AbortError (https://github.com/bitinn/node-fetch/issues/668#issuecomment-530845161)
- [x] Propagate size and timeout to cloned response (#664)
- [x] Replace the calls to `url.parse` and `url.resolve` with calls to `new URL` to use the new UTF-8 awareness and remove the `utf8` package.
    - [x] `url.parse` (https://github.com/node-fetch/node-fetch/pull/701)
	- [x] `url.resolve` (https://github.com/bitinn/node-fetch/commit/e3cd4d22c3f1871fb41b17035ac83b47ad57fdf6)
[...]

PR #778, opened 2020-04-14
Compile CJS modules as a seperate set of files

Richienb:
See: https://github.com/node-fetch/node-fetch/issues/668#issuecomment-612937339


xxczaki:
We might also want to provide the `"module": "src/index.js"` and `"sideEffects": false` fields in our `package.json` file for bundlers like Webpack.

- https://webpack.js.org/guides/tree-shaking/
- https://webpack.js.org/guides/author-libraries/#final-steps
- https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md#typical-usage

PR #782, opened 2020-04-20
fix question mark stripped from url when no params are given (#776) 

### LINKED ISSUE #776: v3 incorrectly strips out question mark from URL when no query params after question mark

xxczaki:
Original PR by @dzek69 (#777)

Fixes #776

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
 		"chai-as-promised": "^7.1.1",
 		"chai-iterator": "^3.0.2",
 		"chai-string": "^1.5.0",
-        "coveralls": "^3.0.11",
+		"coveralls": "^3.0.11",
 		"form-data": "^3.0.0",
 		"mocha": "^7.1.1",
 		"nyc": "^15.0.1",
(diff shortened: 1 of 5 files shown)


xxczaki:
Thank you @dzek69!

dzek69:
no prob, and sorry for no no response and not adding those tests, i'm quite busy, thanks for finishing this for me 👍

PR #795, opened 2020-05-07
Allow passing BODY in GET request via configurable Parameter in OPTIO…

ksarpotdar:
…NS, defaulted to ERROR

Add configurable Parameter overrideGetBodyCheck, if set to true in OPTIONS, then it would bypass Error check for the Request "Body" in the GET request. It would allow the request if the parameter is set to true, otherwise, it would reject the GET requests with the BODY.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -82,15 +82,15 @@ export default class Request {
 		let method = init.method || input.method || 'GET';
 		method = method.toUpperCase();
 
-		let checkBodyllowedInGET = false;
+		let allowBody = false;
 		if ((init.overrideGetBodyCheck && init.overrideGetBodyCheck === true) ||
 			(isRequest(input) && input.overrideGetBodyCheck && input.overrideGetBodyCheck === true)) {
-			checkBodyllowedInGET = true;
+			allowBody = true;
 		}
 
 		// eslint-disable-next-line no-eq-null, eqeqeq
 		if ((init.body != null || isRequest(input) && input.body !== null) &&
-			(method === 'GET' && checkBodyllowedInGET === false || method === 'HEAD')) {
+			(method === 'GET' && allowBody === false || method === 'HEAD')) {
 			throw new TypeError('Request with GET/HEAD method cannot have body');
 		}
 
--- a/src/request.js
+++ b/src/request.js
@@ -83,7 +83,7 @@ export default class Request {
 		method = method.toUpperCase();
 
 		let overrideGetBodyCheck = false;
-		// eslint-disable-next-line no-eq-null
+		// eslint-disable-next-line no-eq-null, eqeqeq
 		if ((init.body != null && init.allowBody === true) ||
 			(isRequest(input) && input.body !== null && input.allowBody === true)) {
 			overrideGetBodyCheck = true;
--- a/src/request.js
+++ b/src/request.js
@@ -82,15 +82,16 @@ export default class Request {
 		let method = init.method || input.method || 'GET';
 		method = method.toUpperCase();
 
-		let allowBody = false;
-		if ((init.overrideGetBodyCheck && init.overrideGetBodyCheck === true) ||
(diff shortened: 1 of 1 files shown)


xxczaki:
@jimmywarting AFAIK something like this is against the spec (https://tools.ietf.org/html/rfc7540#section-8.1.3). 

I've seen an issue in the got repository, which was about why this should be possible: https://github.com/sindresorhus/got/issues/966

At the end, they've added the `allowBody` option.

MitMaro:
> @jimmywarting AFAIK something like this is against the spec ([tools.ietf.org/html/rfc7540#section-8.1.3](https://tools.ietf.org/html/rfc7540#section-8.1.3)).
> 
> I've seen an issue in the got repository, which was about why this should be possible: [sindresorhus/got#966](https://github.com/sindresorhus/got/issues/966)
> 
> At the end, they've added the `allowBody` option.

I think it's worth adding that it's not completely against the spec as it was supported before HTTP/2.

In particular the [documentation in got](https://github.com/sindresorhus/got#allowgetbody) has:

> Note: The RFC 7321 doesn't specify any particular behavior for the GET method having a payload, therefore it's considered an anti-pattern.
> 
> Set this to true to allow sending body for the GET method. However, the HTTP/2 specification says that An HTTP GET request includes request header fields and no payload body, therefore when using the HTTP/2 protocol this option will have no effect. This option is only meant to interact with non-compliant servers when you have no other choice.

In particular, a reason to potentially include this is that [Elasticsearch allows the use of a request body on GET requests](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html).

In the end I guess it would depend on if the Fetch spec allows a body for a GET request.

ksarpotdar:
Dear all,

I don't know why request body is not allowed in GET specifications. I don't see any reason, especially clearly mentioned in the documentations.
1. Please refer the specs here. https://tools.ietf.org/html/rfc2616#page-32 It clearly says that "The rules for when a message-body is allowed in a message differ for requests and responses." or https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html
2. The same GET Request can return the body, so is there any specific reason, that the GET request is not allowed? 
3. Earlier GET used to be only for the web page, when browser opens up the request to server. With more and more REST APIs and the data to be exchanged exceeding 2048 characters, need to be embedded in the GET request BODY.  With REST APIs, the things have changed, without violating the Specs. All REST APIs use the same specs.
4.  BODY in GET is more secured, as against the URL, as the URL is saved and cached by browsers or the applications and visible, while the body is not. Even though the request and response is not cached, the URL is still visible
5. This is a middle-ware and we should not implement the entire specs here. The back-end server, which is actually serving the purpose should do that job. Our job is to provide interface to the server making request and let that server respond whether something is allowed or not allowed. So, when in doubt, if something is required or not, this principle should be followed allowing the back-end to handle it.
6. What I have done is that the default behavior of the error handling remains the same. Its only when the sender specifically wants to override that behavior, they have to enable it in the CONFIG. So that by default it implements the specs, unless there is specific change by the caller.

Please consider this request. Either we can delete the Error, which is thrown in GET request, or we can implement what I have done. We can change the variable name to allowBody as well, if there is no objection. 

Node-fetch is a great middle-ware and I would request to implement this functionality, to enhance it better.

bitinn:
See:

https://github.com/whatwg/fetch/issues/83
https://github.com/whatwg/fetch/issues/551
https://github.com/httpwg/http-core/issues/202
https://github.com/httpwg/http-core/pull/250

Fetch spec doesn't support it, Browsers implementers don't like it, HTTP spec is actively against it, so I am negative on this.

jimmywarting:
I'm with @bitinn on this. It's not allowed in browser. It's against the spec so it is a anti-pattern that should not be followed. 
So unless you can convince the spec to change then we will probably not include this as we are trying to follow the spec as close as possible.

xxczaki:
Closing this, as it's against the current specification and we are not going to support it.

Issue #724, opened 2020-02-11
Extend options param to accept options that can be forwarded to http/https request


jamiechong:
It would be great if the [current options](https://github.com/node-fetch/node-fetch#fetch-options) accepted another value that could forward options to the internal [http.request](https://nodejs.org/api/http.html#http_http_request_options_callback) call. Specifically I'm looking to use the `insecureHTTPParser` param to work around [this issue](https://github.com/nodejs/node/issues/27711).

So something like:
```
fetch(url, { extraHttpOptions: { insecureHTTPParser: true } });
```

xxczaki:
@bitinn @Richienb What do you think? Should we allow users to pass options to the internal `http.request` call?

willhoag:
I also have this need, trying to solve the same problem:
https://github.com/nodejs/node/issues/27711

@jamiechong were you able to solve this (request specific insecure parsing) any another way?

mehmetcc:
> I also have this need, trying to solve the same problem:
> [nodejs/node#27711](https://github.com/nodejs/node/issues/27711)
> 
> @jamiechong were you able to solve this (request specific insecure parsing) any another way?

This is still a valid problem you guys

NotMoni:
We should have the option to do insecure https.

NotMoni:
@jamiechong Have you tried, this?
```js
const https = require('https')
const fetch = require('node-fetch);

const agent = new https.Agent({
// your HTTP options
})
const data = fetch(uri, { agent })
      rejectUnauthorized: false
})

**Comment by jamiechong** [comment]
> @jamiechong Have you tried, this?
> 
> ```js
> const https = require('https')
> const fetch = require('node-fetch);
> 
> const agent = new https.Agent({
> // your HTTP options
> })
> const data = fetch(uri, { agent })
>       rejectUnauthorized: false
> })
> ```

I don’t believe I have but that looks promising. It’s been quite a while since I’ve tried to resolve this (no longer at the organization where this problem exists). But for my own interest I’ll try to test and see if it handles the problem we had. Thanks!!

**Comment by NotMoni** [comment]
Np, let me know if it solves the problem 🦄

**Comment by roychri** [comment]
Unfortunately, it does not help.

Here's a simple way to reproduce the problem...

Install latest node 12.x version

Create a file called `test.js` with this content:

```
const https = require('https');
const fetch = require('node-fetch');
async function main() {
    const agent = new https.Agent({
        rejectUnauthorized: false
    });
[...]

PR #799, opened 2020-05-12
Add TLS options

### LINKED ISSUE #791: getNodeRequestOptions does not honor all the https.request options
https://nodejs.org/api/https.html#https_https_request_url_options_callback documents the options for calling https.request but there is a side note about all the possible tls.connect options.

> The following additional options from tls.connect() are also accepted: ca, cert, ciphers, clientCertEngine, crl, dhparam, ecdhCurve, honorCipherOrder, key, passphrase, pfx, rejectUnauthorized, secureOptions, secureProtocol, servername, sessionIdContext, highWaterMark

These options are not implemented or possible to set through `getNodeRequestOptions`,
this makes it impossible to pin a ca / certificate or specific protocols.

julianhille:
Fixes #791.

Request function takes a bunch of tls options when https is used.
These options are currently not possible to be set through node-fetch.

The possibilities are documented here: 

https://nodejs.org/api/https.html#https_https_request_url_options_callback

This PR just adds the option `tlsOption` to a node fetch call, which is then passed on to `https.request`


xxczaki:
Somewhat related to https://github.com/node-fetch/node-fetch/issues/724. If we were to add additional TLS options, we might also allow people to set other http/s options, like `insecureHTTPParser`.

jimmywarting:
can this be solved by using a own agent?

julianhille:
> can this be solved by using a own agent?

Not as far as i know. But sometimes the docs are a bit misleading and nested.
What i've read:
https://nodejs.org/api/http.html#http_new_agent_options
https://nodejs.org/api/net.html#net_socket_connect_options_connectlistener
https://nodejs.org/api/https.html#https_new_agent_options

julianhille:
> Somewhat related to #724. If we were to add additional TLS options, we might also allow people to set other http/s options, like `insecureHTTPParser`.

At least setting the tlsOption is mandatory imho.
I can't connect to a server with a specific authority set or a client certificate as auth request.

src/request.js
```diff
@@ -138,6 +138,7 @@ export default class Request {
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
 		this.highWaterMark = init.highWaterMark || input.highWaterMark;
+		this.tlsOptions = init.tlsOptions || {};
```
xxczaki, line 138: I'm not sure if that's the way we want to do it. If we were to support [passing additional options to HTTP/S request](https://github.com/node-fetch/node-fetch/issues/724), we would need to add yet an additional node-fetch-only option, like `extraHttpOptions`, which makes everything much more complicated.

cc @node-fetch/reviewers

bitinn: I think we need to test if TLS options aren't actually supported by https.agent;

The document on Agent has been missing options info for years (the `support socket.connect() options` line was added recently).

My bet is it could be missing options for TLS as well. I know for sure it supports client authentication.

JefferyHus: The documentation or event the `node` core isn't clear about the options object passed to either `new http()` or `http.agent()`. However I saw few people doing this with the `https` object:

```javascript
var options = {
  key: '',
  cert: ''
};```

After checking the documents and the `node` core code I only was able to see this [How to use TLS](https://nodejs.org/en/knowledge/cryptography/how-to-use-the-tls-module/)

This should be tested further to be able to define a clean solution

julianhille: It seems like the agent does support it. I found some node modules using the same options and some stack overflows suggesting that it works. Will test it later and report back

julianhille: Im happy to confirm that setting these options on the agent works.
I tested:
`ca`, `cert`, `key`, `rejectUnauthorized`, `secureOptions`, `secureCipehrs` i tend to believe that all the other options work too

julianhille:
Not needed as it seems that http(s).Agent is just underdocumented and allows us to set these options in node.

Thanks for pointing it out, that the doc might not be helpful.

PR #805, opened 2020-05-19
drop Babel (while keeping ESM)

tinovyatkin:
While version 3.x of `node-fetch` is targeting Node 10.x or later, there is little or nothing to do for Babel (add `debug` option in `@babel/preset-env` to see): it only transforms optional chaining, nullish coercing, dynamic imports and ESM modules.

However, there are no mentioned language features in the current codebase, except modules, and there are way more appropriate tools for module conversion.

This PR drops Babel and 7 related dependencies in favor of `rollup`, dedicated only to module conversions and producing better output.

To do so, we also forced to fix _hackish_ named exports (named exports and properties on default export, as [introduced in this commit](https://github.com/node-fetch/node-fetch/commit/d15476bb5d60eef718592bcfd451cfd616128070#diff-b9cfc7f2cdf78a7f4b91a753d10865a2), are not the same thing) with appropriate language construction in the `index.js`

As `nyc` depends on Babel this PR also replaces `nyc` with `c8` and native V8 based code-coverage reports.

This also fixes the bug with version 3.0-beta.5 of the library that was shipped with Istanbul instrumentation included in the production code.

The produced CommonJS code is also more beautiful and performant:

```js
// ---- current 3.0.0-beta.5 as published in dist/ ----
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = fetch;

var _http = _interopRequireDefault(require("http"));

// ...

var _request = _interopRequireWildcard(require("./request"));

var _fetchError = _interopRequireDefault(require("./errors/fetch-error"));

var _abortError = _interopRequireDefault(require("./errors/abort-error"));

function _getRequireWildcardCache() { if (typeof WeakMap !== "function") return null; var cache = new WeakMap(); _getRequireWildcardCache = function () { return cache; }; return cache; }

function _interopRequireWildcard(obj) { if (obj && obj.__esModule) { return obj; } if (obj === null || typeof obj !== "object" && typeof obj !== "function") { return { default: obj }; } var cache = _getRequireWildcardCache(); if (cache && cache.has(obj)) { return cache.get(obj); } var newObj = {}; var hasPropertyDescriptor = Object.defineProperty && Object.getOwnPropertyDescriptor; for (var key in obj) { if (Object.prototype.hasOwnProperty.call(obj, key)) { var desc = hasPropertyDescriptor ? Object.getOwnPropertyDescriptor(obj, key) : null; if (desc && (desc.get || desc.set)) { Object.defineProperty(newObj, key, desc); } else { newObj[key] = obj[key]; } } } newObj.default = obj; if (cache) { cache.set(obj, newObj); } return newObj; }

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

function cov_fps9prjc6() {
  var path = "/home/xxczaki/dev/node-fetch/src/index.js";
  var hash = "a8046cbf7359c2bb2f2d15890c2c22fb8edba5f4";
  var global = new Function("return this")();
  var gcv = "__coverage__";
  var coverageData = {
//
```

code from `dist/index.cjs` after merging this PR:
```js
'use strict';

exports = module.exports = fetch;

const http = require('http');
const zlib = require('zlib');
const Stream = require('stream');
const dataURIToBuffer = require('data-uri-to-buffer');
const Blob = require('fetch-blob');
const url = require('url');

/**
 * Fetch-error.js
 *
 * FetchError interface for operational errors
 */

/**
 * Create FetchError instance
 *
 * @param   String      message      Error message for human
 * @param   String      type         Error type for machine
 * @param   Object      systemError  For Node.js system error
 * @return  FetchError
 */
class FetchError extends Error {
// ...
}
```


tinovyatkin:
One thing where I expecting feedback from project maintainers is regarding extension use for imports: current Node.JS specification (included in LTS) required file extensions and I see intentions to ship 3.x as ECMAScript compatible module, so, maybe it's a good time to review this rule?

jimmywarting:
unrelated... but i feel like javascript needs a breather from coming up with new syntax features so browser, node and deno can focus more on on 
1) implementing missing language syntax (like optional chaining)
2) fix stuff that can't be polyfilled (like web push in Safari, The presentation api and secure sandboxing model for NodeJS, etc) - some things can be solved by the community
3) Node should be more web spec compliant and implement some of the browser api's

I don't want to have to depend on Babel, rollup or WebPack for every new project i start to build.

tinovyatkin:
@Richienb 
A `.cjs` extension is neither my invention nor a personal preference - it's documented and recommended by Node.JS documentation: https://nodejs.org/api/esm.html#esm_writing_dual_packages_while_avoiding_or_minimizing_hazards
What is your reasoning behind `.js` here?

Richienb:
Since we're only compiling the `.cjs` files for distribution, it would be ok then.

package.json
```diff
@@ -43,33 +43,31 @@
...
-		"babel-plugin-add-module-exports": "^1.0.2",
-		"babel-plugin-istanbul": "^6.0.0",
+		"c8": "^7.1.2",
 		"chai": "^4.2.0",
 		"chai-as-promised": "^7.1.1",
 		"chai-iterator": "^3.0.2",
 		"chai-string": "^1.5.0",
-		"coveralls": "^3.0.13",
+		"coveralls": "^3.1.0",
+		"esm": "3.2.25",
```
LinusU, line 54: Maybe add a caret here?

```suggestion
		"esm": "^3.2.25",
```

tinovyatkin:
`c8` uses native code coverage from V8, while `nyc` so named _code instrumentation via babel and Istanbul. It's normal than when you are starting to use more precise instruments you get better results.

LinusU:
Hmmm, seems like a bug though?

https://coveralls.io/builds/30915791/source?filename=src%2Frequest.js#L243

tinovyatkin:
It seems that there is an ongoing issue between `esm` and `c8`: https://github.com/bcoe/c8/issues/34
So, my proposed solution is to fix all imports and make `node-fetch` work as native ESM on Node 14 or Node LTS (done at [a5ccd20](https://github.com/node-fetch/node-fetch/pull/805/commits/a5ccd20c74b8066fba0dba7f7924a5683e703e1a)) and then run coverage only on Node 14 that runs non-transpiled code (and it results in real coverage) - [dc9e992](https://github.com/node-fetch/node-fetch/pull/805/commits/dc9e992f2051b5774351340b0bc9a2a67767e18c). Now we only need to enable GitHub Actions on this repo.

PR #810, opened 2020-05-21
Breaking: Revamp TypeScript declarations

tinovyatkin:
The decision to bring types declaration bundled since version 3.0 is a good move, however, it also makes maintainers of the project responsible for the quality of declarations.

Unfortunately, current typing has several problems:
1. It's buggy:
  * Includes references to `AbortSignal` without including reference to the `dom` lib, so, it will not compile in a project that doesn't have it included (see the included test as an example).
 * Declares `referrer` on the body while `README` explicitly says we don't have that
 * Declares `getAll` on Headers while we don't have such thing in code anymore
 * etc.
2. It's overblown and irrelevant. It declares a lot of things that have nothing to do with the functionality of this library, like `RequestContext` or `ResponseType`, and exports things that we actually don't export.

So, this PR proposes completely different, lightweight, and minimal approach - it uses types, declared in DOM library included with TypeScript, and extends or shrinks them whether possible and required. The resulted declaration is smaller, more compatible, and easier to maintain and read.

As it dedicated to version 3.0 it's OK to have some breaking changes and it's includes breaking functionality for users migrating from `@types/node-fetch`, most notable:
- `response.json()` method returns `unknown`, not `any`. It's a good practice and even a reference [example of a use for `unknown` type](https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type) when it was introduced:
_This is useful for APIs that want to signal “this can be any value, so you must perform some type of checking before you use it”. This forces users to safely introspect returned values._
- we now export from declaration only things that we actually export from the library (i.e. types now like in case if the library were written in TypeScript) - `Request`, `Response`, `Headers`, `FetchError`, `AbortError` and `fetch` (default export) itself. Nothing else (the previous version was product in itself, exporting even `RequestContext` enum that we don't have in code).

To guarantee our typing declaration consistency I also included small automated test for GitHub action.

PS: It doesn't include any changes to JS code, so, Travis is just flacky.


test/types/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+		"noEmit": true, /* Do not emit outputs. */
+		/* Strict Type-Checking Options */
+		"strict": true /* Enable all strict type-checking options. */,
+		"noImplicitAny": true, /* Raise error on expressions and declarations with an implied 'any' type. */
+		"strictNullChecks": true, /* Enable strict null checks. */
+		"strictFunctionTypes": true, /* Enable strict checking of function types. */
+		"strictBindCallApply": true, /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
+		"strictPropertyInitialization": true, /* Enable strict checking of property initialization in classes. */
+		"noImplicitThis": true, /* Raise error on 'this' expressions with an implied 'any' type. */
+		"alwaysStrict": true /* Parse in strict mode and emit "use strict" for each source file. */,
```
LinusU, line 16: `"strict": true` turns in the options that are then listed below, it can just be:

```suggestion
		"strict": true /* Enable all strict type-checking options. */,
```

test/types/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+	"compilerOptions": {
+		"target": "es2019" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
+		"lib": [
+			"es2019.object",
+		],
```
LinusU, line 6: Why are we specially adding `es2019.object`? 🤔 That adds `Object.fromEntries` which I don't see that we are using anywhere...

Since we are supporting Node.js 10.x, maybe we should set this to:

```suggestion
		"target": "es2018",
		"lib": ["es2018"],
```

ref: https://stackoverflow.com/a/57607634/148072

tinovyatkin:
@LinusU I didn't want to introduce a new testing framework for types and resort to this kind of "functional testing" by compilation a file via the latest version of TypeScript on the GitHub Actions.
There are certain areas to improve even further here, but at the moment I feel it's a good step forward and would prefer to wait for first real-world problem feedback from users to understand how to better deal with potential problems should they arrive.

jstewmon:
Referencing dom lib causes all sorts of problems for node projects. This has been an ongoing problem with superagent for a long time. Please reconsider this approach.

tinovyatkin:
@jstewmon I should agree that lack of module isolation (by default) in TypeScript may cause background injection of `lib.dom.d.ts` and I agree too that's it's not a desired side-effect for may pure Node project - thank you for this insight!

I will bring inside all our types (at better review it's not that much we need to bring in). 
@Richienb could you do the same for `fetch-blob` that currently reexports `Blob` from `lib.dom`, so, I will import `Blob` from the `fetch-blob` then?

tinovyatkin:
@LinusU we probably need to revert https://github.com/node-fetch/fetch-blob/commit/bec5a3dceb340ccbf3130f0000807395460fcc17
Also `export =` declaration requires synthetic default imports enabled, is there any particular reason why you avoiding `export default Blob`?

LinusU:
> we probably need to revert node-fetch/fetch-blob@bec5a3d

Hmm, yeah, we can probably copy the classes from `dom.d.ts` right off though 🤔 

> Also export = declaration requires synthetic default imports enabled, is there any particular reason why you avoiding export default Blob?

The types should always reflect what's in the source file, and since this is the source:

```js
module.exports = Blob
```

The correct typing is `export = Blob`. I think that you'll have problem importing it in some setups if the types don't match the source. Here is some guidelines from DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

index.d.ts
```diff
@@ -1,212 +1,145 @@
...
-	// Node-fetch extensions
-	agent?: Agent | ((parsedUrl: URL) => Agent); // =null http.Agent instance, allows custom proxy, certificate etc.
-	compress?: boolean; // =true support gzip/deflate content encoding. false to disable
-	follow?: number; // =20 maximum redirect count. 0 to not follow redirect
-	size?: number; // =0 maximum response body size in bytes. 0 to disable
-	timeout?: number; // =0 req/res timeout in ms, it resets on redirect. 0 to disable (OS limit applies)
-	highWaterMark?: number; // =16384 the maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.
+import { Agent } from 'http';
+import { URL, URLSearchParams } from 'url'
+import { AbortSignal } from 'abort-controller'
```
Richienb, line 5: `abort-controller` is not a dependency so it won't work out of the box. Also, node-fetch is supposed to be a lightweight implementation and adding this might be weighing it down.

tinovyatkin: @Richienb actually `abort-controller` is a `devDependency` of this project for quite a long time. What will be your proposal alternative to referring `AbortSignal` here - bring it from `lib.dom` or copy the whole definition inside (it's quite big)?

Richienb: Is it possible to import from `lib.dom` using ES6 imports?

tinovyatkin: Nope, `lib.dom` is a library, not a module. And you probably should double-check the @jstewmon comment above about downside of just referring `lib.dom` in a triple-slash comment.

tinovyatkin:
> `@types` can probably live in the root directory.

Due to current `tsd` bug I didn't manage to move `index.test-d.ts` into a folder different from `index.d.ts` and leaving a test file in root directory looks weird for me...

Richienb:
@tinovyatkin In that case, we can put `index.d.ts` in the root directory and `index.test-d.ts` in the `tests` directory.

tinovyatkin:
> @tinovyatkin In that case, we can put `index.d.ts` in the root directory and `index.test-d.ts` in the `tests` directory.

Sorry for my English, but it's what I wanted to say. It's impossible to have `index.d.ts` and `index.test-d.ts` in different folders using the current version of `tsd`. So, that's why I moved the whole stuff into a separate folder.

xxczaki:
@tinovyatkin It looks like a lot of bigger projects (those, who have several `*.d.ts` files) store them with tests inside folders. However, projects that maintain only one type definitions file (`index.d.ts`) store them with tests in the root directory (https://github.com/sindresorhus/execa).

So, IMO it's ok for us to place `index.d.ts` and `index.test-d.ts` in the root directory :smile:

tinovyatkin:
I'm OK to move them whatever place the team will decide, but what's particularly wrong with `@types` folder?

xxczaki:
For me, it doesn't matter where the type tests will be placed ;)

xxczaki:
Are we ready to merge or not yet?

tinovyatkin:
For me this is pretty much done and ready for merge. The fact that v3.0 comes bundled with types is documented on README at TypeScript section.

xxczaki:
Let's change the name to `types`, as it seems like most projects call it like so.

tinovyatkin:
That's a great Sunday discussion on one symbol in the folder name at project internal structure 😄   
Let's make some fire 🔥 

@xxczaki I don't have statistics (would be nice to have) how many projects have it at `types` vs `@types`, but for me, a good vote for `@types` is default support in `vscode-icons` - most popular Visual Studio Code files icon theme (and in [Top 10 of all VSCode extensions](https://marketplace.visualstudio.com/search?target=VSCode&category=All%20categories&sortBy=Installs)). I do believe that `@types` will be better user experience event after @Richienb concerns of an extra symbol:

Screenshot 2020-05-24 at 10 43 26

Screenshot 2020-05-24 at 10 44 05

xxczaki:
@tinovyatkin I mean, if one symbol improves DX just a bit and does not really affect any users, we can go for it :laughing:

NotMoni:
Is also support @types

NotMoni:
I have one question 

In the actions file don’t we need to specify the node version

NotMoni:
We should change the name CI in the actions to something more specific b/c we already have a CI yml

tinovyatkin:
@NotMoni 
> In the actions file don’t we need to specify the node version

No, GitHub Actions have LTS version installed by default: https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners

> We should change the name CI in the actions to something more specific b/c we already have a CI yml

No, we shouldn't unless we want more badges in README. GitHub Action will sum up action by this name into one badge status, so, if somebody will push a broken typing into `master` - we will see it if we keep the name as CI and will not if we will change it to something else.

NotMoni:
Alright 🦄

PR #811, opened 2020-05-21
Add bug report issue template

NotMoni:
Since we are moving over to v3.

We should bring some consistency and help **and** make it easier for people to report bugs. Most of the time, ppl don't report them because they can't find a way or they are too lazy to type out every detail. So with this template, we can reduce work for the bug catchers and it will be easier for us to find the bug.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,5 +1,5 @@
 ---
-name: Bug report
+name: 🐞 Bug report
 about: Create a report to help us improve node-fetch
 title: "[BUG] "
 labels: bug
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,32 +6,38 @@ labels: bug
 ---
 
 **Describe the bug**
+
 <!--- A clear and concise description of what the bug is. -->
 
 **To Reproduce**
+
 Steps to reproduce the behavior:
+
 1. 
 2. 
 3. 
 4. 
 
 **Expected behavior**
+
 <!--- If you're suggesting a change/improvement, tell us how it should work -->
 <!--- If fetch is behaving incorrectly, please include a link to the spec or documentation that supports that your expected behavior is correct. -->
 
 **Screenshots**
+
 <!--- If applicable, add screenshots to help explain your problem. -->
 
 **Your Environment**
 
 <!--- Include as many relevant details about the environment you experienced the bug in -->
 
- | software         | version
+| software         | version
 | ---------------- | -------
-| node-fetch          |
+| node-fetch       |
 | node             |
 | npm              |
 | Operating System |
 
 **Additional context**
+
 <!--- Add any other context about the problem here. -->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,15 +1,13 @@
 ---
 name: 🐞 Bug report
 about: Create a report to help us improve node-fetch
-title: "[BUG] "
+title: "Bug: "
 labels: bug
 ---
 
-**Describe the bug**
+<!--- Provide a clear and concise description of what the bug is. -->
 
-<!--- A clear and concise description of what the bug is. -->
-
-**To Reproduce**
+**Reproduction**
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Your Environment**
+
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+
+ | software         | version
+| ---------------- | -------
+| node-fetch          |
```
MitMaro, line 31: Nitpick: But it seems there was some effort in aligning these, should these also be aligned?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. 
+2. 
+3. 
+4. 
+
+**Expected behavior**
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
```
MitMaro, line 19: Should there be a separate issue template for changes/improvements?

.github/ISSUE_TEMPLATE/bug_report.md
NotMoni: It won't actually affect it, when its rendered I think.

.github/ISSUE_TEMPLATE/bug_report.md
NotMoni: check out #816

NotMoni:
> The directions under "Expected behavior" are html comments, but other directions are not. Should other directions also be comments?

I don't see any HTML comments.  🤔 ?

.github/ISSUE_TEMPLATE/bug_report.md
MitMaro: Yup, the render would be the same either way. :)

MitMaro:
> > The directions under "Expected behavior" are html comments, but other directions are not. Should other directions also be comments?
> 
> I don't see any HTML comments.  ?

For example the line `<!--- If you're suggesting a change/improvement, tell us how it should work -->`

Since Markdown doesn't really provide a comment mechanism, they are generally used as a workaround for adding lines to a markdown file that should not be rendered when it is rendered as HTML.

Having the instructions as comments would mean that if the user doesn't remove the instructions, they would not be shown in the final output.

ESLint is the one project that I can think of that does something similar: https://raw.githubusercontent.com/eslint/eslint/3a5fbb3d634be688615950c0a5fa8aead6ff08b5/.github/ISSUE_TEMPLATE/BUG_REPORT.md

NotMoni:
Yep, I see that. It the file is .md so it shouldn't matter? and should it be good?

NotMoni:
What do you suggest as an alternative?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+<!--- If fetch is behaving incorrectly, please include a link to the spec or documentation that supports that your expected behavior is correct. -->
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
```
MitMaro, line 23:
```suggestion
<!--- If applicable, add screenshots to help explain your problem. -->
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
+---
+name: 🐞 Bug report
+about: Create a report to help us improve node-fetch
+title: "[BUG] "
+labels: bug
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
```
MitMaro, line 9:
```suggestion
<!--- A clear and concise description of what the bug is. -->
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+
+ | software         | version
+| ---------------- | -------
+| node-fetch          |
+| node             |
+| npm              |
+| Operating System |
+
+**Additional context**
+Add any other context about the problem here.
```
MitMaro, line 37:
```suggestion
<!--- Add any other context about the problem here. -->
```

Issue #527, opened 2018-09-28
Send node version in user-agent


dmail:
When a server is requested by a browser it can parse browser version from `user-agent`.
`node-fetch` could send node version in `user-agent` header as well.

Something like that maybe : `node/8.6.0 node-fetch/1.0 (+https://github.com/bitinn/node-fetch)`

dmail:
I see, my use case involves a local server so I'm not worried about 'leaking' my nodejs version to myself.
I understand your point about sensitive information.

I set the header myself in order to know the nodejs version pinging my server, that's fine like that.

jimmywarting:
btw, chrome will start removing OS build number.

> This will prevent abuses of that information such as exploit targeting and fingerprinting. It'll also bring Chrome closer in line with [RFC 7231 section 5.5.3.](https://tools.ietf.org/html/rfc7231#section-5.5.3)

https://www.chromestatus.com/features/4558585463832576

Issue #522, opened 2018-09-19
User Agent Version


cristobal:
The header `User-Agent` if it's not set, it will default to:

```js
headers.set('User-Agent', 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)')
```

Should it not default to use the current module version?  e.g.:

```js
const pkg = require('package.json')
...
headers.set('User-Agent', `node-fetch/${pkg.version} (+https://github.com/bitinn/node-fetch)`)
```

Or some automated injection of the version?

cristobal:
Ok I see, seems the 🚢 has sailed then.
Most browser clients standardise on the current browser version, would have expected the node-fetch would have done the same.

We now ship our own `user-agent`  header which works fine, but when diving into source we would have expected that it used the version of the node module. That's why i reported it.

Anyways thanks for the fast reply, just closing this issue then 👍

jimmywarting:
Guess it would maybe be interesting to have the version reflect the version in package.json. 
But then again what are the con/pros?

Con:
Servers need to make special handling to figure out that 
`node-fetch/1.0 (+https://github.com/bitinn/node-fetch)` is the same module as
`node-fetch/2.0 (+https://github.com/bitinn/node-fetch)`

If the exact version is sent it leaves information away that you might not want to share.
I have vittnes Wordpress attacks that sends out `X-Powered-By` version and hackers would take on them cuz they know a security flaw in version `x < y`

If they know that NodeJS or node-fetch had a bug in memory handling they could respond with something that would be bad. 

like the [Unicode of Death](https://www.cultofmac.com/324043/apple-confirms-imessage-bug-is-crashing-iphones/) for iMessage for iPhones :P

Pro: 
Can't think of something...

Fingerprinting? Dynamic User Agent

jimmywarting:
I don't care much...

Issue #503, opened 2018-08-15
Default user-agent string still version 1.0?


jkantr:
https://github.com/bitinn/node-fetch/blob/1d4ab5a0de0df991f92f1f30efd5f4815bbcdb71/src/request.js#L186

Should `node-fetch/1.0` be changed to `node-fetch/2.0`? Or has that ship sailed?

PR #818, opened 2020-05-22
Fix default user agent string

tinovyatkin:
Fixes default user agent string by reading package name, version, and homepage from `package.json` instead of using hard-coded strings.


src/request.js
```diff
@@ -234,7 +238,7 @@ export function getNodeRequestOptions(request) {
 
 	// HTTP-network-or-cache fetch step 2.11
 	if (!headers.has('User-Agent')) {
-		headers.set('User-Agent', 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)');
+		headers.set('User-Agent', `${packageName}/${version} [Node ${process.version}] (+${homepage})`);
```
NotMoni, line 241: since `${version}`  is just a number. I think you should do `v${version}`

jimmywarting:
for ref: #503 #522 #527

jimmywarting:
https://github.com/WICG/ua-client-hints

tinovyatkin:
> https://github.com/WICG/ua-client-hints

Do you mean to put Node version into `()`, not `[]`?

jimmywarting:
I mean, we have had this discussion before.

> User agents identify themselves to servers as part of each HTTP request via the User-Agent header. This header's value has grown in both length and complexity over the years; a complicated dance between server-side sniffing to provide the right experience for the right devices on the one hand, and client-side spoofing in order to bypass incorrect or inconvenient sniffing on the other

UA is complex, long and sometimes unnecessary
It seems like they want to take action in reducing User-Agent Granularity, So I'm fine with keeping it at v1 and expose as little as possible too.

tinovyatkin:
I believe that at least wrong link must be fixed. And version should be either correct or removed completely, makes no sense to _spoof_ it.
If some consumer of the library is worried about information leak then there is a documented way to prevent it.

src/request.js
tinovyatkin: 'v' prefix is traditionally omitted in userAgent strings.

LinusU:
I personally would prefer keeping the hard coded string, or even removing version information. Is there any upside to sending this information out with every request?

Also, what was the rationale for adding the Node.js version to the string?

Chrome is also moving towards hard-coding the value, and never changing it: https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/-2JIRNMWJ7s/yHe4tQNLCgAJ

Maybe I'm missing some upside to this though?

jimmywarting:
what do you think about just using `user-agent: node-fetch` or `user-agent: node`? (trying to imagine how many bytes could be saved worldwide by 14M weekly node-fetch downloads if we shaved of 50 bytes from our current user-agent string)

> Is there any upside to sending this information out with every request?

read this: https://github.com/node-fetch/node-fetch/issues/226

bitinn:
My vote is on a simple string.

My original design was to emulate how google crawler bot would do it:

https://support.google.com/webmasters/answer/1061943?hl=en

I am happy with `node-fetch/3.0` alone, or just `node-fetch`.

Richienb:
As per consensus, it seems that we should just leave the UA alone as a static string.

xxczaki:
Yes, let's change it to `node-fetch`.

caub:
Please, could we change back the user-agent to something less long by default? (including the link to github is a good idea, but that's wasting nabwidth for most stuff and people that don't override this default)

`node-fetch` or `node-fetch/2.x` or similar

jimmywarting:
the current user agent is just `node-fetch`...

PR #819, opened 2020-05-22
Add Brotli support in Request

tinovyatkin:
Brotli is supported in Node.JS core since version 10.16 and it was handled in `node-fetch` under the guard on receiving, however it wasn't set on request. This PR fixes and documents that support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,7 +430,7 @@ If no values are set, the following request headers will be sent automatically:
 
 | Header              | Value                                                    |
 | ------------------- | -------------------------------------------------------- |
-| `Accept-Encoding`   | `gzip,deflate` _(when `options.compress === true`)_      |
+| `Accept-Encoding`   | `gzip,deflate,br` _(when `options.compress === true`)_   |
 | `Accept`            | `*/*`                                                    |
 | `Connection`        | `close` _(when no `options.agent` is present)_           |
 | `Content-Length`    | _(automatically calculated, if possible)_                |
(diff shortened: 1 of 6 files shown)


jimmywarting:
Should br be conditionally added as well when setting Accept-Encoding?

tinovyatkin:
I've adjusted minimal Node requirement in `package.json` to `10.16` - and it's available since this version, so, no guards required.

tinovyatkin:
> Would it be difficult to backport support to earlier versions of Node.js 10?

For what reason?

Richienb:
> For what reason?

It is likely that most consumers are using an older version of Node.js 10 and don't bother updating when new minor versions come out.

tinovyatkin:
> It is likely that most consumers are using an older version of Node.js 10 and don't bother updating when new minor versions come out.

Well... If somebody doesn't care to update a minor version of Node then this theoretical person for sure will not upgrade to the major version of `node-fetch` that brings several other breaking changes. We do a major version, we can set new requirements, and Node 10 is coming out of LTS pretty soon. C'mon, there is old version of `node-fetch` at NPM available FOREVER if you want to stick something _classic_, like Node 10.10.

tinovyatkin:
Just for a reference: Node 10 become a LTS at `10.14`. So, we are discussing just 2 minor versions up.

damianaiamad:
A word of caution for turning on brotli compression by default: it can be **very** slow.

In NodeJS, the default compression level is the highest and slowest (11). It's really slow—see below.

I know it can be tuned and it is fine for build-time compression, but for many people doing a routine upgrade of node-fetch (along with all the other packages) may be in for a rude surprise when their production server slows to a crawl due to brotli. 

Perhaps worth a mention in docs?

For example 12MB json file (real world data):

BROTLI
Level 1 Compression 88.7% Time 36ms
Level 2 Compression 89.3% Time 72ms
Level 3 Compression 89.6% Time 74ms
Level 4 Compression 91.1% Time 112ms
Level 5 Compression 92.0% Time 180ms
Level 6 Compression 92.1% Time 222ms
Level 7 Compression 92.2% Time 302ms
Level 8 Compression 92.3% Time 397ms
Level 9 Compression 92.4% Time 545ms
Level 10 Compression 93.1% Time 6441ms
Level 11 Compression 93.6% Time 33956ms **<<<< NodeJS default - 60x slower than gzip - ouch!**

GZIP
Level 1 Compression 87.4% Time 75ms
Level 2 Compression 87.8% Time 79ms
Level 3 Compression 88.2% Time 91ms
Level 4 Compression 89.1% Time 115ms
Level 5 Compression 89.4% Time 135ms
Level 6 Compression 89.9% Time 169ms
Level 7 Compression 90.0% Time 215ms
Level 8 Compression 90.2% Time 505ms
Level 9 Compression 90.3% Time 557ms

bitinn:
We should allow brotil but use a decent compression level by default (if nodejs latest LTS still has this issue).

damianaiamad:
I guess I wanted to point out that turning on brotli by default does not result in an automatic win. It could be disastrous if the server side is not configured appropriately (which is not something node-fetch can control—it's up to the web server). So perhaps brotli should be optional? Or a helpful warning in the docs that brotli can seriously degrade performance if not tuned correctly?

tinovyatkin:
@damianaiamad You can always set `Accept-Enconding` on your own and it will be respected

PR #821, opened 2020-05-23
Move `isRedirect` function to a seperate file

tinovyatkin:
- Fixes `isRedirect` hackish "export" as property on default export to be a real ES6 named export
- Replace array creation and iteration on each call with static O(1) Set
- Reuse `isRedirect` at `Response` class

### DIFF
--- a/src/utils/is-redirect.js
+++ b/src/utils/is-redirect.js
@@ -1,0 +1,12 @@
+// https://onlinewritingtraining.com.au/plural-of-status/
+const REDIRECT_STATUS = new Set([301, 302, 303, 307, 308]);
+
+/**
+ * Redirect code matching
+ *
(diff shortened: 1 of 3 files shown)


src/utils/is-redirect.js
```diff
@@ -0,0 +1,11 @@
+const redirectStatus = new Set([301, 302, 303, 307, 308]);
+
+/**
+ * Redirect code matching
+ *
+ * @param {number} code - Status code
+ * @return {boolean}
+ */
+export function isRedirect(code) {
+	return redirectStatus.has(code);
```
jimmywarting, line 10: code golfing :D 

```js
export const isRedirect = s => ~~'11100011'[s-301]
```

can someone do better?

PR #825, opened 2020-05-23
Update pull request template

NotMoni:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,15 +6,15 @@ Please read and follow these instructions before creating and submitting a pull 
 - If you are reporting a bug, adding failing units tests can be a good idea.
 -->
 
-**What is the purpose of this pull request? (put an "X" next to item)**
+**What is the purpose of this pull request?**
 
 [ ] Documentation update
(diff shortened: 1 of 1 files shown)


jimmywarting:
After seeing this in use i feel like the checkboxes are abused a bit

first of all, they should really be radio buttons instead. but github don't support it...

checkboxes are more like task that needs to be completed.

Skärmavbild 2020-06-12 kl  21 51 03

And i don't feel like this is the right way to use them... would it be possible to create multiple PR templates instead with more narrow focus on what kind of PR it is?

jimmywarting:
feel like we should remove them, think we can do better

jimmywarting:
if it's a bug fix PR then it would be useful with a template that have something like
```
- [ ] created a test case
- [ ] done code coverage
- [ ] Ready to be reviewed

fixes #000
```

and maybe some html comments or something

NotMoni:
Will do @jimmywarting

NotMoni:
@jimmywarting @tinovyatkin Should we make multiple TEMPLATE or fit them all in one ?

NotMoni:
I would like to do all in one.

jimmywarting:
> I would like to do all in one.

ok

PR #829, opened 2020-05-23
Follow xo linter rules more strictly

xxczaki:
**What is the purpose of this pull request?**

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain:

This PR removes some of the ignored xo linter rules so that the code is more readable.

**What changes did you make? (provide an overview)**

Removed ignored linter rules and linted the files (including docs).

**Which issue (if any) does this pull request address?**

N/A

**Is there anything you'd like reviewers to know?**

This PR might look rather large. However, most of the changes are actually small.

PR #831, opened 2020-05-24
Remove `timeout` option

Richienb:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain:

Removed the `timeout` option and updated the migration guide.

Fixes https://github.com/node-fetch/node-fetch/issues/523


bitinn:
For some reasons my email reply to issues/PRs are no longer getting through, so add it again:

- This PR has the same level of risk as removing browser export, please make sure at least 3 reviewers have approved it before merging.

- The timeout signal modular appear to be the main alternative after deprecating timeout, I suggest moving it to node-fetch org and adding extra npm owners to ensure maintenance.

- The documentation shouldn't just say we deprecate it because of  "non-standard", while this is a part of our reasoning, the main reason is something like "Abort Signal offers more finegrained control of timeout, and is standardized in Fetch spec. For convenience, we provide a timeout signal as a simple replacement for users upgrading from v2".

tinovyatkin:
Should #810 be merged first, to keep changes to typing in this PR?

NotMoni:
Again, here Coveralls Coverage decreased by `-0.01%`

xxczaki:
@NotMoni This is not a big deal here, once we merge #829 the coverage will raise, as it also includes the updated ignore comments.

We might want to try and bring the coverage back to a full 100% soon.

xxczaki:
@tinovyatkin Yes, I think so.

NotMoni:
> @NotMoni This is not a big deal here, once we merge #829 the coverage will raise, as it also includes the updated ignore comments.
> 
> We might want to try and bring the coverage back to a full 100% soon in a separate PR.

Yea, coverage to 100% will be good to do.

tinovyatkin:
This will be a very painful breaking change for may libraries that depends on `node-fetch`, as it will require to introduce `abort-controller` as dependence. Maybe we should just deprecate `timeout` in 3.0 and remove it in the next major?

xxczaki:
@tinovyatkin It would, but only for people, who actually want to use this feature. So the others would benefit, as the "core" node-fetch package would not include it.

tinovyatkin:
@xxczaki I mean, I believe that it will be too hard to implement vs a desire to just stick with `2.x` version. For example, `gaxios` by Google is using `node-fetch` and even has `abort-controller` as a dependency, but only to declare types:
https://github.com/googleapis/gaxios/search?q=abort-controller&unscoped_q=abort-controller

So, now they will need to decide whether to drop `timeout` parameter on their side and release new major of `gaxios` or implement their timeout via AbortSignal or just pin the dependency on the latest 2.x version of `node-fetch` until _they will have the bandwidth to deal with it_. 

`gaxios` engine requirement is already 10.x, so, actually that will be the only breaking change for them...

NotMoni:
pls fix merge conflict with TS Typing before we can merge

cc/ @Richienb

gr2m:
> This will be a very painful breaking change for may libraries that depends on node-fetch, as it will require to introduce abort-controller as a dependency

For what it's worth, I use `node-fetch` heavily in the @octokit libraries, and I allow to pass a `signal` option to each request, which can be set using an AbortController instance. I do not need to add `abort-controller` as a dependency, it's up to the integrator to add that if they need to.

PR #833, opened 2020-05-24
chore: Rewrite tests to AVA

### LINKED ISSUE #832: chore: Rewrite tests to AVA
## Is your feature request related to a problem? Please describe.

The current tests are good, however, I don't really like the fact, that Mocha depends on global variables (`it`, `describe`). Furthermore, Mocha runs tests serially, which is much slower, especially when there are a lot of tests involved.

AVA runs each test in a separate process and therefore takes advantage of multi-core CPUs.

**As we maintain over 200 tests, this will likely require a significant amount of work, but I think it's something that we should consider, as it will make maintaining the project easier.**

## Describe the solution you'd like

- Migrate from Mocha & Chai to AVA.
- During the process, try and split the huge, 2000 LOC+ `test/main.js` file into smaller pieces.

## Describe alternatives you've considered

N/A

## Additional context

- http://zpalexander.com/migrating-from-mocha-to-ava/

xxczaki:
Fixes #832

---

## TODO:

- [ ] Main (assigned: @curtisjohnson1)
	- [ ] Split into multiple files
- [x] External encoding
- [x] Headers
- [x] Request
- [x] Response
- [x] Form Data

## Information

The original tests are in the `test` directory, while the new ones are in the `test-new` directory. We will rename it to `test` when done.


package.json
```diff
@@ -20,8 +20,8 @@
     },
     "scripts": {
         "build": "rollup -c",
-        "test": "node --experimental-modules node_modules/c8/bin/c8 --reporter=html --reporter=lcov --reporter=text --check-coverage node --experimental-modules node_modules/mocha/bin/mocha",
-        "coverage": "c8 report --reporter=text-lcov | coveralls",
+        "test": "c8 --report=none --no-clean ava",
+        "coverage": "c8 report | coveralls",
```
tinovyatkin, line 24: It will not run on Node 10.x and 12.x without flags

xxczaki: That's what I thought :confounded: I just copied the configuration from [AVA](https://github.com/avajs/ava).

xxczaki: @tinovyatkin I updated the configuration to match the old, working behavior.

tinovyatkin:
@xxczaki https://github.com/avajs/ava/blob/master/docs/recipes/es-modules.md

xxczaki:
@tinovyatkin Thanks, to support Node.js 10 & 12 I used the `esm` package.

tinovyatkin:
> @tinovyatkin Thanks, to support Node.js 10 & 12 I used the [`esm` package](https://github.com/avajs/ava/blob/master/docs/recipes/es-modules.md#using-the-esm-package).

Hmm. Remember that `c8` doesn't work well with `esm`. I don't really understand why to introduce a dependency and test transpired code on 10 and 12 where we currently are testing original sources?

xxczaki:
The ESM package is used only for AVA. I can add the experimental modules flag instead, but I'm not sure if that works on Node.js 10.

tinovyatkin:
> The ESM package is used only for AVA. I can add the experimental modules flag instead, but I'm not sure if that works on Node.js 10.

Flag works, see our current tests

xxczaki:
Ok, I will replace it with a flag later today ;)

xxczaki:
Update: all the "basic" test files are now rewritten. The only 1 left is the massive `main.js`.

@curtisjohnson1 Let me know if you need any help ;)

test-new/external-encoding.js
```diff
@@ -0,0 +1,26 @@
...
+	const response = await fetch('data:,Hello%20World!');
+	const text = await response.text();
+
+	t.is(response.status, 200);
+	t.is(response.headers.get('Content-Type'), 'text/plain');
+	t.is(text, 'Hello World!');
+});
+
+test('should reject invalid data uri', async t => {
+	const error = await t.throwsAsync(fetch('data:@@@@'));
```
sindresorhus, line 23:
```suggestion
	await t.throwsAsync(fetch('data:@@@@'), {message: /invalid URL/});
```

test-new/utils/server.js
```diff
@@ -0,0 +1,390 @@
...
+	stop(cb) {
+		this.server.close(cb);
+	}
+
+	mockResponse(responseHandler) {
+		this.server.nextResponseHandler = responseHandler;
+		return `http://${this.hostname}:${this.port}/mocked`;
+	}
+
+	router(request, response) {
```
sindresorhus, line 35: Something to think about for the future, in Got, we ended up with a separate server for each test, which is really nice as you don't need a router and nothing can tests cannot affect each other. We also use a helper and AVA test macros to make it really convenient: https://github.com/sindresorhus/got/blob/a337dfdfc4151593044c8b729245308081a071be/test/http.ts#L9-L15

xxczaki: That's really interesting, will definitely look into it. Thanks :smile:

curtisjohnson1:
> Update: all the "basic" test files are now rewritten. The only 1 left is the massive `main.js`.
> 
> @curtisjohnson1 Let me know if you need any help ;)

Hey @xxczaki - sorry for the delay. I'm looking to get this finished over the weekend 👍

xxczaki:
@curtisjohnson1 Great, no need to hurry :smile:

curtisjohnson1:
hi @xxczaki - sorry for the delay in this. I've been working on porting the test over for the past 2 days in my spare time and I've got 25% through the main.js file so far. My tests are currently here [https://github.com/curtisjohnson1/node-fetch/blob/ava/test-new/main.js](url). Do you want me to either point my branch to this, otherwise, I can use this branch to push commits to?

xxczaki:
@curtisjohnson1 Thanks for letting me know about the progress! Doesn't matter, you can use your branch or this one - just make sure to point it when you finish.

PR #834, opened 2020-05-24
revamp Headers module

### LINKED ISSUE #783: doesn't correctly handle multiple `content-type` values
I found a peculiar issue of incompatibility of this library comparing to native browser fetch version. Affected **version 2.6.0**

Steps to reproduce:

**1.) Create simple node server and run it**
```js
const http = require('http');

const requestListener = function (req, res) {
  res.setHeader('Cookie', ['one=1', 'two=2']);
  res.setHeader('Content-Type', ['text/plain', 'application/json']);
  res.writeHead(200);
  res.end('Hello, World!');
}

const server = http.createServer(requestListener);
server.listen(8080);
```

**2.) Fetch the server**

`In browser`
```js
fetch('http://localhost:8080').then(r => r.headers).then(h => h.get('Content-Type')).then(console.dir);
// text/plain, application/json
```

`node-fetch`
```js
fetch('http://localhost:8080').then(r => r.headers).then(h => h.get('Content-Type')).then(console.dir);
// text/plain
```
As you can see, `node-fetch` doesn't correctly handle multiple headers with the same name.

tinovyatkin:

- [X] Other, please explain:

This is an almost complete rewrite of the `Headers` module (so, diff is looking crazy - it's better just to read new code completely). 

The current implementation of `Headers` is somewhere painful to read (particularly iterator implementation). It's nowhere close to being lightweight too with inventing some wheels and avoiding modern language features.

The only change to test file is the replacement of originally broken test for `createHeadersLenient` - it was testing for falseness properties on `Headers` object that was always false.
Change for other files is renaming of internal methods `createHeadersLenient` (to `fromRawHeaders`) and `exportNodeCompatibleHeaders` (to build-in symbol).

The approach used in this refactoring is to take advantage of the fact, that there is _almost_ completely identical by shape built-in object in Node >= 10.0 - `URLSearchParams` (I do believe it was copied from `Headers` when introduced browser-side, so, now we can reuse it back). We extend here `URLSearchParams` attending minor differences in functionality (case insensitivity, parameters validation, and keys iteration) via lightweight Proxy, and major functionality differences (`get` and values iteration) as overloaded methods.

I'd moved `exportNodeCompatibleHeaders` export into Node.JS built-in symbol, so, it also now serves for _everyone secret debug powerhouses_ - `console.log(response.headers)`

I've replaced `createHeadersLenient` and creating `Headers` from `rawHeaders`, that fixes #783.

With the use of Proxy, constructor return value and generator functions in the source is also a fun weekend reading for everybody who loves modern JavaScript 😄 

This PR purposely doesn't touch any tests to showcase that module is in equal functionality, however, there are some points to discuss:

1. Headers now, by side effect, exposing `getAll` method (as it exists on `URLSearchParams`) that was removed from client-side spec due to security concern and then removed from `node-fetch` too, then `raw` was introduced when we realized than on the server-side, indeed, we need to be able to get all values. 
It's easy to hide this method completely (or make non-enumerable), or leave it undocumented (all other modules as Request and Response are exposing some non-spec methods and we have `raw` here anyway) or to document it. I do believe it's an extremely useful server-side and better fit for Cookies than `raw`. So @node-fetch/core probably should decide here.
See: https://github.com/whatwg/fetch/issues/973#issuecomment-560136041

2. There is no current test for `new Headers(null)` and it passes in both new and original versions of Headers. However, it throws at Chrome. Should behavior be adjusted and test added?

~3. There is a subtle difference in how headers are returned concatenated - `node-fetch` is using `, ` to join them while browsers use just coma. Should it be adjusted?~

4. Proxies are fast in v8: https://v8.dev/blog/optimizing-proxies


tinovyatkin:
Screenshot 2020-05-25 at 03 04 04
:tada: 💪

xxczaki:
**Before merging this, I think we should discuss the last points from the PR description**

Here are my thoughts:

1. I think we can expose `getAll`, while this removed from the specification, I agree that it fits Node.js server environment better than `raw` and certainly helps in a lot of use cases.
2. What's the behavior in other browsers? Is it defined in the Fetch standard?
3. If it's something, that is described in the Fetch standard, we should adjust it.

cc @node-fetch/core @tinovyatkin

tinovyatkin:
My personal thoughts:

1. I'm 👍🏻 for exposing and documenting `getAll`. We get it for free with this implementation and from the hard-voice comment https://github.com/whatwg/fetch/issues/973#issuecomment-559821655 I do believe that WHATWG doesn't care about `fetch` standard interoperability, so, an extension that makes sense for runtime environment will not hurt. I would be inviting @kentonv or whatever person that @cloudflare seems appropriate to join @node-fetch/core and participate in such decisions.

2. Both, Firefox and Chrome throws on `new Headers(null)`

3. I don't remember how this topic get here, our functionality here is equal to browsers.

tinovyatkin:
@xxczaki  I believe with 5 approvals and no opposing comments this PR can be merged and nuances 1 and 2 attended/discussed in separate PRs.

PR #841, opened 2020-05-26
fix: improve TypeScript types

SomaticIT:

- [ ] Documentation update
- [X] Bug fix
- [ ] New feature
- [ ] Other, please explain:

- Export all types and interfaces
- Use `class` instead of `interface`+`var` where it's possible
- Use a minimal type for `AbortSignal` to avoid the need to import 'abort-controller' module
- Use method declaration instead of lambda declaration
- Avoid namespace creation that is not needed

- Impossible to construct `RequestInit` or `ResponseInit` object correctly:
```typescript
const init: RequestInit = { /* ... */ };
```

- Impossible to build a TypeScript project with no `abort-controller` dependency
- Not compatible with `v3.0.0-beta.5` types

- check if `isRedirect()` is useable in this way.

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -6,7 +6,7 @@ import {Agent} from 'http';
 import * as Blob from 'fetch-blob';
 
 export type AbortSignal = {
-	readonly aborted: boolean;
+    readonly aborted: boolean;
 
     addEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { passive?: boolean; once?: boolean; }): void;
     removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -12,7 +12,7 @@ export type AbortSignal = {
     removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
 };
 
-export type HeadersInit = Headers | string[][] | Record<string, string>;
+export type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string[]>;
 
 /**
  * This Fetch API interface allows you to perform various actions on HTTP request and response headers.
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -5,14 +5,14 @@
 import {Agent} from 'http';
 import * as Blob from 'fetch-blob';
 
-export type AbortSignal = {
-    readonly aborted: boolean;
+type AbortSignal = {
+	readonly aborted: boolean;
 
-    addEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { passive?: boolean; once?: boolean; }): void;
-    removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
+	addEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { passive?: boolean; once?: boolean; }): void;
+	removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
 };
 
-export type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string[]>;
+type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string[]>;
 
 /**
  * This Fetch API interface allows you to perform various actions on HTTP request and response headers.
@@ -21,7 +21,7 @@ export type HeadersInit = Headers | Record<string, string> | Iterable<readonly [
  * You can add to this using methods like append() (see Examples.)
  * In all methods of this interface, header names are matched by case-insensitive byte sequence.
  * */
-export class Headers {
+declare class Headers {
 	constructor(init?: HeadersInit);
 
 	append(name: string, value: string): void;
@@ -52,7 +52,7 @@ export class Headers {
 	raw(): Record<string, string[]>;
 }
 
-export interface RequestInit {
+interface RequestInit {
 	/**
 	 * A BodyInit object or null to set request's body.
 	 */
@@ -86,19 +86,22 @@ export interface RequestInit {
 	highWaterMark?: number;
 }
 
-export interface ResponseInit {
+interface ResponseInit {
 	headers?: HeadersInit;
 	status?: number;
 	statusText?: string;
 }
 
-export type BodyInit =
+type BodyInit =
 	| Blob
 	| Buffer
 	| URLSearchParams
 	| NodeJS.ReadableStream
 	| string;
-export interface Body {
+type BodyType = { [K in keyof Body]: Body[K] };
+declare class Body {
+	constructor(body?: BodyInit, opts?: { size?: number });
+
 	readonly body: NodeJS.ReadableStream | null;
 	readonly bodyUsed: boolean;
 	readonly size: number;
@@ -109,14 +112,10 @@ export interface Body {
 	json(): Promise<unknown>;
 	text(): Promise<string>;
 }
-declare var Body: {
-	prototype: Body;
-	new(body?: BodyInit, opts?: {size?: number}): Body;
-}
 
-export type RequestRedirect = 'error' | 'follow' | 'manual';
-export type RequestInfo = string | Body;
-export class Request extends Body {
+type RequestRedirect = 'error' | 'follow' | 'manual';
+type RequestInfo = string | Body;
+declare class Request extends Body {
 	constructor(input: RequestInfo, init?: RequestInit);
 
 	/**
@@ -142,7 +141,7 @@ export class Request extends Body {
 	clone(): Request;
 }
 
-export class Response extends Body {
+declare class Response extends Body {
 	constructor(body?: BodyInit | null, init?: ResponseInit);
 
 	readonly headers: Headers;
@@ -154,9 +153,9 @@ export class Response extends Body {
 	clone(): Response;
 }
 
-export class FetchError extends Error {
+declare class FetchError extends Error {
 	constructor(message: string, type: string, systemError?: object);
-	
+
 	name: 'FetchError';
 	[Symbol.toStringTag]: 'FetchError';
 	type: string;
@@ -164,12 +163,39 @@ export class FetchError extends Error {
 	errno?: string;
 }
 
-export class AbortError extends Error {
+declare class AbortError extends Error {
 	type: string;
 	name: 'AbortError';
 	[Symbol.toStringTag]: 'AbortError';
 }
 
-export function isRedirect(code: number): boolean;
 
-export default function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
+declare function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
+declare class fetch {
+	static default: typeof fetch;
+}
+declare namespace fetch {
+	export function isRedirect(code: number): boolean;
+
+	export {
+		HeadersInit,
+		Headers,
+
+		RequestInit,
+		RequestRedirect,
+		RequestInfo,
+		Request,
+
+		BodyInit,
+
+		ResponseInit,
+		Response,
+
+		FetchError,
+		AbortError
+	};
+
+	export interface Body extends BodyType { }
+}
+
+export = fetch;
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -12,7 +12,7 @@ type AbortSignal = {
 	removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
 };
 
-type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string[]>;
+type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<Iterable<string>>;
 
 /**
  * This Fetch API interface allows you to perform various actions on HTTP request and response headers.
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -3,10 +3,16 @@
 /* eslint-disable no-var, import/no-mutable-exports */
 
 import {Agent} from 'http';
-import {AbortSignal} from 'abort-controller';
-import Blob from 'fetch-blob';
+import * as Blob from 'fetch-blob';
 
-type HeadersInit = Headers | string[][] | Record<string, string>;
+export type AbortSignal = {
+	readonly aborted: boolean;
+
+    addEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { passive?: boolean; once?: boolean; }): void;
+    removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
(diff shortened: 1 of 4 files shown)


@types/index.d.ts
```diff
@@ -3,10 +3,16 @@
 /* eslint-disable no-var, import/no-mutable-exports */
 
 import {Agent} from 'http';
-import {AbortSignal} from 'abort-controller';
-import Blob from 'fetch-blob';
+import * as Blob from 'fetch-blob';
```
LinusU, line 6: Maybe the best thing to do is to import with `require`, as `fetch-blob` is a CommonJS library? 🤔 That should work without any user having to have synthetic imports or es module interop enabled...

```suggestion
import Blob = require('fetch-blob');
```

@types/index.d.ts
```diff
@@ -3,10 +3,16 @@
...
 
-type HeadersInit = Headers | string[][] | Record<string, string>;
+export type AbortSignal = {
+	readonly aborted: boolean;
```
LinusU, line 9: indentation seems off here

@types/index.d.ts
```diff
@@ -131,52 +139,37 @@ interface Request extends Body {
...
 	type: string;
 	name: 'AbortError';
 	[Symbol.toStringTag]: 'AbortError';
 }
 
-export {Headers, Request, Response, FetchError};
-export default fetch;
+export function isRedirect(code: number): boolean;
+
+export default function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
```
LinusU, line 175: Since the `main` field will point to `index.cjs`, and the `types` field (which is the corresponding field for the typings) will point to this file, I think that the correct thing here would be to use a CommonJS export:

```suggestion
declare function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
export = fetch;
```

You can read some more info it here: https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

@types/index.d.ts
SomaticIT: Both methods should work with no `esModuleInterop` nor synthetic imports.
```typescript
import * as Blob from 'fetch-blob';
import Blob = require('fetch-blob');
```

I prefer the `*` import personnally but I can change it if you prefer.

@types/index.d.ts
SomaticIT: Yep, let's fix this

@types/index.d.ts
SomaticIT: If you look at the built `index.cjs`, the resulting `exports` are the following:
```javascript
exports.default = fetch;
```

So the good way to type it is to use `export default`.
`export = fetch` is equivalent to `module.exports = fetch` which is not the case here...

@types/index.d.ts
SomaticIT: Fixed in last commit

@types/index.d.ts
SomaticIT: In the last commit, I forced `esModuleInterop` and `allowSyntheticDefaultImports` to `false` in tsd config to ensure that it works correctly and it seems to.

@types/index.d.ts
LinusU: ahh, cool, I was looking at the currently published package which uses `module.exports = fetch`, my bad!

SomaticIT:
Thanks all reviewers for your time!

@types/index.d.ts
```diff
@@ -3,10 +3,16 @@
...
 
-type HeadersInit = Headers | string[][] | Record<string, string>;
+export type AbortSignal = {
+    readonly aborted: boolean;
+
+    addEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { passive?: boolean; once?: boolean; }): void;
+    removeEventListener(type: "abort", listener: (this: AbortSignal, ev: Event) => any, options?: boolean | { capture?: boolean; }): void;
+};
+
+export type HeadersInit = Headers | string[][] | Record<string, string>;
```
tinovyatkin, line 15: If you want to export `HeadersInit` it must be described fully and correctly. Right now it doesn't cover all possible ways to construct `Headers`, particularly from `Iterables`. Look how it's described in JSDoc at `headers.js`:

```suggestion
export type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string>[];
```

tinovyatkin: But still I'm 👎  to export `HeadersInit`. What is use case for this?

```ts
const hi: HeadersInit = { /*...*/ }
const header = new Headers(hi)
```
?

That looks weird.

@types/index.d.ts
```diff
@@ -82,35 +86,39 @@ interface RequestInit {
...
+	text(): Promise<string>;
 }
 declare var Body: {
 	prototype: Body;
-	new (body?: BodyInit, opts?: {size?: number}): Body;
-};
+	new(body?: BodyInit, opts?: {size?: number}): Body;
+}
+
+export type RequestRedirect = 'error' | 'follow' | 'manual';
```
tinovyatkin, line 117: Why do we need to export this type?

@types/index.d.ts
```diff
@@ -82,35 +86,39 @@ interface RequestInit {
...
+export type RequestInfo = string | Body;
```
tinovyatkin, line 118: What is use case to export this?

@types/index.d.ts
```diff
@@ -82,35 +86,39 @@ interface RequestInit {
 	highWaterMark?: number;
 }
 
-interface ResponseInit {
+export interface ResponseInit {
 	headers?: HeadersInit;
 	status?: number;
 	statusText?: string;
```
tinovyatkin, line 92: Should we add `readonly` here?

SomaticIT:
> `export = fetch` while looks weird and requires synthetic default import solves this problem and so, results in better DX. Should we make that change?

I strongly disagree on this. A d.ts library should never require an option in tsc

@types/index.d.ts
SomaticIT: I agree

@types/index.d.ts
SomaticIT: Imagine a library which do this:
```typescript
function myCustomHttpGet(url: string, headers: HeadersInit): Promise<void>
```

It's important for this library to be able to access all types for the underlying library to avoid duplication and reduce maintenance.

@types/index.d.ts
SomaticIT: Same as previous

@types/index.d.ts
SomaticIT: Same as previous

SomaticIT:
> My major concern with current typings is the export description. I understand that `export default fetch` matches with our code, etc.
> However, it lacks tooling support (read VSCode IntelliSense) while the module is used as CommonJS (try with `test/commonjs/test-artifact.js`:
> Screenshot 2020-05-27 at 09 18 40
> 
> `export = fetch` while looks weird and requires synthetic default import solves this problem and so, results in better DX. Should we make that change?

image

It works on my vs code...

SomaticIT:
Important note:
The type of HeadersInit is incorrect in JSDoc:
```typescript
export type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string>[];
```

Indeed, the evaluation function looks at length in iterable items (line 79):
```javascript
if (pair.length !== 2) {
    throw new TypeError('Each header pair must be a name/value tuple');
}
```

It means that we have to replace `Iterable<string>[]` by `Iterable<string[]>`:
```typescript
export type HeadersInit = Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string[]>;
```

It also performs better TypeScript check:
```typescript
// Works with Iterable<string>[] but not with Iterable<string[]>
const headers = new Headers(['name', 'value]);
```

@types/index.d.ts
LinusU: @SomaticIT I just cloned the repo and made a clean build, and the output in the bundle is `module.exports = fetch`, thus I think that we should type this as `export = fetch`

@types/index.d.ts
tinovyatkin: Well, it makes sense to use the least demanding types - in the `dom.lib` most of the field declared as `readonly`

LinusU:
Just double checked how the module is exported from the built file, and it is exported with the following code:

```js
exports = module.exports = fetch;
```

Thus the correct typing for this module would be:

```ts
export = fetch;
```

ref: https://github.com/node-fetch/node-fetch/pull/841#discussion_r431241142

-----

~~If we type this as `export default function fetch` then you won't be able to correctly import this without `esModuleInterop`, since it will output:~~ Actually we are doing `fetch.default = fetch` so it will still work...

```js
var node_fetch_1 = require("node-fetch");
node_fetch_1["default"]('test');
```

-----

**without `esModuleInterop`**

Our declaration | Users import | Transpiled code
----- | ----- | -----
`export default function` | `import fetch from 'node-fetch` | `var node_fetch_1 = require("node-fetch");` <br> `node_fetch_1["default"]('test');`
`export default function` | `import fetch = require('node-fetch')` | Error: This expression is not callable.
`export = fetch` | `import fetch from 'node-fetch` | Error: This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
`export = fetch` | `import fetch = require('node-fetch')` | `var fetch = require("node-fetch");` <br> `fetch('test');`


**with `esModuleInterop`**

Our declaration | Users import | Transpiled code
----- | ----- | -----
`export default function` | `import fetch from 'node-fetch` | `var node_fetch_1 = __importDefault(require("node-fetch"));` <br> `node_fetch_1["default"]('test');`
`export default function` | `import fetch = require('node-fetch')` | Error: This expression is not callable.
`export = fetch` | `import fetch from 'node-fetch` | `var node_fetch_1 = __importDefault(require("node-fetch"));` <br> `node_fetch_1["default"]('test');`
`export = fetch` | `import fetch = require('node-fetch')` | `var fetch = require("node-fetch");` <br> `fetch('test');`

tinovyatkin:
> The type of HeadersInit is incorrect in JSDoc:

Please check the `test/headers.js` file for all possible ways to init Headers. Particularly badly named test `should accept headers as an iterable of tuples` where Headers are created from `Set<string>` and `.keys` iterator from Map.

xxczaki:
@tinovyatkin Slightly unrelated to this PR, but what would you rename the `should accept headers as an iterable of tuples` test to?

LinusU:
Argh, I have been looking way too long but cannot find it now. I know that I saw a comment from someone on the TypeScript team the other day explaining how to correctly type a library that exports `module.exports = foo; module.exports.default = foo` so that it can be imported with both CommonJS and Module style, without requiring `esModuleInterpo` in any of the cases. I really think that is the best approach here since we are really exporting both versions...

tinovyatkin:
> @tinovyatkin Slightly unrelated to this PR, but what would you rename the `should accept headers as an iterable of tuples` test to?

I'm not a native English speaking person, so, probably not the best source of such type of suggestions, but what this test actually tests is that `Headers` can be created from a linear iterator of string values if the result can be split into tuples evenly. I mean, it's not iterable of tuples, it's iterable of strings.

tinovyatkin:
@LinusU Do you mean this comment? https://github.com/Microsoft/TypeScript/issues/2242#issue-60179656

LinusU:
Hmmm, no unfortunately not, this was a more recent comment. And it had some text that was basically "If your library is really exporting both CommonJS and a default name, you can do this in your typings:"...

SomaticIT:
@LinusU Maybe this one:
https://github.com/microsoft/TypeScript/issues/10854#issuecomment-246183500

@types/index.d.ts
Richienb, line 175:
```suggestion
export = function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
```

LinusU:
@SomaticIT not that specific comment, but it seems to be the same trick 😍 awesome find!! 👏 

With that I have managed to test this out locally and I think that this is the best approach:

```typescript
type Fetch = (url: RequestInfo, init?: RequestInit) => Promise<Response>
declare const fetch: Fetch & { default: Fetch }
export = fetch
```

`esModuleInterop` | Users import | Transpiled code
----- | ----- | -----
`false` | `import fetch from 'node-fetch'` | `var node_fetch_1 = require("node-fetch");` <br> `node_fetch_1["default"]('test');`
`false` | `import fetch = require('node-fetch')` | `var fetch = require("node-fetch");` <br> `fetch('test');`
`true` | `import fetch from 'node-fetch'` | `var node_fetch_1 = __importDefault(require("node-fetch"));` <br> `node_fetch_1["default"]('test');`
`true` | `import fetch = require('node-fetch')` | `var fetch = require("node-fetch");` <br> `fetch('test');`

This is amazing, I think that there is a lot of libraries that are missing this in their types, where the library is actually exporting both `module.exports` and `module.exports.default` 🙌

@types/index.d.ts
LinusU, line 175:
```suggestion
declare type Fetch = (url: RequestInfo, init?: RequestInit) => Promise<Response>;
declare const fetch: Fetch & { default: Fetch };
export = fetch;
```

SomaticIT:
Dear reviewers,

I think I found a very good solution that allows to do `export = `, `export default` and `export types` in the same time: 
```typescript
declare function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>; // function is mergeable
declare class fetch {
	static default: typeof fetch; // default needs to be a value to use import fetch from...
}
declare namespace fetch { // namespace allows import * and exporting types
	// export types
}

export = fetch;
```

This solutions allows any ways of importing in TypeScript and creates compatible transpilation.
It also allows to have classes like Request and Response to be exported in the best way.

Eg:
```typescript
import fetch, {Request} from "node-fetch";
asserts.equals(fetch.Request, Request);
fetch.isRedirect = (code: number) => true;

import * as fetch from "node-fetch";
asserts.equals(fetch.default.Request, fetch.Request);
fetch.isRedirect = (code: number) => true;

import fetch from require("node-fetch");
asserts.equals(fetch.default.Request, fetch.Request);
fetch.isRedirect = (code: number) => true;
```

With this model, I think that we match perfectly the `index.cjs` module exports.

What do you think?

@types/index.d.ts
SomaticIT: I found another way that allows to export types along with `fetch`. See my next comment.

SomaticIT: I found a very good way to do this while keeping default export and types exports. Please have a look to the next commit and comment.

src/headers.js
```diff
@@ -24,7 +24,7 @@ function validateValue(value) {
 }
 
 /**
- * @typedef {Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string>[]} HeadersInit
+ * @typedef {Headers | Record<string, string> | Iterable<readonly [string, string]> | Iterable<string[]>} HeadersInit
```
tinovyatkin: Please check this: https://github.com/node-fetch/node-fetch/pull/841#issuecomment-634764277. It's iterable of string, not string array.

SomaticIT:
> > The type of HeadersInit is incorrect in JSDoc:
> 
> Please check the `test/headers.js` file for all possible ways to init Headers. Particularly badly named test `should accept headers as an iterable of tuples` where Headers are created from `Set<string>` and `.keys` iterator from Map.

You're right, I fixed it in my last commit with `Iterable<Iterable<string>>`, the only problem is that it allows `new Headers(["test", "test"])` to pass type checks but will not pass runtime check.

But it's the real type for the implementation.

src/headers.js
SomaticIT: I just fixed it with `Iterable<Iterable<string>>` that is a more robust and compatible method.

PR #842, opened 2020-05-27
chore: remove code duplication in custom errors

tinovyatkin:
In the current implementation `AbortError` and `FetchError` pretty much repeating the same code (plus some magic strings). This PR moves common functionality into a base class and removes code duplication.

It also removes `isAbortError` and uses `instanceof FetchBaseError` as those errors are our internal classes and always instantiating from it.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -9,7 +9,7 @@ import Stream, {PassThrough} from 'stream';
 import {types} from 'util';
 
 import Blob from 'fetch-blob';
-import FetchError from './errors/fetch-error.js';
+import {FetchError} from './errors/fetch-error.js';
 import {isBlob, isURLSearchParameters, isAbortError} from './utils/is.js';
 
 const INTERNALS = Symbol('Body internals');
--- a/src/body.js
+++ b/src/body.js
@@ -10,7 +10,8 @@ import {types} from 'util';
 
 import Blob from 'fetch-blob';
 import {FetchError} from './errors/fetch-error.js';
-import {isBlob, isURLSearchParameters, isAbortError} from './utils/is.js';
+import {FetchBaseError} from './errors/base.js';
+import {isBlob, isURLSearchParameters} from './utils/is.js';
 
 const INTERNALS = Symbol('Body internals');
 
@@ -60,7 +61,7 @@ export default class Body {
 
 		if (body instanceof Stream) {
 			body.on('error', err => {
-				const error = isAbortError(err) ?
+				const error = err instanceof FetchBaseError ?
 					err :
 					new FetchError(`Invalid response body while trying to fetch ${this.url}: ${err.message}`, 'system', err);
 				this[INTERNALS].error = error;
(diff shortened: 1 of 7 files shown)


tinovyatkin:
@Richienb The abort error is just two lines... Import statement will adds up the same amount of code as implementation....

Richienb:
> @Richienb The abort error is just two lines... Import statement will adds up the same amount of code as implementation....

I still think they should be in separate files since it would make it easier for future maintenance.

LinusU:
I personally don't think that there is a problem with keeping the errors in the same file...

Although, have you checked that the stack traces actually looks identical after making this change? I might be wrong here, but I think that when you do `Error.captureStackTrace` in the parent class, referencing `this.constructor` might be referencing the parent constructor, which will leave the leaf class in the stack trace? Not too sure about this though, it's a shame that it's not more straight forward to `extends Error` 😅

src/errors/errors.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+class FetchBaseError extends Error {
+	constructor(message, type) {
+		super(message);
+		// Hide custom error implementation details from end-users
+		Error.captureStackTrace(this, this.constructor);
+
+		this.type = type;
```
MitMaro, line 9: I've generally done the following to provide a read-only property in errors:

```suggestion
		Object.defineProperty(this, 'type', {
			configurable: false,
			enumerable: true,
			value: type,
			writable: false,
		});
```

tinovyatkin: Well. Do you hate inheritance in that case, don't you?

tinovyatkin:
@Richienb 
> I still think they should be in separate files since it would make it easier for future maintenance.

So, for best maintenance in the future you expect to have 3 files right: `errors/base.js`, `errros/abort.js`, and `errors/fetch.js` - right?  A lot of boilerplate, harder imports, and slower load doesn't matter?

src/errors/errors.js
MitMaro: That is a rather rude, borderline personal attack to a friendly suggestion.

:man_shrugging: 

It's a property passed into the super's constructor and generally would not be changed after it is initially set. Making it read-only seemed, to me, to be a logical conclusion. It's also a pattern I've seen a few times in other projects for a `type` property in error classes.

Richienb:
Using seperate files will allow them to be self-documenting so the future maintainers don't have to manually check the exports to find the use.

tinovyatkin:
@Richienb Adjusted as per your suggestion.

src/errors/errors.js
tinovyatkin: Sorry, @MitMaro I didn't mean to talk about personality,  my bad, I mean just _class inheritance_. But by making the `type` non-configurable property set in base class how we suppose to override it in a child class?

tinovyatkin:
@LinusU 
> Although, have you checked that the stack traces actually looks identical after making this change? I might be wrong here, but I think that when you do `Error.captureStackTrace` in the parent class, referencing `this.constructor` might be referencing the parent constructor, which will leave the leaf class in the stack trace?

Not, it will be referencing the actual class constructor. This is how `super` works.

PR #845, opened 2020-05-28
Drop custom Promises and refactor to `async` functions

### LINKED ISSUE #844: Bug: We already broke custom Promises
Now [famous PR](https://github.com/node-fetch/node-fetch/pull/829/files#diff-f0c01727e5628483090afb5bc8c36be6R84) also introduced `async` functions into `body.js` code, practically rendering useless custom Promises.
That should be finished or rolled back.

tinovyatkin:
Fixes #844 and drops custom Promises.

This PR didn't complete the refactoring of `index.js` into `async` function (as it way more complicated and time-consuming task), but it can be done by iterations later on as non-breaking changes, now when we drop support for custom promises.

No changes to tests, except for the removal of testing for custom promises.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -601,8 +601,9 @@ describe('node-fetch', () => {
 		return fetch(url).then(res => {
 			expect(res.status).to.equal(200);
 			expect(res.ok).to.be.true;
-			return expect(res.text()).to.eventually.be.rejectedWith(Error)
-				.and.have.property('message').includes('Premature close');
+			return res.text().then(text => {
+				expect(text).to.equal('foo');
+			});
 		});
 	});
 
--- a/test/main.js
+++ b/test/main.js
@@ -10,7 +10,6 @@ import chai from 'chai';
 import chaiPromised from 'chai-as-promised';
 import chaiIterator from 'chai-iterator';
 import chaiString from 'chai-string';
-import then from 'promise';
 import resumer from 'resumer';
 import FormData from 'form-data';
 import stringToArrayBuffer from 'string-to-arraybuffer';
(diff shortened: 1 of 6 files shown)


test/main.js
```diff
@@ -601,8 +601,9 @@ describe('node-fetch', () => {
 		return fetch(url).then(res => {
 			expect(res.status).to.equal(200);
 			expect(res.ok).to.be.true;
-			return expect(res.text()).to.eventually.be.rejectedWith(Error)
-				.and.have.property('message').includes('Premature close');
```
LinusU, line 605: Hmmm, this should still reject with an error, right? Seems like the new code is missing to check for errors at some point? 🤔

tinovyatkin: It's Draft PR, not ready for PR yet.
But yes, I'm struggling to understand expected behavior with a partial body. Do we need to put a partial body somewhere or just discard it?

LinusU: Hehe, sorry, I was a bit eager 😁 

I think that we can just discard it, since the request have errored 🤔

test/utils/server.js
```diff
@@ -305,9 +305,9 @@ export default class TestServer {
 		if (p === '/error/premature') {
 			res.writeHead(200, {'content-length': 50});
 			res.write('foo');
-			setTimeout(() => {
-				res.destroy();
-			}, 100);
+			setImmediate(() => {
```
jimmywarting, line 308: Hmm, haven't seen setImmediate being used for a long time anywhere in the wild. Thought it was more a IE thing. after all it seems to be "[none standard](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate)"

guess what most node developer use is `process.nextTick()` but this is the test file so it don't even mather so much.

PR #846, opened 2020-05-28
fix: Correct Node versions were not installed on test matrix

tinovyatkin:
Fixes a bug in GitHub Action workflow resulted in all tests were running only on Node 12.x
Also now tests on minimum node version from `package.json` `engines` instead of any `10.x`

~The test for `network-error partial response` was failing on Node 10.16 and set conditionally.
I've expected it to be fixed by #845, but it seems that there is some underlying problem in Node <12.10 (I do believe it's https://github.com/nodejs/node/pull/29376). This probably may be attended in a separate PR or leave as is (the functionality difference is that on Node < 12.10 `fetch` is resolving with a partial response)~

Fixes test for a partial response that was failing on Node < 12.10 due to https://github.com/nodejs/node/pull/29376 by backporting changes from that PR.


.github/workflows/ci.yml
```diff
@@ -2,46 +2,64 @@ name: CI
...
+      - "package.json"
+      - ".github/workflows/ci.yml"
 
 jobs:
   test:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        node: [14.x, 12.x, 10.x]
+        node: [">=14.3", "^12.17", engines]
```
NotMoni, line 17: why are we hardcoding node versions?

tinovyatkin: For some reason, they are not in sync with the actual latest releases. I particularly want at least 14.3 for #843. It's not much _harder_ than generic 14 or 12 actually as it still the semver range.

NotMoni: Ahhhh ok 👍. I approve the merge

tinovyatkin: @NotMoni great, hit the `Approval` button then 😉

NotMoni: Oh I total forgot. I’m on mobile sorry

SomaticIT:
Hello, this issue is blocking the release `3.0.0-beta.7`...
Maybe I can help moving forward on this issue?

test/request.js
```diff
@@ -1,6 +1,6 @@
 import stream from 'stream';
 import http from 'http';
-import polyfill from 'abortcontroller-polyfill/dist/abortcontroller.js';
+import AbortController from 'abort-controller';
```
Richienb, line 3: If we are going to swap out `abortcontroller-polyfill` for `abort-controller`, the same should be done for `test/main.js`

tinovyatkin: We are checking both, `abortcontroller-polyfill` and `abort-controller` at `test/main.js`. Here we need just one, and `abort-controller` is easier to import from ESM, so, I swap it here.

PR #848, opened 2020-05-29
More permissive `isStream` logic

geelen:
To support other stream APIs e.g. `web-streams-ponyfill`

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain: tweak to logic

For other stream APIs e.g. `web-streams-ponyfill`, they're not detected as streams so they get serialised to `[object Object]`. This change relaxes the `isStream` constraint so they get treated as `stream` objects, and since they imlpement the same methods they just work.

More than happy to discuss alternative checks here, but `instanceof` seems like too strict a check, when there's multiple Stream packages that are compatible.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -15,8 +15,9 @@ import {isBlob, isURLSearchParameters, isAbortError} from './utils/is.js';
 const INTERNALS = Symbol('Body internals');
 
 function isStream(body) {
-	return body instanceof Stream || (body && body.constructor && /Stream/.exec(body.constructor.name))
+	return body instanceof Stream || (body && body.constructor && /Stream/.exec(body.constructor.name));
 }
+
 /**
  * Body mixin
  *
--- a/src/body.js
+++ b/src/body.js
@@ -15,7 +15,7 @@ import {isBlob, isURLSearchParameters, isAbortError} from './utils/is.js';
 const INTERNALS = Symbol('Body internals');
 
 function isStream(body) {
-	return body instanceof Stream || (body.constructor && /Stream/.exec(body.constructor.name))
+	return body instanceof Stream || (body && body.constructor && /Stream/.exec(body.constructor.name))
 }
 /**
  * Body mixin
--- a/src/body.js
+++ b/src/body.js
@@ -14,6 +14,9 @@ import {isBlob, isURLSearchParameters, isAbortError} from './utils/is.js';
 
 const INTERNALS = Symbol('Body internals');
 
+function isStream(body) {
+	return body instanceof Stream || (body.constructor && /Stream/.exec(body.constructor.name))
+}
 /**
  * Body mixin
  *
@@ -43,7 +46,7 @@ export default class Body {
 		} else if (ArrayBuffer.isView(body)) {
 			// Body is ArrayBufferView
 			body = Buffer.from(body.buffer, body.byteOffset, body.byteLength);
-		} else if (body instanceof Stream) {
+		} else if (isStream(body)) {
 			// Body is stream
 		} else {
 			// None of the above
@@ -58,7 +61,7 @@ export default class Body {
 		};
 		this.size = size;
 
-		if (body instanceof Stream) {
+		if (isStream(body)) {
 			body.on('error', err => {
 				const error = isAbortError(err) ?
 					err :
@@ -177,7 +180,7 @@ async function consumeBody(data) {
 	}
 
 	/* c8 ignore next 3 */
-	if (!(body instanceof Stream)) {
+	if (!isStream(body)) {
 		return Buffer.alloc(0);
 	}
 
@@ -236,7 +239,7 @@ export const clone = (instance, highWaterMark) => {
 
 	// Check that body is a stream and not form-data object
 	// note: we can't clone the form-data object without having it as a dependency
-	if ((body instanceof Stream) && (typeof body.getBoundary !== 'function')) {
+	if (isStream(body) && (typeof body.getBoundary !== 'function')) {
 		// Tee instance body
 		p1 = new PassThrough({highWaterMark});
 		p2 = new PassThrough({highWaterMark});
@@ -292,7 +295,7 @@ export const extractContentType = body => {
 	}
 
 	// Body is stream - can't really do much about this
-	if (body instanceof Stream) {
+	if (isStream(body)) {
 		return null;
 	}
 


tinovyatkin:
@geelen if you are absolutely sure that web-streams are the functional equivalent of Node.JS `stream.Readable` then it's easy to spoof any library:

```js
const webStreams = require("web-streams-ponyfill");
const stream = require("streams");
const readable = new webStreams.ReadableStream();
const almostStream = new Proxy(readable, {
  getPrototypeOf() {
    return stream.Readable;
  }
});
```

geelen:
Yeah, I admit the check might be too lax, but is there another way to avoid `const stream = require('stream'); ...; if (body instanceof stream)`?

Feels overly restrictive, given that `node-fetch` is a way to bring a web standard `fetch` API to node, it feels like it ought to be compatible with a project doing the same for the web standard `Stream` api?

geelen:
> @geelen if you are absolutely sure that web-streams are the functional equivalent of Node.JS `stream.Readable` then it's easy to spoof any library that is using `instanceof`:
> 
> ```js
> const webStreams = require("web-streams-ponyfill");
> const stream = require("streams");
> const readable = new webStreams.ReadableStream();
> const almostStream = new Proxy(readable, {
>   getPrototypeOf() {
>     return stream.Readable;
>   }
> });
> ```

Ooh, that might be enough for me to get around this issue. Thanks for the quick response!

jimmywarting:
My idea have kind of been:

- if it is a node stream, use it as is...
- otherwise check if body has `body[Symbol.asyncIterator]` 
  - if it had a async iterator convert body to a node stream using `body = stream.Readable.from(body[Symbol.asyncIterator]())`

This would allow whatwg stream to pass through, but also super simplified iterators that are not bound to any environment or stream, meaning you could pass in the form-data encoder (#603) directly or just a function that yields Uint8Arrays
```js
async function * stream() {
  yield new Uint8Array([97, 98, 99]) // abc
}

await new Response( stream() ).text() // abc
``` 

this would make it super relaxed, but would at least allow whatwg streams.

tinovyatkin:
@jimmywarting with `Readable.from` it's very easy to construct a real Node stream from any AsyncIterator, but I believe this feature doesn't belong to the inside of our library. We may, however, document this as tips in README.

jimmywarting:
@tinovyatkin  yea... asyncIterator should not really be part of a supported body - even doe i think it should be part of the spec - I even recommended it to the spec but they where not so hype on adding more magic functionality to the api. But I think we should at least allow whatwg stream to slip through - meaning that we should convert whatwg streams to node streams ourself somehow.

tinovyatkin:
What we probably need to handle well inside this library is object mode readable for body... 🤔
But it's out of spec, so, need to decide first.

tinovyatkin:
at a certain point, the team should reinforce position whether `node-fetch` is a sort of testing/polyfilling library (like JSDOM) or just a way for a developer to use their knowledge of client-side fetch API to make a request server-side.
Because WHATWG is doing things _strongly_ for browser-side API and following it without compromises will blow out this library into a slow behemoth.

tinovyatkin:
@geelen Sorry, it was just an example written here in a comment :-)
You should return `Stream.Readable.prototype` actually:

```js
const Stream = require("stream");
class HybridReadableStream {
  constructor(...args) {
    return new Proxy(this, {
      getPrototypeOf() {
        return Stream.Readable.prototype;
      }
    });
  }
}

const x = new HybridReadableStream();
console.log(x instanceof Stream.Readable); // true
```

geelen:
Ha, that makes more sense. Sadly, `ReadableStream` has no `on` method which means this bit breaks:

```js
	if (body instanceof Stream) {
		body.on('error', function (err) {
			const error = err.name === 'AbortError' ? err : new FetchError(`Invalid response body while trying to fetch ${_this.url}: ${err.message}`, 'system', err);
			_this[INTERNALS].error = error;
		});
	}
```

(my original hack version only changed the first `body instanceof Stream` check, so this PR would actually not have worked as written)

Expanding that proxy worked fine for my needs tho:

```ts
function HybridReadableStream(...args: any[]) {
  const readableStream = new ReadableStream(...args)
  return new Proxy(readableStream, {
    getPrototypeOf() {
      return Stream.Readable.prototype
    },
    get(target, prop, receiver) {
      if (prop === 'on') return () => {}
      return Reflect.get(target, prop, receiver)
    },
  })
}
```

Thanks heaps for your help!

tinovyatkin:
@jimmywarting yeah... That duck typing... We don't use `AsyncIterator` on the request body today, but may want tomorrow, so, it should be a reason to limit ourselves.
And, again, we should decide whether we do `instanceof`, functionality present test (`.on`, `.pipe`, etc) or `[Symbol.toStringTag]` tests - from my point of view combining any of them together steals all sense from doing so - and we, for example, currently test almost all functions on `URLSearchParams` (_except the one that the library is actually using_ - `.toString`) and `[Symbol.toStringTag]` - so, no abstract class can be passed. 
I mean, it should be well-motivated decision to support an abstract class instead of built-in objects/classes and example above when proposed `web-streams-ponyfill` doesn't have `.on` methods showcase that pretty well...

tinovyatkin:
By the end, there is a **`node`** in the `node-fetch` name - it is supposed to be Node-native first.

geelen:
Yeah, I think this is the crux of the matter:

> at a certain point, the team should reinforce position whether node-fetch is a sort of testing/polyfilling library (like JSDOM) or just a way for a developer to reuse their knowledge of client-side fetch API to make a request server-side.

I'm trying to make web-native code run inside NodeJS, using the _exact_ APIs that work in a browser, for the https://fab.dev project. But my [current implementation](https://github.com/fab-spec/fab/blob/master/packages/sandbox-node-vm/src/index.ts) is just an interim, the _real_ implementation is going to be using a `V8::Isolate` sandbox, where all those globally-available APIs (like `fetch`, `ReadableStream`) need to be constructed to cross the boundary safely, so the user's code will never be invoking `node-fetch` directly, and the problem "goes away" (i.e. becomes my problem instead).

All that is to say, this feels like only a temporary problem, where people are trying to run browser code and Node code together. Once something new shifts that, you'll go back to only Node code needing `node-fetch`, is my thinking...

jimmywarting:
> By the end, there is a node in the node-fetch name - it is supposed to be Node-native first.

think the **node** part have mostly stood for "A light-weight module that brings window.fetch to Node.js" in the scene of "isn't built or menat to support browser, react native or deno". it means we should try to keep i light-weight... bringing in curtain libraries like web-streams-ponyfill would be huge, but supporting them isn't.

PR #853, opened 2020-05-29
Fix Data URI handling and drop all URL analysis RegExps

tinovyatkin:
This PR adds a breaking test for existing dataUri implementation and reimplements it without using RegExp and also drops them in Request, following browsers logic on this case.

Only changes to tests are breaking case (from Wikipedia) for current implementation and adjusting of error messages as they are coming from native URL parser now (unsupported protocol message is copied from Chrome `fetch`)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,7 @@ import http from 'http';
 import https from 'https';
 import zlib from 'zlib';
 import Stream, {PassThrough, pipeline as pump} from 'stream';
-import dataURIToBuffer from 'data-uri-to-buffer';
+import dataUriToBuffer from 'data-uri-to-buffer';
 
 import {writeToStream, getTotalBytes} from './body.js';
 import Response from './response.js';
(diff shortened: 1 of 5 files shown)


tinovyatkin:
@jimmywarting well, install size is not smaller. But spec compliance is way better, as seen in Wikipedia example (on the mimeType part). I believe maintenance is more future-proof too...

jimmywarting:
I'm not 100% convinced yet, just b/c it has more downloads doesn't really mean it's more popular or faster, often with the case of such packages they get backed up by other bigger modules like jsDom to make them seem more popular, like how babel used left-pad for example. 

we haven't received any complains yet of some incompatible data url, so why should we change something that is already working?

in cases like this I would want to see some benchmarking too to compare both packages.

I'm also keen to help data-uri-to-buffer with PR if it's doing something wrong. (just created an [issue](https://github.com/TooTallNate/node-data-uri-to-buffer/issues/14) wondering if he still needs bufferFrom and offered to help)

Think there are better things you can put your effort into, if you like i can assign some issues to you

tinovyatkin:
This PR is to fix another bug (see the breaking test), dependency can stay as it, they are little different in mime types handling, but as I understand it doesn’t matter here.

jimmywarting:
#311 is an old issue worth bringing up, @TimothyGu first recommended that we used data-uri-to-buffer but later suggested that we also would use data-urls but the pr got closed by @Richienb who used native node functions (#659).

tinovyatkin:
yeah, my first intention was to use native Buffer.from, but data uri can be pretty complex...

jimmywarting:
well, #659 should never have been accepted. but that is gone now

src/index.js
```diff
@@ -22,36 +22,32 @@ import {isRedirect} from './utils/is-redirect.js';
...
-		throw new FetchError(`[${request.method}] ${request.url} invalid URL`, 'system');
-	}
-
-	// Wrap http.request into fetch
 	return new Promise((resolve, reject) => {
 		// Build request object
 		const request = new Request(url, options_);
 		const options = getNodeRequestOptions(request);
+		if (!supportedSchemas.has(options.protocol)) {
+			throw new TypeError(`node-fetch cannot load ${url}. URL scheme "${options.protocol.replace(/:$/, '')}" is not supported.`);
```
Richienb, line 40: Shouldn't this be a `FetchError`?

tinovyatkin: All URL checking and unsupported protocol errors earlier were TypeErrors (like in browsers). Should we change?

Richienb: If this is a limitation of node-fetch then I think it should be a `FetchError`. Otherwise, a `TypeError` will do.

tinovyatkin: No, it’s not node-fetch specific - original error is from Chrome in this case and says “Fetch API cannot download...” - I’ve just replaced fetch API with node-fetch

test/main.js
```diff
@@ -2133,7 +2133,6 @@ describe('node-fetch', () => {
 		});
 
 		it('should accept data uri 2', async () => {
-			// this is from wikipedia: https://en.wikipedia.org/wiki/Data_URI_scheme
```
tinovyatkin: @Richienb why do you think this documentation comment is irrelevant? It's the source reference.

Richienb: A link to Wikipedia about data uris is not necessary since it might as well go in the readme.

tinovyatkin: It's not only about data URIs, this specific URI for the test was copied from the article, so, any future maintainer will have no doubts that the actual URI is correct.

PR #855, opened 2020-06-01
Add `httpOptions` option

c-kruse:

- [ ] Documentation update
- [ ] Bug fix
- [X] New feature
- [ ] Other, please explain:

Added extraHTTPOptions option that gets passed through to the http(s).request options here https://nodejs.org/docs/latest-v12.x/api/http.html#http_http_request_options_callback.

It looks like there was recently a similar discussion around this for TLS Options, which can be configured on the agent (undocumented). The impetus for this PR is the ability to pass the `insecureHTTPParser` option, which as far as I can tell can't be set on the agent.

https://github.com/node-fetch/node-fetch/issues/724

Here's some background reading: https://github.com/nodejs/node/issues/27711
Sounds like it's a fairly common issue with the Incapsula CDN.

**Why not just use the Agent?**
```js
async function main() {
    const agent = new https.Agent({
        insecureHTTPParser: true // not a valid option in https://nodejs.org/api/http.html#http_new_agent_options
    });
    const uri = 'https://cams.jacksonhole.com/webcam/SouthHoback.jpg?1590589817';
    const data = await fetch(uri, {
        agent
    });
    console.log(data)
}

main().catch(console.error);
```

Running this on node 12.x LTS will get you this:
```
➜  node test.js
FetchError: request to https://www.dezeen.com/robots.txt failed, reason: Parse Error: Invalid header value char
    at ClientRequest.<anonymous> (/path/to/project/node_modules/node-fetch/lib/index.js:1455:11)
    at ClientRequest.emit (events.js:315:20)
    at TLSSocket.socketOnData (_http_client.js:476:9)
    at TLSSocket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:295:12)
    at readableAddChunk (_stream_readable.js:271:9)
    at TLSSocket.Readable.push (_stream_readable.js:212:10)
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:186:23) {
  type: 'system',
  errno: 'HPE_INVALID_HEADER_TOKEN',
  code: 'HPE_INVALID_HEADER_TOKEN'
}
```

We need the `insecureHTTPParser` option on the request to override this on a per-call basis. https://nodejs.org/api/http.html#http_http_request_options_callback

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,7 +446,8 @@ The default values are shown after each option key.
     compress: true,         // support gzip/deflate content encoding. false to disable
     size: 0,                // maximum response body size in bytes. 0 to disable
     agent: null,            // http(s).Agent instance or function that returns an instance (see below)
-    highWaterMark: 16384    // the maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.
+	highWaterMark: 16384,    // the maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.
+	extraHTTPOptions: {}	// additional options to include in http(s).request options that cannot be configured by the http(s).Agent override.
 }
 ```
 
@@ -536,6 +537,13 @@ const fetch = require('node-fetch');
 	return res.clone().buffer();
 })();
 ```
+
+
+#### Extra HTTP Options
+
+Extra options to pass to http.request.
+See [`http.request`](https://nodejs.org/api/http.html#http_http_request_url_options_callback) for more information.
+
 
 <a id="class-request"></a>
 
--- a/README.md
+++ b/README.md
@@ -539,9 +539,9 @@ const fetch = require('node-fetch');
 ```
 
 
-#### Extra HTTP Options
+#### HTTP Options
 
-Extra options to pass to http.request.
+Additional options to pass to http.request.
 See [`http.request`](https://nodejs.org/api/http.html#http_http_request_url_options_callback) for more information.
 
 
--- a/README.md
+++ b/README.md
@@ -447,7 +447,7 @@ The default values are shown after each option key.
     size: 0,                // maximum response body size in bytes. 0 to disable
     agent: null,            // http(s).Agent instance or function that returns an instance (see below)
 	highWaterMark: 16384,    // the maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.
-	extraHTTPOptions: {}	// additional options to include in http(s).request options that cannot be configured by the http(s).Agent override.
+	httpOptions: {}	// additional options to include in http(s).request options that cannot be configured by the http(s).Agent override.
 }
 ```
 
--- a/src/request.js
+++ b/src/request.js
@@ -100,7 +100,7 @@ export default class Request extends Body {
 		this.counter = init.counter || input.counter || 0;
 		this.agent = init.agent || input.agent;
 		this.highWaterMark = init.highWaterMark || input.highWaterMark || 16384;
-		this.extraHTTPOptions = init.extraHTTPOptions || input.extraHTTPOptions || {};
+		this.httpOptions = init.httpOptions || input.httpOptions || {};
 	}
 
 	get method() {
@@ -204,7 +204,7 @@ export const getNodeRequestOptions = request => {
 
 	// Manually spread the URL object instead of spread syntax
 	const requestOptions = {
-		...request.extraHTTPOptions,
+		...request.httpOptions,
 		path: parsedURL.pathname + search,
 		pathname: parsedURL.pathname,
 		hostname: parsedURL.hostname,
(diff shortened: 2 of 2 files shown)


c-kruse:
> I think we can just call the parameter `httpOptions`.

Agreed!

README.md
```diff
@@ -446,7 +446,8 @@ The default values are shown after each option key.
     compress: true,         // support gzip/deflate content encoding. false to disable
     size: 0,                // maximum response body size in bytes. 0 to disable
     agent: null,            // http(s).Agent instance or function that returns an instance (see below)
-    highWaterMark: 16384    // the maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.
+	highWaterMark: 16384,    // the maximum number of bytes to store in the internal buffer before ceasing to read from the underlying resource.
+	httpOptions: {}	// additional options to include in http(s).request options that cannot be configured by the http(s).Agent override.
```
jimmywarting, line 450: fix indentation, replace tab for space to be consistent

jimmywarting:
wouldn't the new recommended way to pass a Agent be thought the new httpOptions?

c-kruse:
@jimmywarting I’ll fix the whitespace and mull over a way to test this - it’s not obvious to me how I could do that without stubbing out the http package. Any ideas?

As far as the agent goes, I’m conflicted. The other idea I’ve entertained is only adding the insecureHTTPParser Parameter - I didn’t see anyone looking for any of the other options here https://nodejs.org/docs/latest-v12.x/api/http.html#http_http_request_options_callback

bitinn:
Please demostrate what's not possible via Agent?

bitinn:
ok looks like nodejs did not make it easy for us.

https://github.com/nodejs/node/commit/0082f62d9c

My preference is still not to introduce an open `httpOptions` that just pass everything in (as most options should be done via Agent), but rather to handle this specific case via a custom `insecureHTTPParser` option.

But if any members are into `httpOptions`, which is essentially *everything is possible but you are fully responsible for breaking fetch*, then I rest my case.

c-kruse:
@bitinn sounds like you beat me to the punch, but updated the PR description with an example of what the agent cannot do. (Thanks @roychri for writing that up in https://github.com/node-fetch/node-fetch/issues/724)

I'll wait for other maintainers to chime in.

tinovyatkin:
How insecure HTTP parser (and wider allowance in HTTP headers) will then work with our `Headers` that will do validation too?

c-kruse:
> How insecure HTTP parser (and wider allowance in HTTP headers) will then work with our `Headers` that will do validation too?

Thanks for pointing that out @tinovyatkin - it is filtering (silently dropping) invalid headers here: https://github.com/node-fetch/node-fetch/blob/e6bfe4d419aff7eb00887c6f6085dc6b8b43dcc3/src/headers.js#L250

I think that is probably acceptable behavior.

tinovyatkin:
But, I mean, what is the point if you will not be able to access those headers anyway?

tinovyatkin:
Should `setHost` option be handled specially in this case?

jimmywarting:
> it’s not obvious to me how I could do that without stubbing out the http package. Any ideas?

@c-kruse in https://github.com/node-fetch/node-fetch/blob/master/test/utils/server.js you got a test server, you could perhaps create a own response with invalid headers, if node don't allow you to respond with inaccurate headers then maybe you could test other httpOptions in other cases it would also be possible to create a server using `net` and craft a really low level http server/response. then in https://github.com/node-fetch/node-fetch/blob/master/test/request.js you would have to do the request to the local server.

> But, I mean, what is the point if you will not be able to access those headers anyway?
> The only difference of insecure parser is headers: `Use an insecure HTTP parser that accepts invalid HTTP headers when true`, but `node-fetch` `Headers` will not expose them anyway... So, why this feature should be inside `node-fetch`?

maybe so you at least got a graceful degration instead of flat out get rejected with an error of why you where not able to get a response?

jimmywarting:
would a annoying, nagging `console.warn` be useful? saying something like "hey, example.com/hello-world responded with invalid header token 'x-foobar' you should report this to example.com" 😬

c-kruse:
> what is the point if you will not be able to access those headers anyway?
To provide browser-like ability to ignore junk http headers in the node environment for anyone who wishes to opt-in. Today it raises the HPE_INVALID_HEADER_TOKEN error. If you force your browser to use http/1.1 you can catch it dropping invalid http headers from most sites running behind an Incapsula CDN.

> why this feature should be inside node-fetch?
I could be convinced either way. It seems like a relatively small number of us who are blessed with this problem. Selfishly, I don't want to maintain a fork of node-fetch or refactor the handful of projects we have using it to another client.

tinovyatkin:
@jimmywarting 
If it's browser-like behavior, maybe we should just enable `insecureHttpParser` for all requests by default? `node-fetch` filters them anyway and we will be on par with browser fetch...

jimmywarting:
> maybe we should just enable insecureHttpParser for all requests by default

nah, isn't the world going to a bit better if everyone just instead fix there bugs? i don't expect that the insecure http parser will be around for much long, insecure is... well, insecure

c-kruse:
Thanks everyone for the direction! Closing this for https://github.com/node-fetch/node-fetch/pull/856 as I think we're in agreement on the suggestion @bitinn made.

PR #863, opened 2020-06-05
fix: lint

xxczaki:
The latest version of `xo` introduced some new rules, one of which is the `unicorn/no-reduce`. More information available here: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/623.

My question is - do we want to use this rule or ignore it. It seems a bit controversial.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,7 +1,7 @@
 import {builtinModules} from 'module';
 import {dependencies} from './package.json';
 
-export default {
+const configuration = {
 	input: 'src/index.js',
 	output: {
 		file: 'dist/index.cjs',
@@ -16,3 +16,5 @@ export default {
 	},
 	external: [...builtinModules, ...Object.keys(dependencies)]
 };
+
+export default configuration;
--- a/package.json
+++ b/package.json
@@ -1,126 +1,126 @@
 {
-	"name": "node-fetch",
-	"version": "3.0.0-beta.6-exportfix",
-	"description": "A light-weight module that brings window.fetch to node.js",
-	"main": "./dist/index.cjs",
-	"module": "./src/index.js",
-	"sideEffects": false,
-	"type": "module",
-	"exports": {
-		"import": "./src/index.js",
-		"require": "./dist/index.cjs"
-	},
-	"files": [
-		"src",
-		"dist",
-		"@types/index.d.ts"
-	],
-	"types": "./@types/index.d.ts",
-	"engines": {
-		"node": ">=10.16"
-	},
-	"scripts": {
-		"build": "rollup -c",
-		"test": "node --experimental-modules node_modules/c8/bin/c8 --reporter=html --reporter=lcov --reporter=text --check-coverage node --experimental-modules node_modules/mocha/bin/mocha",
-		"coverage": "c8 report --reporter=text-lcov | coveralls",
-		"test-types": "tsd",
-		"lint": "xo",
-		"prepublishOnly": "node ./test/commonjs/test-artifact.js"
-	},
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/node-fetch/node-fetch.git"
-	},
-	"keywords": [
-		"fetch",
-		"http",
-		"promise"
-	],
-	"author": "David Frank",
-	"license": "MIT",
-	"bugs": {
-		"url": "https://github.com/node-fetch/node-fetch/issues"
-	},
-	"homepage": "https://github.com/node-fetch/node-fetch",
-	"funding": {
-		"type": "opencollective",
-		"url": "https://opencollective.com/node-fetch"
-	},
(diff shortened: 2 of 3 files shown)


Richienb:
It looks like we can rewrite all of the .reduce calls.

tinovyatkin:
I’ll be out of maintainers/contributors of this project should such rule be introduced, as I see no sense in limiting some portion of language that’s quite normal and easy to understand.

Richienb:
https://www.youtube.com/watch?v=qaGjS7-qWzg

LinusU:
That thread was a great read!

Our usage of `reduce` are both the kind that modifies and then passes the initial value back to be passed in as the accumulator in the next iteration. As discussed in the thread, I think that this is much clearer by creating a variable and mutating it directly, which is exactly what is happening ☺️ 

Using a `for` loop would avoid having to splat the iterator into an array, just to be able to call `reduce` on it.

https://github.com/node-fetch/node-fetch/blob/e6bfe4d419aff7eb00887c6f6085dc6b8b43dcc3/src/headers.js#L197-L200

```js
const result = {};
for (const key of this.keys()) result[key] = this.getAll(key); 
return result;
```

https://github.com/node-fetch/node-fetch/blob/e6bfe4d419aff7eb00887c6f6085dc6b8b43dcc3/src/headers.js#L207-L218

```js
const result = {}
for (const key of this.keys()) { 
 	const values = this.getAll(key); 
 	// Http.request() only supports string as Host header. 
 	// This hack makes specifying custom Host header possible. 
 	if (key === 'host') { 
 		result[key] = values[0]; 
 	} else { 
 		result[key] = values.length > 1 ? values : values[0]; 
 	} 
 }
return result;
```

-----

If we want to avoid mutating, we could use `fromEntries`:

```js
return Object.fromEntries([...this.keys()].map(key => [key, this.getAll(key)]));
```


```js
return Object.fromEntries([...this.keys()].map(key => {
 	const values = this.getAll(key); 

 	// Http.request() only supports string as Host header. 
 	// This hack makes specifying custom Host header possible. 
 	if (key === 'host') { 
 		return ['host', values[0]];
 	}

 	return [key, values.length > 1 ? values : values[0]];
}));
```

xxczaki:
Note to self: `Object.fromEntries` is available in Node.js 12+, so a polyfill would be needed.

Richienb:
We can use [fromentries](https://npmjs.com/fromentries) as a lightweight ponyfill.

tinovyatkin:
Introducing a polyfill to avoid the use of `Array.prototy.reduce` due to a breaking change in minor version of some lint library... Call me crazy... 👎

tinovyatkin:
I would better replace `xo` with direct use of ESLint/Prettier with their default `recommended` settings. 1000% more popular industry solution with great tooling support (is anybody installing `vscode-xo`?) without this kind of discussion.

rollup.config.js
```diff
@@ -1,7 +1,7 @@
 import {builtinModules} from 'module';
 import {dependencies} from './package.json';
 
-export default {
+const configuration = {
```
tinovyatkin, line 4: I know (and agree) that `default exports` are evil. But what is the sense to create named default exports instead of just default exports? 😮

xxczaki: XO is opinionated, so we might disagree with some rules, that it introduces. I personally like most of the default options. However, since we are working in a team, I feel like more controversial rules (such as `no-reduce`) should be discussed.

> But what is the sense to create named default export instead of just default export?

I would say readability, most likely.

---

I think that it's also worth mentioning, that XO doesn't only add additional rules with each release - when a controversial rule is given a strong NO from the community, it's removed. Maybe this will be the case with `unicorn/no-reduce`.

My current recommendation is:

If there are any "more readable" ways to replace `Array.reduce`, that do not require adding polyfills - we should use them. However, if the only good solution is `Object.fromEntries` - let's ignore this rule and maybe we could use it once our minimum Node.js version will be 12+ :laughing:

JefferyHus:
I don't see any problem using `reduce`, even when it comes to readability. There is no downside of using it. I am not familiar with `xo`, but are we able to disable a rule? If yes, just disable it and keep reduce code

rollup.config.js
jimmywarting: I never use linters in my own project they just come up with stupid opinionated rules. I follow some basic style guide and do whatever I want! Everything I want linters to do is to spot error/misstakes deprecation (that sort of things) in my code not force me with stylistic rules. copy/pasting code between projects becomes a pain than.

[Linting, It's Pretty Much All Pain And No Gain](https://www.bennadel.com/blog/3312-after-3-months-of-javascript-linting-its-pretty-much-all-pain-and-no-gain.htm)

For instance I still see `var` as useful and empty `catch` block as useful... if you want to use reduce fine, go ahead!

PR #866, opened 2020-06-05
[Spec] Should check body _source_ on redirect

tinovyatkin:

- [x] Bug fix

While working on #210 I went into a particular interpretation of fetch redirect step 9.

[fetch standard](https://fetch.spec.whatwg.org/#http-redirect-fetch):

```
If actualResponse’s status is not 303, request’s body is non-null, and request’s body’s source is null, then return a network error.
```

However, our source instead of body _source_ is checking created `Request.body`. This PR fixes that fact.
This is also important due to fact that we may (_want_) convert the body to a stream internally, so, we actually must check body source here.

In addition, well-outdated test dependency `resumer@0.0.0` was replaced with native `Readable.from`.

In addition, while doing the above replacement, the _hacky_ test for `Buffer.concat` failure was replaced with the correct implementation of the stream that can't be concatenated into a buffer.

This PR also bumps Node.JS minimum version to 10.17 from 10.16 (will also be required for #603 and for #210).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
 	],
 	"types": "./@types/index.d.ts",
 	"engines": {
-		"node": ">=10.16"
+		"node": ">=10.17"
 	},
 	"scripts": {
 		"build": "rollup -c",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
 		"mocha": "^7.1.2",
 		"p-timeout": "^3.2.0",
 		"parted": "^0.1.1",
-		"resumer": "0.0.0",
 		"rollup": "^2.10.8",
 		"string-to-arraybuffer": "^1.0.2",
 		"tsd": "^0.11.0",
--- a/test/main.js
+++ b/test/main.js
@@ -1,4 +1,5 @@
 // Test tools
+/* eslint-disable node/no-unsupported-features/node-builtins */
 import zlib from 'zlib';
 import crypto from 'crypto';
 import http from 'http';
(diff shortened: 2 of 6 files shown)


xxczaki:
@node-fetch/reviewers Could you please take a look at this PR?

package.json
```diff
@@ -17,7 +17,7 @@
 	],
 	"types": "./@types/index.d.ts",
 	"engines": {
-		"node": ">=10.16"
+		"node": ">=10.17"
```
larson-carter, line 20: Why was the node engine bumped up a version?

test/main.js
```diff
@@ -404,7 +404,7 @@ describe('node-fetch', () => {
 		const url = `${base}redirect/307`;
 		const options = {
 			method: 'PATCH',
-			body: resumer().queue('a=1').end()
+			body: stream.Readable.from('tada')
```
larson-carter, line 407: Why did we go from resumer to a stream?

larson-carter:
I just have a few questions as to why we are bumping up the dependencies?

package.json
tinovyatkin: To allow use of `Readable.from`. It will be also required for new version of `fetch-blob` and #603 and #210.

test/main.js
tinovyatkin: One dependency less? `resumer` is unmaintained and never got out of version `0.0.0` while `Readable.from` is built-in Node.JS functions that doing the same.

tinovyatkin: If the question is how the removal of `resumer` is relevant to this PR subject then the answer is easy - the "stream" that `resumer` is creating is not an instance of `Readable`.

Richienb:
@tinovyatkin @larson-carter wants to know why the minimum version of Node.js is being increased.

JefferyHus:
> @tinovyatkin @larson-carter wants to know why the minimum version of Node.js is being increased.

He already answered this question. 

`To allow use of Readable.from. It will be also required for the new version of fetch-blob and #603 and #210.`

Issue #541, opened 2018-10-25
Handling a BOM with .json()


pjirsa:
Anybody know how I should be handling a BOM with a returned json payload? Right now i'm getting this error when calling res.json()
`Exception: FetchError: invalid json response body at https://api_that_returns_json_with_bom reason: Unexpected token ? in JSON at position 0`

I tried adding an Accept header of `application/json; charset=utf-8` but that didn't resolve the issue.

jimmywarting:
Maybe something like this:
```js
let text = await res.text()
if (text.charCodeAt(0) === 0xFEFF) {
  text = text.substr(1)
}
const json = JSON.stringify(text)
```

Just a quick question: is native browser fetch able to handle it?

bitinn:
I don't think BOM in JSON is valid, even if browsers *might* handles it. Going to label this as won't fix.

ref: https://stackoverflow.com/questions/4990095/json-specification-and-usage-of-bom-charset-encoding

memonjabbar:
```
let text = await res.text()
try {
return JSON.parse(text)
} catch(error) {
console.error("Error:", err);
}
```

How about this solution @jimmywarting

jimmywarting:
Hmm, chrome seems to be able to decode BOM
```js
await new Response(String.fromCharCode(0xFEFF) + '1').json() // 1
```
What we are essentially doing is:
```js
const buffer = Buffer.from(String.fromCharCode(0xFEFF) + '1'
JSON.parse(buffer.toString()) // Error
```
We should stop using buffer and turn towards using TextDecoder instead, this removes the BOM
```js
JSON.parse(new TextDecoder().decode((Buffer.from(String.fromCharCode(0xFEFF) + '1'))))
```

While we are at it:
```js
text() {
  let text = ''
  const decoder = new TextDecoder()
  for await (const uint8 of this.body) {
    text += decoder.decode(uint8, { stream: true })
  }
  text += decoder.decode() // flush
  return text
}

json() {
  return this.text().then(JSON.parse)
}
```

rpaterson:
A more terse workaround:
```javascript
const json = JSON.parse((await res.text()).trim());
```
I think calling `trim()` on a String of JSON should always be safe.

jimmywarting:
Actually i think using textencoder is the right way to handle it.

ralyodio:
THis fixes it for me. was getting a BOM with authorize.net api and cloudflare workers:

[...]

PR #871, opened 2020-06-10
feat: parse large JSON responses in a separate thread

### LINKED ISSUE #611: BUG: Gateway timeout with large json response (~150KB)
I think the response.json() call is sychronous and causes random but frequent gateway timeouts from Express because it blocks the thread for large responses (~150KB). Going to try and work around this by using `text()` and having another async json library handle the json but would be good if this "just worked".

tinovyatkin:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [x] Other, please explain:

This PR fixes #611 by using standard `worker_threads` module when a big JSON body is received and `Worker` is available (available under `--experimental-worker` since Node 10.5, unflagged in 11.7).

The most difficult part here is to create a failing test, but this one seems to be working:

```js
it("should not block message loop on large json", async () => {
  const buf = readFileSync("./test/utils/big-fixture.json");
  const res = new Response(buf);
  let ticks = 0;
  const [json] = await Promise.race([res.json(), new Promise(() => {
				setInterval(() => {
					ticks++;
				}, 0);
			})]);
 expect(ticks).to.be.greaterThan(5);
});
```

We use a workaround in requiring `working_threads` until it becomes available in our minimal supported platform.


jimmywarting:
> @jimmywarting requested changes

^
Better title 😉

src/body.js
```diff
@@ -113,11 +115,20 @@ export default class Body {
...
 	 * Decode response as json
 	 *
-	 * @return  Promise
+	 * @returns {Promise<unknown>}
 	 */
 	async json() {
-		const buffer = await consumeBody(this);
-		return JSON.parse(buffer.toString());
+		const text = await this.text();
+		if (text.length > 50000 && typeof Worker === 'function') {
```
jimmywarting, line 122: don't know why I'm bothered by 50000. Prefer 2 to the power of x such as 

```
2<<15
65535
0xFFFF
0b10000000000000000
```
it is the maximum number representable by 2 bytes and can be defined by the type Uint16
calculating base10 is harder for computers

----

Damaged by all binary knowledge i got and knows how to count and gets easily bothered when someone mistakenly use KB wrongs when they might have meant to use the term `KiB` instead of `KB`

tinovyatkin: Old school? Me too :smile: But I was recently surprised by discovering that KB is officially power of 10 for quite a long time (see http://simplicitysimplifiedfurther.blogspot.com/2013/12/difference-between-kb-and-kib-mb-and.html) and we should resort to `KiB` for the power of 2. But it's just a threshold, we may put whatever number here.

jimmywarting: Yea, i'm old school i have even had 2 binary watches - one stationary and one arm wrist

tinovyatkin: So, 32768 here? I believe most people these days will say such numbers degrade maintainability.

tinovyatkin: Remember that even `.reduce` is blocked by `xo` and `++i` is blocked by AirBnB code style as "hard to understand" 😖

jimmywarting: 32768? Woot?
It shows that the time is 01:03:54

it displays as hh:mm:ss

reading from left to right - bottom to top

--

reduce i can understand `i++` or `++i` is not hard to understand. But like i mention before - hate linters opinion rules - just want a linter that can spot errors.

tinovyatkin: I mean 32768 instead of 50000

jimmywarting: oh, my bad - nah go full max 16bit value :)
65535 || 0xFFFF

tinovyatkin: I love how `0o100000` looks 😉

bitinn:
Color me surprised that error handling is exactly the same when you are on a worker thread. Does  worker thread really not introduce any other complication (security, eval permission, portability etc.)?

tinovyatkin:
@bitinn The parsing error is propagated into the main thread and exactly equal to `JSON.parse` error in the main thread (we don't have any tests for it currently, I may add one if you feel it's needed).

We are not evaluating any arbitrary text here, only two lines of our hardcoded code, so, no extra security concerns, neither _evaluation permissions_ are attached.

`worker_threads` are part of stable API of current LTS of Node.JS so, all the security, portability, and other concerns are exactly the same as for the rest of Node.js API.

bitinn:
Do we feel the experimental worker thread on Node 10 is stable enough to use *by default* for large response?

Or do we think the threshold should be configurable and/or only work when you are on Node 12?

tinovyatkin:
@bitinn it will use `worker_threads` on Node 10 only when a user is running node with `--experimental-worker-threads` and use default method otherwise.

bitinn:
@tinovyatkin I am nitpicking here but conditional require has been historically troublesome for bundling system to work with, usually they require extra use configuration (`webpack` did); I would like to avoid them if possible, but perhaps you already have a solution I am not aware of.

bitinn:
And the second question is whether `50000` is a good threshold, generally speaking.

tinovyatkin:
@bitinn most of the code changes in this PR is workaround for conditional require via Commonjs fallback. 
Remember that we are not bundling anything for Node 12+ and it directly uses our ESM code.
We have tests in place for commonjs bundle and they all passing.

tinovyatkin:
> And the second question is whether `50000` is a good threshold, generally speaking.

I did some manual checks (you may do too by outputing ticks variable at test to console) and I do thing it’s a good balance between overhead and benefits from the worker.

bitinn:
You have convince me this is a solid PR, but:

Personally I still think we shouldn't spawn thread on large size by default, most package doesn't do this, so the default assumption is `node-fetch` doesn't neither. I don't know if there are strict environment out there that limits threading, I feel this warrants an readme update, as crawler using node-fetch might run into issues.

@node-fetch/core @node-fetch/reviewers let us know how you feel about the non-configurable threshold and spawn by default :)

tinovyatkin:
@bitinn may be workers in Node are still too new, so, it feels like it's a big deal, but a decision, whether to launch some task in a new thread or current thread, is pretty common for other languages libraries (C++, etc) and in rare case configurable.
But taking a hint from your mention of _environments_ I believe that in the rare case when a user needs to customize/disable this behavior (let's imagine something like Electron / react native / etc) I believe this setting belongs more to an environment variable. 
So, I've introduced `NODE_FETCH_THREADED_PARSER_THRESHOLD` variable to be able to configure this in such environments. Setting it to `9007199254740991` (`Number.MAX_SAFE_INTEGER.toString()`) will effectively disable workers spawn.

It should be in line with configurations like https://nodejs.org/api/cli.html#cli_environment_variables or https://www.electronjs.org/docs/api/environment-variables

tinovyatkin:
I've also tested that [--disallow-code-generation-from-strings](https://nodejs.org/api/cli.html#cli_disallow_code_generation_from_strings) option is not affecting `Worker` spawn with `eval` option and they still work fine even in this case.

JefferyHus:
I am still convinced that using worker threads is an overkill for this situation, not to mention that 50kb or even 1mb is not huge and will not result in blocking the `event loop`. As mentioned in one of the comments I'd better go with `Asynchronous JSON parsing using Big-friendly JSON`

Also, in the situation of large JSON response, the parsing is mostly in the hands of the user because in a package we are not supposed to solve every life problem (I know this is good).

tinovyatkin:
@JefferyHus you are wrong about message loop blocking. You may test it yourself with the test included in this PR

larson-carter:
> let us know how you feel about the non-configurable threshold and spawn by default :)

I don't think that the right solution is to make a span by default. That could cause a few issues down the line.

I don't think that we should spawn a thread on a large scale like that. Especially by default.

tinovyatkin:
@larson-carter I'm assuring you can give an example of mentioned _issue down the line_ of using a stable API from Node LTS? Because I can't imagine any.

larson-carter:
@tinovyatkin I wouldn't necessarily call it an issue. I'm mainly wanting to bring the point of us dropping previous support

jimmywarting:
v3 got released with a minimum requirement on node v12.20
maybe revisit this? is this still something we would like to have?

tekwiz:
@tinovyatkin Would you mind rebasing on main?

jimmywarting:
@tekwiz
> Would you mind rebasing on main?

I don't think that would be enough... some of this stuff is quite outdated.
I would like to revisit this in a new PR and do a researching on my own if a `worker_thread` is worth the effort and also looking at what the cost of transferring/cloning the json object over a worker_thread is, if it really is worth the effort. or if there is any other possible solutions that don't involve using worker_threads and StructuralCloning, (or maybe even use [script.runInNewContext](https://nodejs.org/dist/latest-v17.x/docs/api/vm.html#scriptruninnewcontextcontextobject-options)) 

i can also note that using `buffer.toString()` should not be used anymore. Something better would be to use TextDecoderStream (or the older `TextDecoder().decode(chunk, {stream: true}))` solution that also handles BOM and utf8 for you. (see: https://github.com/node-fetch/node-fetch/issues/541)

another possible improvement could be to transfer the chunks to the worker instead and converting it to a string on the worker_thread instead, then you won't have any cost of cloning stuff and converting the buffer to string will be done outside of the main thread.

jimmywarting:
closing as it's rather outdated...

PR #873, opened 2020-06-11
Remove duplicated tests

xxczaki:
Data URI tests are already located in the `test/external-encoding.js` file.


test/main.js
```diff
@@ -2125,37 +2125,4 @@ describe('node-fetch', () => {
...
-
-			const b = await r.text();
-			expect(b).to.equal('the data:1234,5678');
-		});
-
-		it('should reject invalid data uri', () => {
-			return fetch(invalidDataUrl).catch(error => {
-				console.assert(error);
-				console.assert(error.message.includes('invalid URL'));
-			});
```
tinovyatkin, line 2158: It was my breaking case for earlier implementation of Data URI via RegExp (while every other test was passing). Data URI is super undocumented an I would prefer to keep as much test with different examples as possible. Can we move these tests together with other Data URI tests instead of just dropping it?

xxczaki: My bad, I didn't actually notice that some of these tests were different! I moved them to the `test/external-encoding.js` file ;)

test/external-encoding.js
```diff
@@ -30,5 +39,12 @@ describe('external encoding', () => {
 				expect(error.message).to.include('malformed data: URI');
 			});
 		});
+
+		it('should reject invalid data uri 2', () => {
+			return fetch('data:@@@@').catch(error => {
+				console.assert(error);
+				console.assert(error.message.includes('invalid URL'));
```
tinovyatkin, line 46: @xxczaki while we are here, can we replace `console.assert` here with proper Chai assertions?

xxczaki: @tinovyatkin I just removed this test, as with Chai assertions it would be identical to the previous one ;)

xxczaki: @tinovyatkin I just removed this test, as with Chai assertions it would be identical to the previous one ;)

PR #874, opened 2020-06-11
fix: min semver Node version

tinovyatkin:
I was wrong with my understanding of what _backported_ means in Node.js versions and `Readable.from` seems to skip the whole range of Node 11 (reached end of life already) and 3 earlier versions of Node 12.
This PR fixes that and removes the unneeded exemption from `eslint-plugin-node` rules.

~Not sure why diff become crazy here 😕~
Unrelated change to prevent such diffs in the future: added `package.json` into our `editorconfig`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,3 +14,8 @@ trim_trailing_whitespace = false
 
 [*.yml]
 indent_style = space
+
+[package.json]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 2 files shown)


tinovyatkin:
> annoying when tab vs space change all the time mixed with what the actual PR is about...

That should be changed by `editorconfig` changes in this PR.

Issue #751, opened 2020-03-17
fetch doesn't allow for AbortSignal custom impl with constructor name other then exactly AbortSignal


Veetaha:
May I ask you to elaborate on the purpose of this validation?

https://github.com/node-fetch/node-fetch/blob/7c66f9ce822a3a06fe01eb583a30b4c087c596c2/src/utils/is.js#L53-L58

For my use case, it is redundant. I want to pass a custom adapter from `vscode.CancellationToken` to `AbortSignal`:
```ts
import { AbortSignal } from "node_fetch/externals";

class CancellationSignal implements AbortSignal {
    subscriptions = new WeakMap<Listener, vscode.Disposable>();

    constructor(private readonly ct: vscode.CancellationToken) {}

    get aborted() {
        return this.ct.isCancellationRequested;
    }

    addEventListener(_type: "abort", listener: Listener, _opts?: unknown) {
        this.subscriptions.set(listener, this.ct.onCancellationRequested(listener, this));
    }

    removeEventListener(_type: "abort", listener: Listener) {
        this.subscriptions.get(listener)?.dispose();
    }

    // Some excess APIs that are not used by `node_fetch` impl
    onabort = null
    dispatchEvent() { return false; }
}
```
This is very annoying that I cannot name my class how I want. Besides that why can't I pass a simple object literal that conforms to `AbortSignal`? There is no such explicit contract that an instance of abort signal passed to fetch must be a custom class instance, since `AbortSignlas` is nothing but an interface...

dzek69:
I agree. This is also very annoying when you bundle and minify your code. Minifiers will change the name of your class and node-fetch will fail on that.

If somebody finds that useful - that's fine. But having some global configuration to disable this behavior would be nice.

jimmywarting:
Bundlers may minify classes and minify names but if you add `Symbol.toStringTag` to your class that's equal to `AbortSignal` then you may be fine with minimizing it.

Also i don't think speced polyfilled api's should mangle the class names to something short.

ww-k:
> Bundlers may minify classes and minify names but if you add [Symbol.toStringTag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) to your class that's equal to `AbortSignal` then you will be fine with minimizing it and using a own custom abort signal. That's how we check if it's an AbortSignal
> 
> Also i don't think speced polyfilled api's should mangle the class names to something short.

[...]

PR #876, opened 2020-06-11
Add a missing method when checking if an object is a spec-compliant `FormData` object

xxczaki:
MDN: https://developer.mozilla.org/en-US/docs/Web/API/FormData/has]
Spec: https://xhr.spec.whatwg.org/#dom-formdata-has

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -57,6 +57,7 @@ export function isFormData(object) {
 		typeof object.set === 'function' &&
 		typeof object.get === 'function' &&
 		typeof object.getAll === 'function' &&
+		typeof object.has === 'function' &&
 		typeof object.delete === 'function' &&
 		typeof object.keys === 'function' &&
 		typeof object.values === 'function' &&


jimmywarting:
OCD kicking in? :P

tinovyatkin:
Not exactly related to this PR, but I think we should start this discussion again 😄 
I'm pretty skeptical about this way of object testing that we are using in `node-fetch`. It's completely different things to check an object provided by an environment (`DOM` for example) just by `Symbol.toStringTag` vs concluding out of this symbol that some object will act as we believe in an environment where it doesn't belong natively.
Should we test methods arguments too (`set.length`)?
I would prefer to have some small reference implementations of `Blob`, `FormData`, and `AbortSignal` inside the `node-fetch` and just do `instanceof` - as it can be extended by any developer if needed.
For example, the following class will pass your test perfectly, but will not work as expected:

```js
class FormData extends URLSearchParams {
  get [Symbol.toStringTag]() {
    return "FormData";
  }
}
```

jimmywarting:
> I would prefer to have some small reference implementations of Blob, FormData, and AbortController inside the node-fetch organization and just do instanceof

That would increase the size of node-fetch. And you wouldn't be able to use similar polyfills (hint hint -> #751)

tinovyatkin:
@jimmywarting we have Blob already, FormData (without serialization) is almost complete in my example above and we have serialization built-in into `node-fetch`. `AbortController` is small too.

`instanceof` can be extended by base class via `Symbol.hasInstance` as I did today for `fetch-blob` or can be spoofed via Proxy, as in the example in #848, nothing particularly difficult.

jimmywarting:
@tinovyatkin Would like @bitinn or someone else opinion on that first before we decide on it - create a issue about it first

LinusU (approved):

jimmywarting:
I don't think this is worth merging if #1314 gets merged.
it basically brings in formdata-polyfill and fixes some stuff like the missing `body.formData()` while also using `symbol.hasInstance` to check if something is a spec compatible formData

this util function got wiped out completely by #1314

LinusU:
Ah, right, #1314 seems to be very close to getting merged so lets close this one ✔️

PR #880, opened 2020-06-12
Allow Mocha tests to be run in parallel

tinovyatkin:
This PR fixes our test suite to allow it to run in parallel with new mocha 8 feature:
`npx mocha --parallel`

It doesn't switch to this mode by default due to the fact that at the moment nor `c8` nor `nyc` can collect test coverage in this mode.

Basically this PR makes the test server to bind to a random port (0) instead of fixed one and removes the use of mocha _root hooks_.
One test was refactored into an async function because for some reason it was failing in parallel mode otherwise.

There are no speed improvements right now from running tests in parallel, mostly due to the fact that our current test suite is the monolith of one file, but it will/may improve both with mocha improvements and out tests de-composition, so, it's nice to have:

```sh
mocha --reporter=dot 
212 passing (2s)

# vs
mocha --parallel --reporter=dot
212 passing (3s)
```

Partially deprecates/helps for  #832 / #833

### DIFF
--- a/test/headers.js
+++ b/test/headers.js
@@ -1,5 +1,8 @@
 import {Headers} from '../src/index.js';
 import chai from 'chai';
+import chaiIterator from 'chai-iterator';
+
+chai.use(chaiIterator);
 
 const {expect} = chai;
 
--- a/test/main.js
+++ b/test/main.js
@@ -48,17 +48,6 @@ chai.use(chaiString);
 chai.use(chaiTimeout);
 const {expect} = chai;
 
-const local = new TestServer();
-const base = `http://${local.hostname}:${local.port}/`;
-
-before(done => {
-	local.start(done);
-});
-
-after(done => {
-	local.stop(done);
-});
-
 function streamToPromise(stream, dataHandler) {
 	return new Promise((resolve, reject) => {
 		stream.on('data', (...args) => {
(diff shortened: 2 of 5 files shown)


NotMoni:
Awesome job @tinovyatkin

tinovyatkin:
This only motivation to rewrite in AVA at #833 is parallel test execution. I really appreciate big work @xxczaki is doing there, but I do think that this rewrite will introduce more risks than benefits and parallel tests are already here.

Richienb:
I still think ava is the way to go because of its overarching simplicity and explicitly.

tinovyatkin:
I didn't mean to discuss `ava` vs `mocha` here as I have not enough experience to work with `ava` and these changes is beneficial anyway, I just saying that the only motivation of test runner switch at original issue #832 was the ability to run tests in parallel and it's here. 

<hr>

Regarding simplicity, I reserve my opinion that for an average JavaScript developer mocha will be simpler just because of the fact that it is over 20x more popular on **npmjs** and so _familiar_.

Richienb:
> 20x more popular on **npmjs**

I'm quoting Sindre here:

> Popularity is rarely a good metric for quality. Yes, it’s cool to have a popular project, but the most fun projects I’ve done are not very popular, and I couldn’t care less. Do it for you, not the stars.

The second motivation for migrating to ava was that **every other** repository in this org already uses it.

I also recommend reading [this article](http://zpalexander.com/migrating-from-mocha-to-ava/#whatswrongwithmochaanyway) which is pretty useful for mocha developers to get their head around ava.

tinovyatkin:
@Richienb I have some recommendations for you too 😄 
https://kentcdodds.com/blog/migrating-to-jest

tinovyatkin:
You can't compare test framework needs for `timeout-signal` with the same for `node-fetch`. Every other repository in @node-fetch was created on the shoulders of `node-fetch` popularity and have 1/20 of if tests count (at best). So, it's `node-fetch` needs should dictate the requirements in this organization.

Richienb:
Your opinion is yours and my opinion is mine.

Since this PR allows the current tests to run 30% faster, I'm approving it.

PR #884, opened 2020-06-13
chore: remove `p-timeout` test dependency

tinovyatkin:
[p-timeout](https://github.com/sindresorhus/p-timeout/blob/master/index.js) may be a good library for some complicated use case, but it's not small (57 LOC + 1 dependency) and we use it just at one place - to implement our `chai-timeout` that can be trivially implemented without an external dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
 		"form-data": "^3.0.0",
 		"formdata-node": "^2.2.0",
 		"mocha": "^8.0.0",
-		"p-timeout": "^3.2.0",
 		"parted": "^0.1.1",
 		"rollup": "^2.15.0",
 		"string-to-arraybuffer": "^1.0.2",
--- a/test/utils/chai-timeout.js
+++ b/test/utils/chai-timeout.js
@@ -1,11 +1,8 @@
-import pTimeout from 'p-timeout';
-
 export default ({Assertion}, utils) => {
 	utils.addProperty(Assertion.prototype, 'timeout', async function () {
-		let timeouted = false;
-		await pTimeout(this._obj, 150, () => {
-			timeouted = true;
-		});
+		const timeouted = await Promise.race([this._obj.then(() => false), new Promise(resolve => {
+			setTimeout(() => resolve(true), 150);
+		})]);
 		return this.assert(
 			timeouted,
 			'expected promise to timeout but it was resolved',


tinovyatkin:
@Richienb if under `reduces complexity` you mean [cyclomatic_complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) then it's lower in my rewrite.

Rewriting in AVA is no way close to complete and badly motivated (parallel tests are slower in our case). Anyway, `chai-timeout` will be useless in AVA, so, `p-timeout` will have to go anyway and this reference is non-related.

Richienb:
I'm not talking about the underlying complexity, just the surface complexity that abstracts everything away.

tinovyatkin:
yea... sorry, I don't know what surface complexity means in software development, and I have troubles to understand how I increased that...
My solution is faster and saves global internet bandwidth from bloated code if that matters.

Richienb:
> My solution is faster and saves global internet bandwidth from bloated code if that matters.

These are unit tests. The developers are the **only** primary stakeholder for these. Using p-timeout allows us to decrease the bloat code that **we** are responsible for maintaining.

tinovyatkin:
@Richienb 
Screenshot 2020-06-13 at 01 14 40

Richienb:
image

vs

image

tinovyatkin:
That even more complicated than `Array.prototype.reduce`! :smile:

jimmywarting:
I usually go nuts can do this kind of optimization to other 3rd party package i depend on (even if they are dev dependency). i sometimes look at `npm ls` and look for stuff i can dispose of (like how the hole lodash is included and only use 1% of it) so I welcome this kind of stuff. so points to you!

But since 

> It's also a devDependency which is used in tests so optimizing for performance is not the top priority

1 point to @Richienb

> Anyway, chai-timeout will be useless in AVA, so, p-timeout will have to go anyway and this reference is non-related.

1 point to @Richienb again since this will essentially be a unnecessary commit anyway

should instead put the focus on rewriting test to AVA - even if it may take a while to get there.

xxczaki:
I will take the opportunity and refer to @tinovyatkin's comments from another thread (#880) as it was closed.

> I'm just saying that the only motivation of test runner switch at original issue #832 was the ability to run tests in parallel and it's here.

One of the goals of rewriting the current test suite to AVA was to split the huge `main.js` test file into smaller pieces. This should allow us to utilize parallel testing better and make things easier for future contributors.

---

When it comes to this PR, let the others decide. I will focus on rewriting tests to AVA :smile:

tinovyatkin:
@jimmywarting I believe wording like _should instead put the focus on..._  - is the last thing an open-source project maintainer should say to an unpaid third-party contributor. But I'm hearing that second time here from you and I'd finally got your point.

Also you:
Screenshot 2020-06-13 at 20 19 56

PR #888, opened 2020-06-26
fix: add missing exports subpath "./package.json"

resessh:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

`require('node-fetch/package.json')` fails because package exports missing "./package.json" sub-path.
My repo which using nuxt needs to read package.json.

It's my first time to send a PR to node-fetch, so please let me know if there is anything to fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,11 @@
 	"sideEffects": false,
 	"type": "module",
 	"exports": {
-		"import": "./src/index.js",
-		"require": "./dist/index.cjs"
+		".": {
+			"import": "./src/index.js",
+			"require": "./dist/index.cjs"
+		},
+		"./package.json": "./package.json"
 	},
 	"files": [
 		"src",


xxczaki:
> My repo which using nuxt needs to read package.json.

Could you please share a reason why you need to read `package.json` files from dependencies?

resessh:
This is the repo for the reproduction.
https://github.com/resessh/reproduce-node-fetch-888

Briefly, nuxt checks the version of the package it depends on, and depends on node-fetch.
If nuxt fails to get `package.json` of dependencies, it throws an error and the process dies.

https://github.com/nuxt/nuxt.js/blob/6ab65fbf4f2b0d2348363c43f5c2fe81ef47cbc6/packages/builder/src/builder.js#L218-L224

I don't think it is bug that node-fetch doesn't export `package.json`. However, this is a breaking change, and I would like you to export it if possible.

Richienb:
Fixed in https://github.com/node-fetch/node-fetch/commit/b7076bb24f75be688d8fc8b175f41b341e853f2b

resessh:
@Richienb Thanks!!!

Mr1262:
@Richienb Thanks!!!

PR #902, opened 2020-07-20
Use generic typings for response.json()

tamas-nemeth:

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Other, please explain:

made response json() generic

```typescript
export interface User {
  id: string;
}

async function getUser(id: string) {
  const response = await fetch<User>(`http://localhost/api/user/${id}`);
  return response.json(); // Promise<User>
}
```

no breaking changes. generic is optional

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -100,7 +100,7 @@ type BodyInit =
 	| NodeJS.ReadableStream
 	| string;
 type BodyType = { [K in keyof Body]: Body[K] };
-declare class Body {
+declare class Body<T = unknown> {
 	constructor(body?: BodyInit, opts?: { size?: number });
 
 	readonly body: NodeJS.ReadableStream | null;
@@ -110,7 +110,7 @@ declare class Body {
 	buffer(): Promise<Buffer>;
 	arrayBuffer(): Promise<ArrayBuffer>;
 	blob(): Promise<Blob>;
-	json(): Promise<unknown>;
+	json(): Promise<T>;
 	text(): Promise<string>;
 }
 
@@ -142,7 +142,7 @@ declare class Request extends Body {
 	clone(): Request;
 }
 
-declare class Response extends Body {
+declare class Response<T = unknown> extends Body<T> {
 	constructor(body?: BodyInit | null, init?: ResponseInit);
 
 	readonly headers: Headers;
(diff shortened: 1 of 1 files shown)


LinusU:
I think that this is undesirable since it will hide an unsound cast that was previously explicit. Is there a downside to using an `as` cast instead?

```ts
export interface Item {
  id: string;
}

async function getUser(id: string) {
  const response = await fetch(`http://localhost/api/user/${id}`);
  return await response.json() as User;
}
```

tamas-nemeth:
Specifying the response type with a generic is a common pattern with `@angular/common/http`
`axios` and `got` also support that

The benefit of not having to repeat the type assertion when reading the data from the same response object is less relevant here, as `.json()` may only be called once. However, I still think it would be useful to support this pattern.

Also, as the generic is optional, users can still choose to make a type assertion instead.

LinusU:
How do you feel about the fact that one could write code that looks perfectly valid in strict mode TypeScript, but that would fail at runtime?

```typescript
export interface User {
  id: string
  name: {
    first: string
    last: string
  }
}

async function getUser(id: string) {
  const response = await fetch<User>(`http://localhost/api/user/${id}`)
  const user = await response.json()

  console.log(user.name.first) // <-- oops the server actually returned `{ id: 'a', firstName: 'b' }`
}
```

junaid1460:
Really it could be just

```typescript

type ParsedDataType = ReturnType<JSON['parse']> // is any

```

sergiojoker11:
> How do you feel about the fact that one could write code that looks perfectly valid in strict mode TypeScript, but that would fail at runtime?
> 
> ```ts
> export interface User {
>   id: string
>   name: {
>     first: string
>     last: string
>   }
> }
> 
> async function getUser(id: string) {
>   const response = await fetch<User>(`http://localhost/api/user/${id}`)
>   const user = await response.json()
> 
>   console.log(user.name.first) // <-- oops the server actually returned `{ id: 'a', firstName: 'b' }`
> }
> ```

There is very little you can do about that. In fact, the code would also break even without casting at runtime.
At the boundary of a HTTP client, all is uncertain. You will never be able to ensure what the API response is going to be like. (Unless you put more effort in and do something clever involving schemas or something like that. However, that is not the point of this PR) 
IMO, casting for this bit is acceptable. I'd argue very hard If saw casting inside an app though.

PR #909, opened 2020-08-06
Added support for capturing timing information for the request. 

tomkoufakis:

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Other, please explain:

The legacy request npm package provided the capability to return timing related data to the caller in the response. For example:

`phases: {
     wait: 1,
     dns: 53,
     tcp: 156,
     request: 313,
     firstByte: 153,
     download: 2,
     total: 678
   }`

We need this functionality and there is not a good away to add this, even with using shimmer. I have added this functionality in this PR.

To implement this, I used the package from: https://www.npmjs.com/package/@szmarczak/http-timer which includes 1 transitive dependency: https://www.npmjs.com/package/defer-to-connect.

Let me know if you have another way to include the timing data that I have not considered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Changelog
 
 # 2.x release
 
+## v2.6.1
+- Enhance: `Response` object now includes conditional `timings` property.
+
 ## v2.6.0
 
 - Enhance: `options.agent`, it now accepts a function that returns custom http(s).Agent instance based on current URL, see readme for more information.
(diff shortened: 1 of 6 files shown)


jimmywarting:
This would not be the road to go about implementing timing information for node-fetch.
The timing information shouldn't be part of a response object. and it shouldn't be an option for request to include a timings boolean option either

For something like this we need: [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) 

works something like this:
```js
const res = await fetch(url)
const text = await res.text()

const performanceResourceTiming = performance.getEntriesByType('resource').pop()
console.log(performanceResourceTiming)
```

LinusU:
I agree with @jimmywarting, we want to stay as close to the upstream `fetch` as possible.

Closing this since a long time has passed, but if you are interested in continued work on this, please open an issue and we could discuss the API ☺️

jimmywarting:
yup, now when Deno supports fetch builtin and NodeJS wanting to incorporate it into the core as well, it's most important that we follow the spec more strictly so you can expect the same behavior in all scenarios. the path forward would be PerformanceResourceTiming

PR #921, opened 2020-09-01
Allow passing agent option as an object 

### LINKED ISSUE #15: Allow to pass all available https options

faceleg:
See https://github.com/node-fetch/node-fetch/pull/236#issue-106457886

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Other, please explain:

Reinstated changes in #236 

We're using Ember Fastboot. We need to change our Fastboot URL to use HTTPS instead of HTTP. Without this change we experience SSL errors and are unable to proceed with updating our Fastboot URL.

I'm also planning on creating a PR into master, due to internal reasons we are unable to upgrade to node-fetch@3 and so need the change to go both into 2.6.0 and master.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,7 +46,13 @@ export default function fetch(url, opts) {
 		const request = new Request(url, opts);
 		const options = getNodeRequestOptions(request);
 
-		const send = (options.protocol === 'https:' ? https : http).request;
+    const protocol = (options.protocol === 'https:' ? https : http)
+		const send = protocol.request;
+
+ 		if (options.agent && typeof options.agent === 'object' && !(options.agent instanceof http.Agent)) {
+ 			options.agent = new protocol.Agent(options.agent)
+ 		}
+
 		const { signal } = request;
 		let response = null;
 
--- a/test/test.js
+++ b/test/test.js
@@ -1775,6 +1775,20 @@ describe('node-fetch', () => {
 		});
 	});
 
+  it('should create http.Agent class if options.agent is provided as an object', function() {
+		const url = `${base}inspect`;
+ 		const opts = {
+ 			agent: {
+ 				keepAlive: true
+ 			}
+ 		};
+ 		return fetch(url, opts).then(res => {
+ 			return res.json();
+ 		}).then(res => {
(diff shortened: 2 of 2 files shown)


bitinn:
We currently do not plan to maintain 2.x but you could create a fork if that’s mission critical to you and no workaround are possible. (2.6.0 is very stable just without some latest features).

tnawanna (approved):

src/index.js
```diff
@@ -46,7 +46,13 @@ export default function fetch(url, opts) {
 		const request = new Request(url, opts);
 		const options = getNodeRequestOptions(request);
 
-		const send = (options.protocol === 'https:' ? https : http).request;
+    const protocol = (options.protocol === 'https:' ? https : http)
```
LinusU, line 49: Indentation seems of here

```suggestion
		const protocol = (options.protocol === 'https:' ? https : http)
```

LinusU:
I think that this should be landed on `master` before back porting to 2.x? 🤔 

> I'm also planning on creating a PR into master, due to internal reasons we are unable to upgrade to node-fetch@3 and so need the change to go both into 2.6.0 and master.

Maybe we don't even need this on 2.x, is this still relevant? If it is, I think that landing this on master and then on 2.x should be the way forward ☺️

jimmywarting:
somewhat agree with @bitinn also about stop maintaining 2.x.
It needs a stronger case for landing in v2.

also the v3 has a function signature to allow for custom http or https agent

```js

fetch(url, {
  agent (url) { 
    return url.protocol === 'https:' ? httpsAgent : httpAgent
  }
})
```

so introducing a object into v2 while v3 don't support it would be weird, if anything then v2 should have the same functional approach as v3

 > Maybe we don't even need this on 2.x, is this still relevant? If it is, I think that landing this on master and then on 2.x should be the way forward ☺️

yea, i think so too, first make something into v3 and then backport something it if it's **really** needed (with same signature)

PR #924, opened 2020-09-04
fix(Body.body): Normalize `Body.body` into a `node:stream`

### LINKED ISSUE #210: body conversion

### LINKED ISSUE #1091: preparation for whatwg stream on `blob.stream()`
We want to support the real stream at some point and don't have to care whether or not it's a node stream or a blob stream

https://github.com/node-fetch/node-fetch/blob/6ee9d3186f8b4bdc4288ccf92f0a776aa17644ae/src/body.js#L178-L181

I think we need change that to this:
```diff
 // Body is blob 
 if (isBlob(body)) { 
- 	body = body.stream(); 
+       body = stream.Readable.from(body.stream());
 } 
```

---

I where namely thinking of if we should scale down fetch-blob's Blob.prototype.stream function into returning a partial piece of the whatwg stream (which is the async iterator that yields uint8arrays)

jimmywarting:
## Purpose
Normalize `Body.body` into a stream to be more in line with spec.
and also to simplify other parts so we deal with less underlying types of bodies

## Changes
Constructing a Response or Request with a string, blob or anything else gets converted into a `node:stream` now

Previously, what you passed into the Response or Request as a body would in the end be the same thing in `Body.body` (`new Response('abc').body === 'abc' `). This is wrong.

___

- [x] I added unit test(s)

## Additional 
Right now it's converted into a node:stream but i'm thinking it can be converted to a whatwg:stream in v4
it would be suppressing to me if someone rely on Response.body being whatever was passed to the Response constructor, not necessarily a stream. so perhaps this can slip in as a non major change?

I also lowercased the `nodefetchformdataboundary` cuz one issue is that Blob's type is casted to lowercase also
___

- fixes #210
- fixes #1092
- fixes #1091


gr2m (approved):

LinusU (approved):
Neat! 👍

PR #956, opened 2020-09-17
Minor optimization

jimmywarting:
No doc, bug or feature changes
No related issue

just a small optimization to avoid creating unnecessary anonymous functions

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -178,15 +178,13 @@ export default async function fetch(url, options_) {
 			}
 
 			// Prepare response
-			response_.once('end', () => {
-				if (signal) {
+			if (signal) {
+				response_.once('end', () => {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -178,15 +178,13 @@ export default async function fetch(url, options_) {
 			}
 
 			// Prepare response
-			response_.once('end', () => {
-				if (signal) {
+			if (signal) {
```
LinusU, line 181: note: `signal` here is declared with `const` and thus cannot change from null to object or vice versa

LinusU (approved):
Neat 👍

xxczaki (approved):

src/index.js
rami-HP: > note: `signal` here is declared with `const` and thus cannot change from null to object or vice versa

PR #1001, opened 2020-10-28
Support instances of `EventTarget` as a `signal`

worr:
Node has added its own AbortController in 15.0.0. The AbortController signal
inherits from EventTarget. This changes the check to also accept signals of
type EventTarget.

https://nodejs.org/docs/latest-v15.x/api/globals.html#globals_class_abortsignal

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Modified the `isAbortSignal` check to accept `EventTarget` as well.

Incompatibility with the new node `AbortController`

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -74,8 +74,10 @@ export function isFormData(object) {
  */
 export const isAbortSignal = object => {
 	return (
-		typeof object === 'object' &&
-		object[NAME] === 'AbortSignal'
+		typeof object === 'object' && (
+			object[NAME] === 'AbortSignal' ||
+			object[NAME] === 'EventTarget'
+		)
 	);
 };
 
--- a/src/request.js
+++ b/src/request.js
@@ -83,7 +83,7 @@ export default class Request extends Body {
 		}
 
 		if (signal !== null && !isAbortSignal(signal)) {
-			throw new TypeError('Expected signal to be an instanceof AbortSignal');
+			throw new TypeError('Expected signal to be an instanceof AbortSignal or EventTarget');
 		}
 
 		this[INTERNALS] = {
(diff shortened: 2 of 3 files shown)


jimmywarting:
> Node has added its own AbortController in 15.0.0.

Cool - have looked forward for this! More spec'ed browser apis!

> The AbortController signal inherits from EventTarget. 

Not as cool - then i hope that EventTarget later get an alias for addEventListener, removeEventListener and dispatchEvent

xxczaki (approved):

tinovyatkin (approved):

Richienb (approved):

src/utils/is.js
```diff
@@ -74,8 +74,10 @@ export function isFormData(object) {
  */
 export const isAbortSignal = object => {
 	return (
-		typeof object === 'object' &&
-		object[NAME] === 'AbortSignal'
+		typeof object === 'object' && (
+			object[NAME] === 'AbortSignal' ||
+			object[NAME] === 'EventTarget'
```
TimothyGu, line 79: I don't understand this. On Node.js v15.3.0, I see
```js
> new AbortController().signal[Symbol.toStringTag]
'AbortSignal'
```
So it seems like a bug to accept non-AbortSignal EventTarget objects.

Richienb: // @worr

jimmywarting: Just notice this. yea, it's weird that we check for the name `EventTarget`. a proper test was not even created... still using the polyfill for this...

node v12 goes EOL soon, so we are thinking of dropping support for node below v14.17 so that we can start using AbortController/AbortSignal provided by NodeJS so that we can pass this signal directly to node's own http.request.

Will create an issue for this. (targeted for v4)

PR #1050, opened 2020-12-24
Allow `signal` to be `null` or `undefined`

### LINKED ISSUE #1029: {signal: undefined} crashes node-fetch
**Reproduction**

```js
require('node-fetch')('http://example.com', { signal: undefined })
// TypeError: Expected signal to be an instanceof AbortSignal
```

**Expected behavior**

Same as
```js
require('node-fetch')('http://example.com', {})
```

---

Now, technically there is nothing in the fetch spec that says this is wrong:
```
28. If init["signal"] exists, then set signal to it.
30. If signal is not null, then make this’s signal follow signal.
```
where the [infra spec says](https://infra.spec.whatwg.org/#nulls)
```
The value null is used to indicate the lack of a value. It can be used interchangeably with the JavaScript null value. [ECMA-262]
```
`node-fetch` has implemented the `exists` using the `in` operator, which finds that `signal` exists even if it is `undefined`, and then uses it because it is not (`!==`) `null`. However, bearing in mind that the Fetch and Infra specs almost never mention the `undefined` value, I don't think it should be read this pedantically and `{signal: undefined}` as an `init` is intended to break the `fetch` call.
Also, with Firefox, Chromium, and https://github.com/github/fetch all implementations that I've checked handle this gracefully.

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.0.0-beta.8
| node             | 12.18.3
| npm              |
| Operating System |

tekwiz:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

This allows a falsey `signal` to be ignored.  e.g. 
```
fetch('https://google.com', { signal: undefined })
```

Fixes #1029

no

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -82,7 +82,8 @@ export default class Request extends Body {
 			signal = init.signal;
 		}
 
-		if (signal !== null && !isAbortSignal(signal)) {
+		// eslint-disable-next-line no-eq-null, eqeqeq
+		if (signal != null && !isAbortSignal(signal)) {
 			throw new TypeError('Expected signal to be an instanceof AbortSignal');
 		}
 
--- a/test/main.js
+++ b/test/main.js
@@ -1038,6 +1038,17 @@ describe('node-fetch', () => {
 		]);
 	});
 
+	it('should gracefully handle a nullish signal', () => {
+		return Promise.all([
+			fetch(`${base}hello`, {signal: null}).then(res => {
+				return expect(res.ok).to.be.true;
+			}),
+			fetch(`${base}hello`, {signal: undefined}).then(res => {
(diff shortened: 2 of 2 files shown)


tekwiz:
Note that the decreased coverage is due to existing coverage of <100% and a reduction of the number of times `isAbortSignal` is called in tests. There is no actual reduction in coverage, and there is a new test case specifically for this change.

xxczaki (approved):

jimmywarting (changes requested):

Richienb (approved):

jimmywarting (approved):

Issue #1055, opened 2021-01-01
response.body should fail if server terminates connection before response is fully transmitted


Gozala:
If server terminates connection before response is fully transmitted that should cause an error. In fact that seems to be the case when consuming the response e.g. via `response.text()` API. However if `response.body` is used directly via for await loop no errors are reported.

**Reproduction**

Test case is provided by #1056

**Expected behavior**

Following code should log error if connection is terminated before transmission is complete, instead loop is complete without any errors

```js
try {
  for await (const chuck of response.body) {
    // ...
  }
} catch(error) {
  console.error(error)
}
```

**Screenshots**

Same code in browsers produces error as expected here is an example (querying same server) as the attached test:

<img width="1022" alt="Screen Shot 2020-12-31 at 5 27 15 PM" src="https://user-images.githubusercontent.com/21236/103432198-20a52600-4b90-11eb-8c5f-2cfb3a3e78f2.png">

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | master
| node             | v12.18.2
| npm              |6.14.5
| Operating System | Mac

tekwiz:
@Gozala Please take a look at #1064 and see if it solves this problem.  You should be able to try it in your project with:

```sh
npm install 'github:tekwiz/node-fetch#fix-chunked-transfer-premature-close-2'
```

PR #1056, opened 2021-01-01
terimnated connections should propagate into response.body

Gozala:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain: Test case for the bug report

- Added a test case that illustrates the probem

#1055

~**Is there anything you'd like reviewers to know?**~

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -588,6 +588,24 @@ describe('node-fetch', () => {
 		});
 	});
 
+	it('should handle network-error in chunked response', () => {
+		const url = `${base}error/premature/chunked`;
+		return fetch(url).then(res => {
+			expect(res.status).to.equal(200);
+			expect(res.ok).to.be.true;
+
+			const read = async (body) => {
+				const chunks = [];
+				for await (const chunk of body) {
+					chunks.push(chunk);
+				}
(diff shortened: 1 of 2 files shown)


tekwiz:
@Gozala Is this intended to be a draft pull request?  I only see a failing test case for Node 12.  Also, it appears that Node 14 is unaffected by this bug.  Can you confirm?

tekwiz:
@Gozala  Nevermind my previous comment.  I understand what this is for now.  Thank you for the test case.  I may have a solution for this in the works already, but I'll keep you posted.

Gozala:
@tekwiz I did some investigation of this and I think the fix would require lifting this, up into the `response.body` itself: 

https://github.com/node-fetch/node-fetch/blob/d016690cd3ff3cfe48e87ff522379b4c85e402a9/src/body.js#L218-L230

However I was not sure what made sense here as I don't exactly understand rationale behind choosing to use node stream for `response.body` (as opposed to `ReadableStream`) & if propagating those errors was aligned with it.

tekwiz:
This was merged with #1064.

PR #1073, opened 2021-01-20
Removing confusingly named 'json' variable in examples

bartdorsey:
Using a variable named "json" for the parsed JSON data is confusing, because at that point the data is __not__ JSON anymore.

- [ X] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Other, please explain:

I changed the variable named "json" in all the examples to "data". The result of calling `response.json()` is __not__ JSON anymore.

Clarified usage of the node-fetch API

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,9 @@ console.log(body);
 const fetch = require('node-fetch');
 
 const response = await fetch('https://api.github.com/users/github');
-const json = await response.json();
+const data = await response.json();
 
-console.log(json);
+console.log(data);
 ```
 
 ### Simple Post
@@ -169,9 +169,9 @@ console.log(json);
 const fetch = require('node-fetch');
 
 const response = await fetch('https://httpbin.org/post', {method: 'POST', body: 'a=1'});
-const json = await response.json();
+const data = await response.json();
 
-console.log(json);
+console.log(data);
 ```
 
 ### Post with JSON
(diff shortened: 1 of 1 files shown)


tekwiz (approved):

jimmywarting:
have never thought about it like you do.
in a sens the variable json could also mean that the "okey this variable is part of a json response from a REST api i got earlier"
**data** could mean mostly anything. but you are right about that it isn't "json" anymore... it's a javascript literals. 

one annoying thing i find in certain api is that there response looks something like `{ errors: null, data: ... }`
then you will have `data.data` after calling `var data = await res.json()`

maybe a better more generic name would be something like `dataFromJson` but that is to long
something better maybe would be by calling it by what the response really is. `var user = await res.json()` but then you need to know what you expect from it

jimmywarting (approved):

bartdorsey:
> have never thought about it like you do.
> in a sens the variable json could also mean that the "okey this variable is part of a json response from a REST api i got earlier"
> **data** could mean mostly anything. but you are right about that it isn't "json" anymore... it's a javascript literals.
> 
> one annoying thing i find in certain api is that there response looks something like `{ errors: null, data: ... }`
> then you will have `data.data` after calling `var data = await res.json()`
> 
> maybe a better more generic name would be something like `dataFromJson` but that is to long
> something better maybe would be by calling it by what the response really is. `var user = await res.json()` but then you need to know what you expect from it

What I would normally do is to name the variable according to the context of what kind of data I was receiving. 

Example if I called an API to fetch a list of users....

```js
const response = await fetch("http://hostname/api/users")
const users = await response.json();
```

bartdorsey:
Oh also, if you have an API that returns `{error: '', data: ''}` then I typically will use destructuring to fix the name.

For the `/users` example, assuming it returned `{error: '', data: [<array of users>]}`

```js
const response = await fetch('/users');
const { data: users } = await response.json();
```

xxczaki (approved):

Issue #1072, opened 2021-01-20
Header.forEach inconsistent with standard fetch API


lquixada:
Hi this is the `cross-fetch` author. In order to keep maximum consistency between `node-fetch`, `whatwg-fetch` and native browser fetch, I'm running the same suite of tests across all these implementations. For the `3.0.0-beta.9` version, there are only two issues with Headers' `forEach` method that differs from browsers' fetch and `whatwg-fetch`:

1. It does not accept `thisArg` as the second parameter
2. The callback does not provide a third argument: the header object itself.

Here's a code to illustrate that:

```javascript
const headers = new Headers()
headers.append('Accept', 'application/json')

const obj = {}
const results = []
headers.forEach(function (value, key, object) {
  results.push({ value, key, object, thisArg: this })
}, obj);

console.log(results[0].object === headers) // true on browser, false on node-fetch 3.0.0-beta.9
console.log(results[0].thisArg === obj) // true on browser, false on node-fetch 3.0.0-beta.9
```

I realize that on version `2.6.1`, it used to have both features. So my question: is this intentional? If not, does anyone mind if I open a PR?

tekwiz:
@lquixada I believe this to be an oversight since `thisArg` is a part of the `forEach` definition for map-like iterators. A pull request would certainly be appreciated!

lquixada:
@tekwiz thanks! PR done: #1074

PR #1074, opened 2021-01-21
Improve Header's forEach method compliance with browser implementation.

lquixada:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Added `thisArg` to Header's forEach method and `header` parameter on its callback.

#1072 

There's an unrelated lint issue on `src/request.js` that's breaking the CI checks.

### DIFF
--- a/src/headers.js
+++ b/src/headers.js
@@ -167,9 +167,9 @@ export default class Headers extends URLSearchParams {
 		return value;
 	}
 
-	forEach(callback) {
+	forEach(callback, thisArg) {
 		for (const name of this.keys()) {
-			callback(this.get(name), name);
+			callback.call(thisArg, this.get(name), name, this);
 		}
 	}
 
--- a/src/headers.js
+++ b/src/headers.js
@@ -169,7 +169,7 @@ export default class Headers extends URLSearchParams {
 
 	forEach(callback, thisArg) {
 		for (const name of this.keys()) {
-			callback.call(thisArg, this.get(name), name, this);
+			Reflect.apply(callback, thisArg, [this.get(name), name, this]);
 		}
 	}
 
--- a/src/headers.js
+++ b/src/headers.js
@@ -167,7 +167,7 @@ export default class Headers extends URLSearchParams {
 		return value;
 	}
 
-	forEach(callback, thisArg) {
+	forEach(callback, thisArg = undefined) {
 		for (const name of this.keys()) {
 			Reflect.apply(callback, thisArg, [this.get(name), name, this]);
 		}
--- a/test/headers.js
+++ b/test/headers.js
@@ -69,12 +69,19 @@ describe('Headers', () => {
 		expect({key: 'content-type', value: 'text/html', object: headers}).to.deep.equal(results[1]);
 	});
 
-	it('should accept second thisArg argument for forEach', () => {
+	it('should accept thisArg as a second argument for forEach', () => {
 		const headers = new Headers({Accept: 'application/json'});
 		const thisArg = {};
 		headers.forEach(function () {
 			expect(this).to.equal(thisArg);
 		}, thisArg);
+	});
+
+	it('should allow only one argument for forEach', () => {
+		const headers = new Headers({Accept: 'application/json'});
+		headers.forEach(function () {
+			expect(this).to.be.undefined;
+		});
 	});
 
 	it('should allow iterating through all headers with for-of loop', () => {
--- a/test/headers.js
+++ b/test/headers.js
@@ -69,19 +69,19 @@ describe('Headers', () => {
 		expect({key: 'content-type', value: 'text/html', object: headers}).to.deep.equal(results[1]);
 	});
 
+	it('should set "this" to undefined by default on forEach', () => {
+		const headers = new Headers({Accept: 'application/json'});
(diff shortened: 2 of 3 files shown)


src/headers.js
```diff
@@ -167,9 +167,9 @@ export default class Headers extends URLSearchParams {
 		return value;
 	}
 
-	forEach(callback) {
+	forEach(callback, thisArg) {
```
TimothyGu, line 170:
```suggestion
	forEach(callback, thisArg = undefined) {
```

This allows `headers.forEach.length === 1`, which matches browsers.

src/headers.js
```diff
@@ -167,9 +167,9 @@ export default class Headers extends URLSearchParams {
...
 		for (const name of this.keys()) {
-			callback(this.get(name), name);
+			callback.call(thisArg, this.get(name), name, this);
```
TimothyGu, line 172: This allows `callback` to be an object with a `call` property, which isn't what we want. Can we instead use `Reflect.apply`?

test/headers.js
```diff
@@ -53,6 +53,30 @@ describe('Headers', () => {
...
+		expect({key: 'accept', value: 'application/json, text/plain', object: headers}).to.deep.equal(results[0]);
+		expect({key: 'content-type', value: 'text/html', object: headers}).to.deep.equal(results[1]);
+	});
+
+	it('should accept second thisArg argument for forEach', () => {
+		const headers = new Headers({Accept: 'application/json'});
+		const thisArg = {};
+		headers.forEach(function () {
+			expect(this).to.equal(thisArg);
+		}, thisArg);
```
TimothyGu, line 77: Can you also test the `this` value where `thisArg` is not supplied?

src/headers.js
lquixada: hmm a couple of things here:
1. I can add that for extra clarity but the end result feels the same as `thisArg` will be `undefined` when `headers.forEach.length === 1` anyways.
2. To match browsers, the default would be a global [object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description) such as window. On node, I guess `global` would do the job. Try this: `[1].forEach(function () { console.log(this === global) })`

src/headers.js
lquixada: sure! good call 👍

test/headers.js
lquixada: sure! waiting for considerations on https://github.com/node-fetch/node-fetch/pull/1074#discussion_r561536734.

src/headers.js
TimothyGu: > I can add that for extra clarity

The result indeed is the same inside `forEach`, but the `= undefined` is _essential_ to get the correct `.length` property. Try

```js
console.log(({ f(a, b) {} }).f.length);
console.log(({ f(a, b = undefined) {} }).f.length);
```

You will see that the first statement prints 2, while the second prints 1. We want the second behavior for `forEach`, and that's why you should add it. But indeed, the expected behavior is that if `thisArg` is not supplied, then it defaults to `undefined`.

By the way, the `.length` property of the `forEach` method is defined in the Web IDL standard, see ["define the iteration methods"](https://heycam.github.io/webidl/#define-the-iteration-methods) step 2.4.4.

> To match browsers, the default value would be a global object such as window.

That's not quite right. In browsers, the default value is `window` _only if_ we are not in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode):

image

Our test environments are all strict mode, but you can run `Reflect.apply(function() { console.log(this); }, undefined, [])` in the Node.js REPL (which is sloppy by default) to see that `this` is actually `global` in Node.js too.

All this is to say, you don't have to do anything special to get the global object behavior. (In fact, you _cannot_ do anything here to replicate that in pure JavaScript, since there's no easy way of detecting whether a function is strict or sloppy.)

If you're interested in finding where this behavior is specified, it's in the ECMAScript standard, within [OrdinaryCallBindThis](https://tc39.es/ecma262/#sec-ordinarycallbindthis) abstract operation's steps 5–6.

lquixada: hey @TimothyGu ! Thanks for that! It was very informative. All suggestions are committed now. Let me know if there's anything else I should do.

Btw, there's just a lint issue going on that I'm not able to fix. :/

TimothyGu:
See https://github.com/node-fetch/node-fetch/blob/master/src/request.js#L85. You could add a comment to ignore the check, like https://github.com/node-fetch/node-fetch/blob/4abbfd231f4bce7dbe65e060a6323fc6917fd6d9/src/request.js#L53

lquixada:
> See https://github.com/node-fetch/node-fetch/blob/master/src/request.js#L85. You could add a comment to ignore the check, like
> 
> https://github.com/node-fetch/node-fetch/blob/4abbfd231f4bce7dbe65e060a6323fc6917fd6d9/src/request.js#L53

done! 👍

TimothyGu (approved):
LGTM. I think someone else has to review this though.

xxczaki (approved):
Nice work!

tekwiz (approved):

PR #1078, opened 2021-01-27
Add support for Response.error()

tkrotoff:

- [x] New feature

Add support for `Response.error()` and `type`:
- https://developer.mozilla.org/en-US/docs/Web/API/Response/error
- https://developer.mozilla.org/en-US/docs/Web/API/Response/type

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -142,6 +142,8 @@ declare class Request extends Body {
 	clone(): Request;
 }
 
+type ResponseType = "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect"
+
 declare class Response extends Body {
 	constructor(body?: BodyInit | null, init?: ResponseInit);
 
@@ -150,8 +152,11 @@ declare class Response extends Body {
 	readonly redirected: boolean;
 	readonly status: number;
 	readonly statusText: string;
+	readonly type: ResponseType;
 	readonly url: string;
 	clone(): Response;
+
+	static error(): Response;
 }
 
 declare class FetchError extends Error {
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@
 		- [new Response([body[, options]])](#new-responsebody-options)
 		- [response.ok](#responseok)
 		- [response.redirected](#responseredirected)
+		- [response.type](#responsetype)
 	- [Class: Headers](#class-headers)
 		- [new Headers([init])](#new-headersinit)
 	- [Interface: Body](#interface-body)
@@ -587,9 +588,6 @@ An HTTP(S) response. This class implements the [Body](#iface-body) interface.
 
 The following properties are not implemented in node-fetch at this moment:
 
-- `Response.error()`
-- `Response.redirect()`
-- `type`
 - `trailer`
 
 #### new Response([body[, options]])
@@ -614,6 +612,12 @@ Convenience property representing if the request ended normally. Will evaluate t
 <small>_(spec-compliant)_</small>
 
 Convenience property representing if the request has been redirected at least once. Will evaluate to true if the internal redirect counter is greater than 0.
+
+#### response.type
+
+<small>_(deviation from spec)_</small>
+
+Convenience property representing the response's type. node-fetch only supports `'default'` and `'error'` and does not make use of [filtered responses](https://fetch.spec.whatwg.org/#concept-filtered-response).
 
 <a id="class-headers"></a>
 
--- a/src/response.js
+++ b/src/response.js
@@ -21,7 +21,6 @@ export default class Response extends Body {
 	constructor(body = null, options = {}) {
 		super(body, options);
 
-		// TODO Replace with nullish coalescing operator (??) when support for Node.js < 14 is dropped
 		// eslint-disable-next-line no-eq-null, eqeqeq, no-negated-condition
 		const status = options.status != null ? options.status : 200;
 
--- a/src/response.js
+++ b/src/response.js
@@ -21,7 +21,10 @@ export default class Response extends Body {
 	constructor(body = null, options = {}) {
 		super(body, options);
 
-		const status = options.status || 200;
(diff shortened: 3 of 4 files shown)


@types/index.d.ts
```diff
@@ -150,8 +150,11 @@ declare class Response extends Body {
 	readonly redirected: boolean;
 	readonly status: number;
 	readonly statusText: string;
+	readonly type: string;
```
tekwiz, line 153: @tkrotoff `type` should be the `ResponseType` enum: https://fetch.spec.whatwg.org/#responsetype

src/response.js
```diff
@@ -21,7 +21,8 @@ export default class Response extends Body {
 	constructor(body = null, options = {}) {
 		super(body, options);
 
-		const status = options.status || 200;
+		// TODO Replace with nullish coalescing operator (??) when support for Node.js < 14 is dropped
+		const status = options.status !== null && options.status !== undefined ? options.status : 200;
```
tekwiz, line 25: @tkrotoff Please use `options.status != null` for this comparison, and make sure to put the eslint-disable directive on the preceding line: `// eslint-disable-line no-eq-null, eqeqeq`

README.md
```diff
@@ -587,9 +587,6 @@ An HTTP(S) response. This class implements the [Body](#iface-body) interface.
 
 The following properties are not implemented in node-fetch at this moment:
 
-- `Response.error()`
```
tekwiz, line 590: @tkrotoff Please add a note in an appropriate place in the README that only `default` and `error` response types are used and that node-fetch does not make use of filtered responses https://fetch.spec.whatwg.org/#concept-filtered-response

tekwiz (changes requested):

@types/index.d.ts
tkrotoff: Should I
- use ResponseType from TypeScript lib.dom.d.ts
- create my own with just `type ResponseType = "default" | "error"`
- create my own with everything: `type ResponseType = "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect"`

tekwiz: Please create your own with everything from the spec.  While we only make use of `default` and `error`, they could be supported at some time in the future, and having them all maintains parity with the spec and other implementations.

tkrotoff: Done. However *could be supported at some time in the future* sounds YAGNI to me. Either use lib.dom.d.ts if you want to be future proof or provide your own types if you want to inform your users some stuffs are missing. I don't see the point of copy-pasting lib.dom.d.ts.

src/response.js
tkrotoff: Done.
```JS
// eslint-disable-next-line no-eq-null, eqeqeq, no-negated-condition
const status = options.status != null ? options.status : 200;
```

README.md
tkrotoff: Done.

```
#### response.type

<small>_(deviation from spec)_</small>

Convenience property representing the response's type. node-fetch only supports `'default'` and `'error'` and does not make use of [filtered responses](https://fetch.spec.whatwg.org/#concept-filtered-response).
```

src/response.js
tekwiz: @tkrotoff Sorry, one more thing.  Since Node.js 12 will be supported for at least 16 more months, please take out the TODO comment.  I'm sure we'll do a big push to update those kinds of changes at that point.

@types/index.d.ts
tekwiz: @tkrotoff You make very good points.  Your thoughts about using lib.dom.d.ts are probably better suited for a larger refactor, and I wouldn't want to hold up this one.  I only have a cursory understanding of TypeScript, but I would recommend making a separate PR for that or opening an issue to ask about the appetite for such a refactor.  Also, I understand the YAGNI concern, and I would agree if this were a product. With it being a library, we want to ensure things can be augmented as needed.

src/response.js
tkrotoff: Done.

tinovyatkin (commented):
We are supporting `redirect: "manual"`, so, any particular reason to not support `opaqueredirect` to `type` implementation?

@types/index.d.ts
```diff
@@ -142,6 +142,8 @@ declare class Request extends Body {
 	clone(): Request;
 }
 
+type ResponseType = "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect"
```
tinovyatkin, line 145: Why do we want to include type values that we will never actually return?

tkrotoff: I've made the same comment above.
Let's settle this:

- Use ResponseType from TypeScript lib.dom.d.ts
  Votes: none 
- Create our own with just `type ResponseType = "default" | "error"`
  Votes: 
  - my preference
  - @tinovyatkin 
- Create our own with everything: `type ResponseType = "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect"`
  Votes:
  - @tekwiz 
  
What do we do?

tekwiz (approved):

@types/index.d.ts
tekwiz: @tinovyatkin My argument is this: just because we don't use them doesn't mean they're disallowed.  In addition to maintaining parity with the spec, it should remain possible for someone to extend node-fetch to include filtered responses, and limiting the ResponseType seems unnecessarily restrictive.  That said, it is entirely possible that my limited understanding of TypeScript is leading me to the wrong conclusion about the complexity of adding to the type.

tekwiz:
> We are supporting `redirect: "manual"`, so, any particular reason to not support `opaqueredirect` to `type` implementation?

I don't see any reason why not, but I also think that would be outside the scope of the described intention for this pull request.

tkrotoff:
Any news? :-)

tkrotoff:
Ping

tekwiz:
@tinovyatkin Do you have any other changes that need to be made on this?

tinovyatkin (approved):

PR #1092, opened 2021-02-11
Wrap blobs into a node stream

### LINKED ISSUE #1091: preparation for whatwg stream on `blob.stream()`

jimmywarting:
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Other, please explain:

Regarding what type of blob.stream it is (whether it's a node stream, whatwg stream or just a async iterator that yields uint8arrays)
always wrap the stream it returns in to a node stream.

closes #1091

`buffer.Blob()` is soon becoming a thing...

### DIFF
--- a/src/utils/blob-to-stream.js
+++ b/src/utils/blob-to-stream.js
@@ -1,5 +1,6 @@
 import {Readable} from 'stream';
 
+/* c8 ignore start */
 async function * read(blob) {
 	let position = 0;
 	while (position !== blob.size) {
@@ -10,6 +11,7 @@ async function * read(blob) {
 		yield new Uint8Array(buffer);
 	}
 }
+/* c8 ignore end */
 
 export function blobToNodeStream(blob) {
 	return Readable.from(blob.stream ? blob.stream() : read(blob), {
--- a/src/utils/blob-to-stream.js
+++ b/src/utils/blob-to-stream.js
@@ -1,0 +1,18 @@
+import {Readable} from 'stream';
+
+async function * read(blob) {
+	let position = 0;
+	while (position !== blob.size) {
+		const chunk = blob.slice(position, Math.min(blob.size, position + 524288));
+		// eslint-disable-next-line no-await-in-loop
+		const buffer = await chunk.arrayBuffer();
+		position += buffer.byteLength;
+		yield new Uint8Array(buffer);
+	}
+}
+
+export function blobToNodeStream(blob) {
+	return Readable.from(blob.stream ? blob.stream() : read(blob), {
+		objectMode: false
+	});
+}
--- a/src/utils/blob-to-stream.js
+++ b/src/utils/blob-to-stream.js
@@ -1,10 +1,13 @@
 import {Readable} from 'stream';
 
+// 64 KiB (same size chrome slice theirs blob into Uint8array's)
+const POOL_SIZE = 65536
+
 /* c8 ignore start */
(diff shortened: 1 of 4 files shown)


jimmywarting:
**Edit** I also added a test for feature buffer.Blob and i also made a helper function to read blobs that don't have any stream function yet.

jimmywarting:
there is three issues with buffer.Blob

- don't have stream
- don't have Symbol.toStringTag
- slicing is not the same as in the spec, get ERR_OUT_OF_RANGE error when slicing something with lower/larger size

src/utils/blob-to-stream.js
```diff
@@ -0,0 +1,20 @@
+import {Readable} from 'stream';
+
+/* c8 ignore start */
+async function * read(blob) {
+	let position = 0;
+	while (position !== blob.size) {
+		const chunk = blob.slice(position, Math.min(blob.size, position + 524288));
```
tinovyatkin, line 7: Can we add some explanation (comment or meaningful constant name) for this magic number? I understand it workaround for Node.js incomplete implementation, but we don't know how long this code will stay, so, let's make it a little bit more maintanable

tinovyatkin (approved):

src/utils/blob-to-stream.js
jimmywarting: Don't actually remember where i got it from. it's exactly 512 KiB or a half MiB
My aim was to get something reasonable small that would fit nicely into any typed array so it would be the power of 2 and also dividable by 8.
and also make it fit nicely into one so called "pool size" of any optimal size allocation

Maybe it was too big?

looking at what slice size chrome uses it seems to be 64 KiB / slice
```js
(await new Blob(['s'.repeat(10000000)]).stream().getReader().read()).value.byteLength // 65536
// it is the same size as 
new BigUint64Array(8192).buffer.byteLength // 65536
// 8192 is the same size of nodes default pool size
// So it means 8192 * node's default buffer pool size... any correlation?
```
- FF slice size is only 4096 - many calls back and forth
- Nodes buffer default pool size is [8192](https://nodejs.org/api/buffer.html#buffer_class_property_buffer_poolsize)
- node streams default high watermark is set to 16 KiB.

don't know what it's but 65536 seems appealing to me. also one of those [tiles](https://www.csie.ntu.edu.tw/~b01902112/65536/) games

tinovyatkin: 64K sounds better for me, still with a comment that its Chome behavior

jimmywarting: i can replace it with 65536 as a constant with some comment above it

jimmywarting:
guess we might as well turn this to v4's whatwg:stream instead

jimmywarting:
picked up my old #924 PR instead.

Issue #417, opened 2018-03-08
Support access to raw response location header values


mfulton26:
`node-fetch` appears to be transforming response headers and resolving relative URLs against the request URL which means in order to assert that a correct relative URL is returned by the server one cannot use fetch. 😞 

Consider the following HTTP messages:

- Request:
  ```
  GET / HTTP/1.1
  Host: localhost:3000
  ```
* Response:
  ```
  HTTP/1.1 200 OK
  X-Powered-By: Express
  Location: /
  Content-Type: text/html; charset=utf-8
  Content-Length: 170
  ETag: W/"aa-z+ebXSEdArbZ+EXlN/WQjf6HV8c"
  Date: Thu, 08 Mar 2018 18:18:38 GMT
  Connection: keep-alive
  
  <!DOCTYPE html><html><head><title>Express</title><link rel="stylesheet" href="/stylesheets/style.css"></head><body><h1>Express</h1><p>Welcome to Express</p></body></html>
  ```

Using `node-fetch` to request the resource and extract the `Location` response header might look like the following:

```js
const response = await fetch("http://localhost:3000/")
const location = response.headers.get("Location");
assert.equal(location, "/");
```

The assertion, however, will fail because the `Location` response header value is in fact `"http://localhost:3000/"` instead of simply `"/"` even though the server is in fact simply returning `"/"`.

A `Location` response header value of `"/"` is accessible through `fetch` in Google Chrome (I tested it via the DevTools console); it is also available when using Node.js's `http` module directly. From what I can tell `node-fetch` is transforming these response headers for some reason but now the raw values are inaccessible.

It would be very helpful to either not transform these values and return them in `Response` objects as the server sent them.

Thank you.

TimothyGu:
What version of node-fetch are you using? I suspect this might have been fixed in v2.1.1.

mfulton26:
Oh. `"node-fetch": "^1.7.3"`. I'll try 2.1.1+.

mfulton26:
I just checked v2.1.1 and I'm seeing the same issue. I also found some what I believe is the relevant node-fetch code for this: https://github.com/bitinn/node-fetch/blob/780598ad72d7205aa5616be2215692785270eda3/src/index.js#L79-L80

It references HTTP fetch step 5.3 which says "If location is a value, then set location to the result of parsing location with actualResponse’s url." I'm not sure what that means but it almost makes it sound like Google Chrome should be behaving more like node-fetch but isn't? Not sure.

[...]

PR #1093, opened 2021-02-12
Update @types with manual-dont-change value for RequestRedirect #417 a201667c

GrzegorzKozub:
Commit a201667c addressed an issue #417 by providing a workaround but a small omission was not to reflect the changes in TypeScript type definitions.


GrzegorzKozub:
This is a misunderstanding.

Issue #917, opened 2020-08-18
Bug: I get "TypeError: Illegal invocation" when using node-fetch@3 with electron


badersur:
I'm not sure if this is a bug but I think so. I have an electron app that uses node-fetch@2 which is working fine (in a GUI and the terminal), but with version 3 when I only run the app in the terminal (I get a TypeError when run in the GUI).

This is maybe due to dropping support for the browser but I don't think so! The app uses node-fetch in the GUI using a preload script which have access to full node APIs as far as I understand, and it should be running as if it's running from a terminal! Or maybe it's an issue with [Electron](https://github.com/electron/electron/issues/21517) itself?!

**Reproduction**

Using node >=12.14.1 should be fine but I use version 14.8.0.

1. `git clone https://github.com/badersur/miniapp.git`
2. `cd miniapp`
3. `git checkout node-fetch-3`
4. `yarn`
5. `yarn dev`

**Expected behavior**
The app should work fine as shown in the master branch:

```
git checkout master
yarn
yarn dev
# Passing `-t` to the app will output to the terminal (which is working fine using both versions), for Linux I use:
yarn dist:dir && ./dist/linux-unpacked/miniapp -t
```

**Screenshots**

![Screenshot from 2020-08-18 18-11-01](https://user-images.githubusercontent.com/1686391/90523672-3801ea80-e17e-11ea-8f53-41eca671dc5f.png)

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.0.0-beta.8
| node             | 14.8.0
| npm              | 6.14.7
| yarn              | 1.22.4
| Operating System | Ubuntu 20.04.1 LTS

**Additional context**

TimoZikeli:
Have you found some workaround for this bug? Thanks!

badersur:
No, but I tried other libraries (`got`, `axios` & `superagent`) that worked so I'm thinking it's a bug with node-fetch not Electron!

I'm still wondering why `got` worked even though they say that it have *no* support for Electron & the browser!

jimmywarting:
`Illegal invocation` normally happens when a function gets unaware/detached from it's class.

example:
![Skärmavbild 2020-09-08 kl  20 07 19](https://user-images.githubusercontent.com/1148376/92512359-f6111500-f20e-11ea-9cc5-6fc684ca03e0.png)

This can easily be fixing if you bind the function `set = header.set.bind(header)`

[...]

PR #1111, opened 2021-03-08
Fix `this` arg for URLSearchParams prototype call

JosephusPaye:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Changed the `this` arg passed to the URLSearchParams prototype call in the Headers proxy handlers for `append()`, `set()`, `delete()`, `has()`, `getAll()`. Currently it's set to `receiver`, but it should be `target`, as it is [with the `keys()` handler](https://github.com/node-fetch/node-fetch/blob/d8fc32d6b29bd43d1ad377e80b3e439fe37f2904/src/headers.js#L137).

#917 

Ran the tests in Node (v14.15.0) and everything continues to pass. Manually tested in Electron (v11.1.1, with Node v12.18.3 and Chrome v87.0.4280.88), and it fixes the issue described in #917.

### DIFF
--- a/src/headers.js
+++ b/src/headers.js
@@ -114,7 +114,7 @@ export default class Headers extends URLSearchParams {
 							validateHeaderName(name);
 							validateHeaderValue(name, String(value));
 							return URLSearchParams.prototype[p].call(
-								receiver,
+								target,
 								String(name).toLowerCase(),
 								String(value)
 							);
@@ -126,7 +126,7 @@ export default class Headers extends URLSearchParams {
 						return name => {
 							validateHeaderName(name);
 							return URLSearchParams.prototype[p].call(
-								receiver,
+								target,
 								String(name).toLowerCase()
 							);
 						};


tekwiz (changes requested):
The fact that tests were already passing is concerning. Please add a test case

JosephusPaye:
The error is not reproducible in Node, which is why the tests are passing. It occurs in Electron though, which suggests it could be a difference in the implementation of `URLSearchParams` in Node vs Chromium, or something different about how that class interacts with a Proxy. I don't have any evidence of a difference though, other than the fact that the library definitely fails in Electron with a proxy error originating from the lines changed in the PR, as documented in #917.

From reading the current code, as well as [the Proxy docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/get#parameters), it seems the current usage only works incidentally. From what I can tell, the object being bound to `this` is meant to be the original object being proxied (`target`), not `receiver` (which is itself a proxy). This will match the correct usage in the [`keys()` handler](https://github.com/node-fetch/node-fetch/blob/d8fc32d6b29bd43d1ad377e80b3e439fe37f2904/src/headers.js#L137).

I'm happy to add a test if you can point me to a way to run the Mocha tests in Electron. I've tried [this library](https://www.npmjs.com/package/electron-mocha), but it doesn't work (I get ES module errors originating from within Mocha's code about using `require()` instead of `import`).

TimothyGu:
The cause for this bug is inconsistent [brand checks](https://es.discourse.group/t/strong-brand-checking-in-javascript/557) between Node.js' implementation of URLSearchParams compared to browser's. Consider the following:

```js
const u = new URLSearchParams();
u.get.call(Object.create(u), '');
```

In the browser (and hence Electron), the third line results in a thrown exception, usually with a message like "Invalid invocation" or something similar. However, Node.js uses a different style of brand checks that [checks](https://github.com/nodejs/node/blob/3b3b02f8bbf99e2a746eb264ef4a0f864346a1a5/lib/internal/url.js#L263-L265) if the object has a special symbol property. Since property accesses fall back to the prototype object in JavaScript, Node.js is fine with this code, even though `Object.create(u)` is not technically a URLSearchParams object. (The blame for this incompatibility could fall on me, as I was the person who implemented much of URLSearchParams in Node.js.)

The v3 node-fetch way of implementing Headers depends on the Node.js behavior. The current code does:

```js
return URLSearchParams.prototype[p].call(
	receiver,
	String(name).toLowerCase(),
	String(value)
);
```

where `receiver` is the Proxy object returned by Headers.constructor. The Node.js URLSearchParams implementation will happily accept this code, since property accesses to `receiver` eventually go to `target`, which _is_ a URLSearchParams object. However, the browser will not, since `receiver` is the Proxy object and not the URLSearchParams object. (This is why replacing `receiver` with `target` works.

As far as this PR goes, I think it is absolutely the right fix. In fact, I encourage people to merge this as soon as possible, since having a popular package like node-fetch depend on the incorrect behavior of Node.js itself prevents Node.js from fixing this bug. If you want to actually create a test for this PR, the best bet is to monkey-patch the Node.js URLSearchParams so that it acts more like the browser:

```js
// Monkey-patch URLSearchParams to do brand checks more accurately.
// This makes the implementation more similar to browsers (or Electron),
// and works around a Node.js bug.

const prevCtor = URLSearchParams;
const uspSet = new WeakSet();
URLSearchParams = function(...args) {
  const obj = Reflect.construct(prevCtor, args, new.target);
  uspSet.add(obj);
  return obj;
};
Object.setPrototypeOf(URLSearchParams, prevCtor);
URLSearchParams.prototype = prevCtor.prototype;
URLSearchParams.prototype.constructor = URLSearchParams;

const prevHas = prevCtor.prototype.has;
URLSearchParams.prototype.has = function (...args) {
  if (!uspSet.has(this)) {
    throw new TypeError('has called on an object that is not a URLSearchParams');
  }
  return Reflect.apply(prevHas, this, args);
};

/////////////////////////////////////////////

const { Headers } = require('node-fetch');
const h = new Headers();
console.log(h.has('abc'));
```

This code fragment throws an error under the current definition of `Headers`, but passes after this PR.

----

That said, I think it's absolutely crazy that Headers is a _subclass_ of URLSearchParams, and that canonical Headers objects are proxies. I have filed #1119 to consider reverting this new implementation.

TimothyGu (approved):
As I mentioned in my comment, IMO testing this change is more trouble than it's worth. LGTM as it is.

JosephusPaye:
Thanks for the details @TimothyGu, very illuminating.

bitinn:
Look like we could have used some expertise in this PR, personally I am convinced by Timothy, but would be good to hear some feedbacks from node.js team members :)

Same regarding #1119 too.

tinovyatkin (approved):

xxczaki (approved):

Issue #1113, opened 2021-03-10
[v3] Add credentials to RequestInit


benmccann:
**Is your feature request related to a problem? Please describe.**

`node-fetch`'s request type does not include `credentials`: https://github.com/node-fetch/node-fetch/blob/cbd4d895767e33491219e40013dd8daa5c7ac024/%40types/index.d.ts#L56

This is a field provided by the web platform: https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials

**Describe the solution you'd like**

Add `credentials` field

**Describe alternatives you've considered**

I could work around it, but would prefer not to have to

**Additional context**

This is used in the upcoming SvelteKit project. The source is private for now, but will be opened up soon and then I can share where exactly this is being used

SvelteKit's predecessor Sapper did something similar though:
https://github.com/sveltejs/sapper/blob/339c417b24e8429d3adc9c9f196bf159a5fce874/runtime/src/server/middleware/get_page_handler.ts#L137

tekwiz:
I do think we should add the property to stay with the spec, but I'm not sure what benefit the option provides for node-fetch since it has no concept of an origin.  Can you elaborate on an example?

benmccann:
We implemented isomorphic fetching (i.e. fetching with the browser's implementation on the client and `node-fetch` on the server) and wanted to surface to the user during SSR if the client-side call would likely fail due to cross-origin fetching, credentials, etc. E.g. we can detect if the user is making a call with only the path specified or if they have a fully specified URL with domain name. You can see the implementation here: https://github.com/sveltejs/sapper/blob/339c417b24e8429d3adc9c9f196bf159a5fce874/runtime/src/server/middleware/get_page_handler.ts#L137

eikaramba:
i think i run into this problem now because i wanted to implement authentication with server based cookies. and if i use the node-fetch implementation of sveltekit to request a ressource on the client it works. the exact same code does not work however on the server. on the previous sapper app it worked with the exact same code.

jimmywarting:
the only thing we could support is the `omit`
the other two (same-site and include) would have the same meaning on backend

So there is a tiny usecase for it... but since we don't have a cookie jar either... then it have even less usefulness

johnnysprinkles:
If you really wanted to achieve parity between client-side and server-side fetch, I think you'd need to pass in the incoming request, do preflighting, and look at access headers to voluntarily block access to things, which goes against the spirit of server-side code being omnipotent.

[...]

PR #1116, opened 2021-03-14
Add credentials field to RequestInit type definition

benmccann:

- [ ] Documentation update
- [X] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Align the request type with the spec

Fixes a type error we're getting in SvelteKit:
https://github.com/sveltejs/kit/blob/5cd6f11aee6cd5edc5a758a5deb88925d3c9a50e/packages/kit/src/runtime/server/page.js#L59

Closes https://github.com/node-fetch/node-fetch/issues/1113

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -75,6 +75,8 @@ interface RequestInit {
 	 */
 	signal?: AbortSignal | null;
 
+	credentials?: 'omit' | 'same-origin' | 'include';
+
 	// Node-fetch extensions to the whatwg/fetch spec
 	agent?: Agent | ((parsedUrl: URL) => Agent);
 	compress?: boolean;


eikaramba:
is there something we can do to speed up this pull request? it currently prevents sveltekit from beeing used in SSR mode for any API where server credentials are necessary.

benmccann:
> it currently prevents sveltekit from beeing used in SSR mode for any API where server credentials are necessary

No it doesn't. I've simply told SvelteKit to ignore the type error from `credentials` missing. If there's some other issue or missing feature it's not blocked by this. That being said, I wouldn't mind getting this change in :smile:

LinusU:
Hmmmm, `credentials` is listed as one of the options that we do not implement.

ref: https://github.com/node-fetch/node-fetch#class-request

Should we really add it to the type definitions if we don't support it? 🤔

jimmywarting:
> Hmmmm, `credentials` is listed as one of the options that we do not implement.
> 
> ref: https://github.com/node-fetch/node-fetch#class-request
> 
> Should we really add it to the type definitions if we don't support it? 🤔

Have kind of forgotten the credentials option. I'm only now looking back at some old PR's that have never gotten resolved or closed.

Seeing as if we just made a security fix of exposing sensitive headers...

Maybe it is about time that we ship this option and honor the `omit` rule? that is something we could do on the backend.

The other two options: `same-origin` & `include` i guess is not something we could solve cuz we have no concept of an origin on the backend. 

- `same-origin` refers to the website you are currently on.
- and `include` has to do with CORS

b/c we do not have any origin. Both the `same-origin` and `include` would have the same meaning on the backend when you make a request to example.com. 

---

a bit of information from mdn:

> The credentials read-only property of the Request interface indicates whether the user agent should send cookies from the other domain in the case of cross-origin requests.
> 
> Syntax
> - `request.credentials` - A RequestCredentials dictionary value indicating whether the user agent should send cookies from the other domain in the case of cross-origin requests. Possible values are:
> 
> - `omit`: Never send or receive cookies.
> - `same-origin`: Send user credentials (cookies, basic http auth, etc..) if the URL is on the same origin as the calling script. This is the default value.
> - `include`: Always send user credentials (cookies, basic http auth, etc..), even for cross-origin calls.

if a node-fetch developer would do: `fetch(url, { credentials: 'omit', ...options })` then we could strip away cookies from the request and remove the set-cookie from the response.

benmccann:
We use undici now

PR #1118, opened 2021-03-17
Fix/form data in response

Gozala:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

1. Added a test case illustrating an issue
2. Addressed a bug caused by missed function argument in `Response` implementation.
3. Addressed issue with form-data body where boundary fragments are emitted as strings while content fragments as binary data. Which caused exceptions.

See test case illustrating the issue. It throws exception without this fix.

1. I had to workaround code coverage problem which reported issue for the line https://coveralls.io/builds/38068042/source?filename=src%2Fbody.js#L350 which is unreachable because `FormData` body is normalized into a stream and I don't believe there are no code paths in which request/response `.body` could be `FormData`.
   
    https://github.com/node-fetch/node-fetch/blob/1780f5ae89107ded4f232f43219ab0e548b0647c/src/body.js#L53-L56

    I do not believe changes here affected code coverage, however instead of removing those lines (which break no tests) I've added a assertion to exercise that code path.

2. Normalizing `FormData` to stream during body initialization has a problem of making content length undetectable: 
    https://github.com/node-fetch/node-fetch/blob/1780f5ae89107ded4f232f43219ab0e548b0647c/src/request.js#L172

    And this code path will never be called

    https://github.com/node-fetch/node-fetch/blob/1780f5ae89107ded4f232f43219ab0e548b0647c/src/body.js#L352-L355

    But I'll defer this to another issue

3. (Last) commit dc5be05 adds bunch of type annotations which helped me to reason through by getting better IntelliSense as opposed to remembering all the ins and outs. Feel free to get rid of it if you don't want them.

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -197,14 +197,15 @@ async function consumeBody(data) {
 
 	try {
 		for await (const chunk of body) {
-			if (data.size > 0 && accumBytes + chunk.length > data.size) {
+			const bytes = typeof chunk === "string" ? Buffer.from(chunk) : chunk
+			if (data.size > 0 && accumBytes + bytes.byteLength > data.size) {
 				const err = new FetchError(`content size at ${data.url} over limit: ${data.size}`, 'max-size');
 				body.destroy(err);
 				throw err;
 			}
 
-			accumBytes += chunk.length;
-			accum.push(chunk);
+			accumBytes += bytes.byteLength;
+			accum.push(bytes);
 		}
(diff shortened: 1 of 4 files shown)


jimmywarting:
superseded by #1314

LinusU:
#1314 should be landed soon which fixes this, thank you for taking the time with this ☺️

PR #1141, opened 2021-05-03
Require Node.js 12.20.0 and move to ESM

xxczaki:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain:

- Converted the module to pure ESM (the minimal Node.js version is now 12.8.0)
- Updated all dependencies & devDependencies
- Removed rollup and commonjs-related things (mainly tests)
- Linted everything with the newest version of xo
- Updated readme, changelog and funding.yml.
- Removed commonjs GitHub action

#668

> Reference material: https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77


@types/index.d.ts
```diff
@@ -176,28 +176,20 @@ declare class AbortError extends Error {
 	[Symbol.toStringTag]: 'AbortError';
 }
 
-
 declare function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
-declare class fetch {
-	static default: typeof fetch;
-}
```
xxczaki, line 183: I removed this class declaration, as it seems like it changes nothing when it comes to the types. Please let me know if I'm wrong here - maybe this was added intentionally due to a bug that is not covered by our types test suite.

@types/index.test-d.ts
```diff
@@ -62,31 +62,24 @@ async function run() {
...
-	expectType<number>(reqRes.size);
-	expectType<number>(reqRes.status);
-	expectType<string>(reqRes.statusText);
-	expectType<() => Response>(reqRes.clone);
+	const wildResponse = await _fetch('https://google.com');
+	expectType<boolean>(wildResponse.ok);
+	expectType<number>(wildResponse.size);
+	expectType<number>(wildResponse.status);
+	expectType<string>(wildResponse.statusText);
+	expectType<() => Response>(wildResponse.clone);
```
xxczaki, line 70: This part fails the tests, meaning TypeScript users cannot import node-fetch like so:

```ts
import * as fetch from 'node-fetch';
```

I tried to fix this issue, but without success. I also wonder whether we really need to test this - are there any benefits from using the aforementioned `import..as` style than just doing:

```ts
```

?

example.js
```diff
@@ -1,4 +1,4 @@
-const fetch = require('node-fetch');
+import fetch from 'node-fetch';
 
 // Plain text or HTML
 (async () => {
```
xxczaki, line 4: Do we want to get rid of these async functions and utilize top-level await?

tinovyatkin:
@xxczaki 
[This logic](https://github.com/node-fetch/node-fetch/blob/ffef5e3c2322e8493dd75120b1123b01b106ab23/.github/workflows/ci.yml#L51-L58) should also be adjusted

@types/index.test-d.ts
dnalborczyk: this seems to be correct. `import * as ...` is importing the module, which is **always** an object.

if you want to use `import * as fetch` you would need to run `fetch.default()` in order to use it.

I believe the reason `import * as fetch from 'node-fetch';` worked in typescript with commonjs is because of module interop: https://www.typescriptlang.org/tsconfig#esModuleInterop

xxczaki: @dnalborczyk Thanks! This explains everything :smile:

example.js
xxczaki: Done!

package.json
```diff
@@ -1,34 +1,25 @@
...
-  },
   "files": [
     "src",
     "dist",
     "@types/index.d.ts"
   ],
   "types": "./@types/index.d.ts",
   "engines": {
-    "node": "^10.17 || >=12.3"
+    "node": ">=12.8"
```
LinusU, line 15: What do you think of the following specifier to not include any versions of Node that doesn't full support ESM?

```suggestion
    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
```

more information in this discussion: https://github.com/sindresorhus/meta/discussions/15#discussioncomment-688195

jimmywarting:
Helped out a bit with.

@tinovyatkin what changes needs to be done to ci.yaml? maybe you can contribute to this PR?

also, i'm no expert at type definitions files - can someone else have a crack at this? (or tell whats needs to be done?)

LinusU:
@jimmywarting I could to the type definition files, do you want me to push directly to this branch or submit a PR to this branch?

jimmywarting:
@LinusU grate thx

> do you want me to push directly to this branch or submit a PR to this branch?

Push directly to this PR (esm branch)

@types/index.test-d.ts
```diff
@@ -95,8 +88,7 @@ async function run() {
 		['b', '2'],
 		new Map([['a', null], ['3', null]]).keys()
 	]);
-
-	fetch.isRedirect = (code: number) => true;
```
LinusU, line 99: This was removed since this is no longer possible. `isRedirect` is now exported with `export function isRedirect` and is thus no longer a property on the `fetch` function...

LinusU:
@jimmywarting done ✅

jimmywarting (approved):

README.md
```diff
@@ -391,40 +385,11 @@ const data = await response.json();
...
-console.log(data)
-```
-
-node-fetch also supports spec-compliant FormData implementations such as [form-data](https://github.com/form-data/form-data) and [formdata-node](https://github.com/octet-stream/form-data):
-
-```js
-const fetch = require('node-fetch');
-const FormData = require('formdata-node');
+import fetch from 'node-fetch';
+import FormData from 'formdata-node';
```
octet-stream, line 392: formdata-node does not have default export

```diff
- import FormData from "formdata-node"
+ import {FormData} from "formdata-node"
```

gr2m:
I was hoping to be able to test your esm rewrite with my @octokit libraries but no luck, I can't get it to work with the current TypeScript/build setup :( I plan to migrate all of @octokit to esm soon, but that'll at least another few more months, it's not a priority yet.

I'll try some other ways to use the new node-fetch, but don't wait on me.

From just looking at the code, the changes look good, I can't find any obvious problems.

package.json
```diff
@@ -1,34 +1,23 @@
...
+  "main": "./src/index.js",
   "sideEffects": false,
   "type": "module",
-  "exports": {
-    ".": {
-      "import": "./src/index.js",
-      "require": "./dist/index.cjs"
-    },
-    "./package.json": "./package.json"
-  },
```
gr2m, line 15: I think the correct thing to do is to remove `"main"`, and to add

```js
  "exports": "./src/index.js"
```

gr2m:
I found a way to test it after all

I found a problem when setting the `authorization` header

```js
const response = await fetch("https://api.github.com/", {
	headers: {
		authorization: "test",
	},
});

await response.json();
```

throws with

> FetchError: Invalid response body while trying to fetch https://api.github.com/: Premature close
>     at consumeBody (file:///Users/gregor/Projects/node-fetch/src/body.js:210:60)
>     at processTicksAndRejections (node:internal/process/task_queues:96:5)
>     at async Response.json (file:///Users/gregor/Projects/node-fetch/src/body.js:119:18)
>     at async file:///Users/gregor/Projects/node-fetch/gregor.js:9:1 {
>   type: 'system',
>   errno: 'ERR_STREAM_PREMATURE_CLOSE',
>   code: 'ERR_STREAM_PREMATURE_CLOSE',
>   erroredSysCall: undefined
> }

But the same error seems to exist in `master`, too. Shall I file a separate issues?

package.json
jimmywarting: Don't think we need exports at all?

gr2m: I’m pretty sure we do, how else would the package consumers know what file they should load when importing?

jimmywarting: From the `main` ofc

fetch-blob@3 is a ESM only package, it dose not include any exports or cjs, only a `main` + `type=module`

gr2m: with type=module, you no longer set `main`, you set `exports`. `main` might still be supported but `exports` is the new way to go, it's going to be expected by lots of tooling out there. When using exports, it will also be easier to define different exports for different environments, such as browser, node, deno, cloudflare workers, etc.

Richienb:
Because of the new `node:` protocol, it may be worth bringing it forwards to `>=12.20`.

jimmywarting:
related: https://github.com/node-fetch/fetch-blob/issues/98

jimmywarting:
@Richienb, @LinusU  could you review this?
looks finish

README.md
```diff
@@ -102,7 +102,7 @@ See Jason Miller's [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic
 
 ## Installation
 
-Current stable release (`3.x`)
+Current stable release (`3.x`) requires at least Node.js 12.8.0.
```
LinusU, line 105:
```suggestion
Current stable release (`3.x`) requires at least Node.js 12.20.0.
```

README.md
LinusU: @jimmywarting should we fix this before merging?

README.md
jimmywarting: Resolved, thanks

README.md
jimmywarting: sure

package.json
LinusU: From [Sindre's guide](https://github.com/sindresorhus/meta/discussions/15), which is the one I've been following in my repos, one should add `"exports": "./src/index.js"`. I don't have any more info on this though, but what @gr2m said seems right to me 🤔

.github/workflows/ci.yml
```diff
@@ -14,52 +14,27 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        node: ["14", "12", engines]
+        node: ["12.22.3", "14.13.1", "16.0.0"]
```
LinusU, line 17: I might have missed this in the other PR, but was this required to get the tests passing? If we indeed require 12.22.3 instead of 12.20.0 we need to update `package.json -> engines.node` and the readme as well. Otherwise we can move this back down to `12.20.0`.

@types/index.d.ts
```diff
@@ -1,19 +1,16 @@
 /// <reference types="node" />
+/// <reference lib="dom" />
```
LinusU, line 2: I've seen several issues when using both `node` and `dom` at the same time due to conflicting declarations. What was the reasoning for importing `URL`, `URLSearchParams`, and `Blob` from `dom` rather than `url` and `fetch-blob`?

Maybe the issues have been fixed in recent versions of the `node` and `dom` definitions though 🤔

README.md
```diff
@@ -389,7 +389,8 @@ node-fetch also supports spec-compliant FormData implementations such as [formda
 
 ```js
-import FormData from 'formdata-node';
+import {FormData} from 'formdata-polyfill/esm-min.js';
+import {FormData} from 'formdata-node';
```
LinusU, line 393: Should both of these be here? 🤔

.github/workflows/ci.yml
jimmywarting: I think we can revert this back to 12.22.0

@types/index.d.ts
jimmywarting: Um, I changed fetch-blob to be 100% spec compatible by introducing whatwg streams
i though you could just use the DOM blob cuz they are now indistinguishable

sry, but i'm no expert at TypeScript. I only use jsdoc for typing

README.md
jimmywarting: They should both work, you can pick one or the other

.github/workflows/ci.yml
jimmywarting: looks like the test passed.

@types/index.d.ts
jimmywarting: Is it possible to use a namespace like `NodeJS.Buffer.Blob` or something like it?

package.json
jimmywarting: I thought exports was just like some kind of [import map](https://github.com/WICG/import-maps)

but when i read upon it it says:

> In a package’s package.json file, two fields can define entry points for a package: "main" and "exports". The "main" field is supported in all versions of Node.js, but its capabilities are limited: it only defines the main entry point of the package.
> 
> The "exports" field provides an alternative to "main" where the package main entry point can be defined while also encapsulating the package, preventing any other entry points besides those defined in "exports". This encapsulation allows module authors to define a public interface for their package.
> 
> If both "exports" and "main" are defined, the "exports" field takes precedence over "main". "exports" are not specific to ES modules or CommonJS; "main" is overridden by "exports" if it exists. As such "main" cannot be used as a fallback for CommonJS but it can be used as a fallback for legacy versions of Node.js that do not support the "exports" field.

ref: https://nodejs.org/api/packages.html#packages_package_entry_points

So... if you have a exports than you can't simply just use `import('node-fetch/src/response.js')` what if ppl want to do that? it sounds like `main` is sufficient

jimmywarting (approved):

LinusU (approved):
Great work everyone! 🚀

@types/index.d.ts
LinusU: Hmm, I'm not sure about this one actually... Maybe try it like this in the beta.10 and then we can change it if we encounter any issues?

package.json
gr2m: > So... if you have a exports than you can't simply just use `import('node-fetch/src/response.js')`? or i'm wrong?
> what if ppl want to do that? it sounds like `main` is sufficient..?

Importing `node-fetch/src/response.js` is using an internal API that could break any time because it's not tested. If we wanted to make it possible to import the `Response` class selectively we could do

```js
  "exports": {
    ".": "./src/index.js",
    "./response": "./src/response.js"
  },
```

And we should cover the import in our tests.

I haven't seen an ESM package which still uses `"main"`, it feels odd to migrate to `esm` but not use the `exports` key. Did you test that the import of `node-fetch` is working for you? I'll try to give it a try myself

test/main.js
```diff
@@ -1897,18 +1899,28 @@ describe('node-fetch', () => {
 	});
 
 	it('should not timeout on cloning response without consuming one of the streams when the second packet size is less than default highWaterMark', function () {
+		// TODO: fix test.
+		if (!isNodeLowerThan('v16.0.0')) {
+			this.skip();
+		}
```
gr2m, line 1905: do we need to create a follow up issue for the free `// TODO:` comments?

test/main.js
```diff
@@ -1897,18 +1899,28 @@ describe('node-fetch', () => {
...
+
 		this.timeout(300);
 		const url = local.mockResponse(res => {
 			const firstPacketMaxSize = 65438;
 			const secondPacketSize = 16 * 1024; // = defaultHighWaterMark
 			res.end(crypto.randomBytes(firstPacketMaxSize + secondPacketSize - 1));
 		});
 		return expect(
-			fetch(url).then(res => res.clone().buffer())
+			fetch(url).then(res => res.clone().buffer()).then(console.log)
```
gr2m, line 1914: was the `.then(console.log)` added for debugging?
```suggestion
			fetch(url).then(res => res.clone().buffer())
```

jimmywarting: Probably right. Should remove it

test/main.js
jimmywarting: That is a Good idea

jimmywarting:
@gr2m 
> Did you test that the import of node-fetch is working for you? I'll try to give it a try myself

No i haven't tested node-fetch but i have tried out formdata-polyfill and fetch-blob that is a ESM only pacakge. They are ESM only package and only have a `main`

test/main.js
jimmywarting: #1210

gr2m:
I won't die on my `exports` with explicitly defined exports hill 😁 If you feel strong about leaving `main` let's do that and see how it goes. It's a problem when it's a problem, we shall find out.

jimmywarting:
I don't have so strong feelings for it either. (even doe it might sound like it)
but yea, "It's a problem when it's a problem" then we will fix it

PR #1150, opened 2021-05-07
stream in formdata entries is abnormal

jimmywarting:
Changed one spec compatible formdata test

Now depends on #1141 to be solved first


octet-stream:
Would be nice if you update formdata-node version to 3.x :)
It will also fix #1151 then.

jimmywarting:
superseeded by the ESM PM

PR #1170, opened 2021-05-25
Test on all minimum supported Node.js versions

LinusU:

- [x] Other, please explain: CI update for ESM

When switching to ESM in #1141 we are introducing three new minimum Node.js versions that we support:

- `12.20.0`
- `14.13.1`
- `16.0.0`

In order to not accidentally break any of those versions I propose that we test on them all.

This was previously sort of done with the `minimum-node-version` but it only supports a single version, so would thus only run on `12.20.0`.

n/a

n/a

-----

ping @jimmywarting: I didn't push this directly to the `esm` branch since this might something we want to discuss first?

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -1895,10 +1895,12 @@ describe('node-fetch', () => {
 		).to.timeout;
 	});
 
-	const nodeVersion = +process.version.substring(1, process.version.indexOf('.'));
+	const nodeVersion = Number(process.version.slice(1, process.version.indexOf('.')));
 	it('should not timeout on cloning response without consuming one of the streams when the second packet size is less than default highWaterMark', function () {
 		// TODO: fix test.
-		if (nodeVersion >= 16) this.skip();
+		if (nodeVersion >= 16) {
+			this.skip();
+		}
 
 		this.timeout(300);
 		const url = local.mockResponse(res => {
@@ -1913,7 +1915,9 @@ describe('node-fetch', () => {
 
 	it('should not timeout on cloning response without consuming one of the streams when the second packet size is less than custom highWaterMark', function () {
 		// TODO: fix test.
-		if (nodeVersion >= 16) this.skip();
+		if (nodeVersion >= 16) {
+			this.skip();
+		}
 
 		this.timeout(300);
 		const url = local.mockResponse(res => {
@@ -1928,7 +1932,9 @@ describe('node-fetch', () => {
 
 	it('should not timeout on cloning response without consuming one of the streams when the response size is double the custom large highWaterMark - 1', function () {
 		// TODO: fix test.
-		if (nodeVersion >= 16) this.skip();
+		if (nodeVersion >= 16) {
+			this.skip();
+		}
 
 		this.timeout(300);
 		const url = local.mockResponse(res => {
--- a/test/main.js
+++ b/test/main.js
@@ -634,9 +634,9 @@ describe('node-fetch', () => {
 			const read = async body => {
 				const chunks = [];
 
-				if (process.version < 'v14') {
-					// In Node.js 12, some errors don't come out in the async iterator; we have to pick
-					// them up from the event-emitter and then throw them after the async iterator
+				if (process.version < 'v14.15.2') {
+					// In older Node.js versions, some errors don't come out in the async iterator; we have
+					// to pick them up from the event-emitter and then throw them after the async iterator
 					let error;
 					body.on('error', err => {
 						error = err;
--- a/test/main.js
+++ b/test/main.js
@@ -38,11 +38,7 @@ import chaiTimeout from './utils/chai-timeout.js';
 const AbortControllerPolyfill = abortControllerPolyfill.AbortController;
 
 function isNodeLowerThan(version) {
-	return !~process.version.localeCompare(version, undefined, { numeric: true })
-}
-
-function isNodeHigherThan(version) {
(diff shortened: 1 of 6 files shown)


LinusU:
Seems like this uncovered some issues that needs to be fixed:

- Node.js 14.13.1 doesn't seem to have the fix checked for here:
  
  ```js
  if (process.version < 'v14') {
  ```

  > should handle network-error in chunked response async iterator:
  >  AssertionError: expected promise to be rejected with 'Error' but it was fulfilled with [ Array(2) ]

- Node.js 16.x seems to have changed error message:
  > AssertionError: expected 'Invalid response body while trying to fetch http://localhost:34163/error/premature: aborted' to match /Premature close|The operation was aborted/

- Node.js 16.x seems to have changed something with highWaterMark
  > should not timeout on cloning response without consuming one of the streams when the second packet size is less than default highWaterMark:
  >   AssertionError: expected promise not to timeout but it timed out
  >
  > should not timeout on cloning response without consuming one of the streams when the second packet size is less than custom highWaterMark:
  >  AssertionError: expected promise not to timeout but it timed out
  >
  > should not timeout on cloning response without consuming one of the streams when the response size is double the custom large highWaterMark - 1:
  >  AssertionError: expected promise not to timeout but it timed out

jimmywarting:
> edit: I'm not sure about the last one, would love some input!

I don't know either,  but like the looks of this PR

LinusU:
@xxczaki it seems like you originally did the high watermark implementation (#671). Would you be able to take a quick look at the failure on Node.js 16 and see if you know what's wrong? Thanks!

jimmywarting:
Don't know where all the old maintainers/contributors have gone... vacation maybe? 
Speculating wether or not to just disable the test for the response cloning and expect things to fail again if you clone the response without using the other stream

Want to push for the v3 and ESM out for release soonish

Also speculating if WeakRef/Finalizers could help to clean up un-used pipe/stream if it's necessary for one of the stream to finish. It seems to be the major issue for developers that have that problem. That it won't finish the 2nd stream if the 1st one is buffered/filled and forgotten/unused. It was also the reason why setting a none standard highWatherMark was added... Ideally i would like to get rid of the `highWatherMark` option if we can to stay closer to the spec

jimmywarting:
looking back at old PR's it seems like @davesidious has a solution/pr that i have overlooked that solves highWaterMark #1162 ?

Don't know if it's related to node v16 or just some old mistake we have done in v3

LinusU:
@jimmywarting 👍

test/main.js
```diff
@@ -634,9 +634,9 @@ describe('node-fetch', () => {
 			const read = async body => {
 				const chunks = [];
 
-				if (process.version < 'v14') {
-					// In Node.js 12, some errors don't come out in the async iterator; we have to pick
-					// them up from the event-emitter and then throw them after the async iterator
+				if (process.version < 'v14.15.2') {
```
jimmywarting, line 637: Umm, when have it ever been possible to compare strings like this?

jimmywarting:
@LinusU  Yea, trying to figure that ubuntu bug one out... got the same issue in fetch-blob too

jimmywarting:
And now there where only `Unexpected token '.'` left
the code seems legit/valid and runs fine localy... wonder if it has something to do with ESM.

```
SyntaxError[ @/home/runner/work/node-fetch/node-fetch/test/external-encoding.js ]: Unexpected token '.'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:140:18)
```

jimmywarting:
@LinusU do you think it requires a higher node version?

jimmywarting:
ok... so it complains on v12.x.x not on v14.x.x

what should we do now? suggestoins?

test/main.js
LinusU: Well, it will compare lexicographically, so as long as Node.js doesn't release a version `14.100.0` or `100.0.0` I think it should work...

But we could use e.g. [`semver-compare`](https://www.npmjs.com/package/semver-compare) instead ☺️

jimmywarting: > But we could use e.g. semver-compare instead ☺️

like fewer dependencies. found this on https://stackoverflow.com/a/55466325/1008999

jimmywarting:
Think i figure it out. Somehow fetch-blob must have slipped in with a optional changing and used fs/promise
this dose not work in node v12...
i have degraded them in v3.1.0 and also added real whatwg stream polyfill to blob#stream (since node is going to make whatwg happen - they have already shipped it as experimental) 

problem is... fetch-blob it's a ESM only package now. maybe need to backport the `fs/promise` & `?.` to fetch-blob@v2

jimmywarting:
Found what was wrong with fetch-blob and fixed it. it's released now.

Dose the changes looks good to be merged @LinusU ? I'm rdy to merge

jimmywarting (approved):

test/main.js
```diff
@@ -37,6 +37,10 @@ import chaiTimeout from './utils/chai-timeout.js';
 
 const AbortControllerPolyfill = abortControllerPolyfill.AbortController;
 
+function isNodeLowerThan(version) {
+	return !~process.version.localeCompare(version, undefined, {numeric: true});
+}
```
LinusU, line 42: Neat! 👍

LinusU:
Looks great! Thanks for landing this 🚀

PR #1064, opened 2021-01-10
Fix premature close with chunked transfer encoding and for async iterators in Node 12

tekwiz:
**What is the purpose of this pull request?**

- [x] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

**What changes did you make? (provide an overview)**

re. 47b00e179dd8a450f3fffd9ac22d148197a09229: For responses using chunked `Transfer-Encoding` without a `Content-Length`, this adds a check at the socket's `close` event to verify that the last bytes received were the final chunk bytes, i.e. `0\r\n`.  If not, it creates a `Premature close` error and sends it to `response.body.destroy()`.  Certain network errors evidently bypass the HTTP parser's error handling resulting in silent failures.  @davidje13 confirmed fix in #739 https://github.com/node-fetch/node-fetch/issues/739#issuecomment-754173198.

re. 13cdcf3d2fa514b76d60b5e73b5a8602e7d336b7: Before Node.js 14, pipeline() does not fully support async iterators and does not always properly handle when the socket close/end events are out of order. This fix emits a `Premature close` error if the socket's readable `end` event occurs before the writable `close` event.

re. 9cf8c1820e4f72739b8921cc7cec6e2f0e8ccfae: Usage of async iterators for both Node.js 12 and 14 is added to the README.

**Which issue (if any) does this pull request address?**

#739 #766 #968 #1055 #1056 and possibly #736

**Is there anything you'd like reviewers to know?**

PR #1172, opened 2021-05-27
fix: premature close with chunked transfer encoding and for async iterators in Node 12

achingbrain:
This PR backports the fix from #1064 to the `2.x.x` branch following the [comment here](https://github.com/node-fetch/node-fetch/pull/1064#issuecomment-849167400).

I had to add some extra babel config to allow using the `for await..of` syntax in the tests.  The config is only needed for the tests as this syntax is not used in the implementation.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -299,7 +299,6 @@ export default function fetch(url, opts) {
 };
 
 function fixResponseChunkedTransferBadEnding(request, errorCallback) {
-	const LAST_CHUNK = Buffer.from('0\r\n');
 	let socket;
 
 	request.on('socket', s => {
@@ -309,14 +308,11 @@ function fixResponseChunkedTransferBadEnding(request, errorCallback) {
 	request.on('response', response => {
 		const {headers} = response;
 		if (headers['transfer-encoding'] === 'chunked' && !headers['content-length']) {
-			let properLastChunkReceived = false;
-
-			socket.on('data', buf => {
-				properLastChunkReceived = Buffer.compare(buf.slice(-3), LAST_CHUNK) === 0;
-			});
-
 			socket.addListener('close', () => {
-				if (!properLastChunkReceived) {
+				// if a data listener is still present we didn't end cleanly
+				const hasDataListener = socket.listenerCount('data') > 0;
+
+				if (hasDataListener) {
 					const err = new Error('Premature close');
 					err.code = 'ERR_STREAM_PREMATURE_CLOSE';
 					errorCallback(err);
--- a/src/index.js
+++ b/src/index.js
@@ -307,7 +307,7 @@ function fixResponseChunkedTransferBadEnding(request, errorCallback) {
 	request.on('response', response => {
 		const {headers} = response;
 		if (headers['transfer-encoding'] === 'chunked' && !headers['content-length']) {
-			socket.addListener('close', hadError => {
+			response.once('close', hadError => {
 				// if a data listener is still present we didn't end cleanly
 				const hasDataListener = socket.listenerCount('data') > 0;
 
--- a/src/index.js
+++ b/src/index.js
@@ -54,7 +54,7 @@ export default function fetch(url, opts) {
 			let error = new AbortError('The user aborted a request.');
 			reject(error);
 			if (request.body && request.body instanceof Stream.Readable) {
-				request.body.destroy(error);
+				destroyStream(request.body, error);
 			}
 			if (!response || !response.body) return;
 			response.body.emit('error', error);
@@ -95,6 +95,11 @@ export default function fetch(url, opts) {
 
 		req.on('error', err => {
 			reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
+
+			if (response && response.body) {
+				destroyStream(response.body, err);
+			}
+
 			finalize();
 		});
 
@@ -103,13 +108,7 @@ export default function fetch(url, opts) {
 				return
 			}
 
-			if (response.body.destroy) {
-				response.body.destroy(err);
-			} else {
-				// node < 8
-				response.body.emit('error', err);
-				response.body.end();
-			}
+			destroyStream(response.body, err);
 		});
 
 		/* c8 ignore next 18 */
@@ -308,11 +307,11 @@ function fixResponseChunkedTransferBadEnding(request, errorCallback) {
 	request.on('response', response => {
 		const {headers} = response;
 		if (headers['transfer-encoding'] === 'chunked' && !headers['content-length']) {
-			socket.addListener('close', () => {
+			socket.addListener('close', hadError => {
 				// if a data listener is still present we didn't end cleanly
 				const hasDataListener = socket.listenerCount('data') > 0;
 
-				if (hasDataListener) {
+				if (hasDataListener && !hadError) {
 					const err = new Error('Premature close');
 					err.code = 'ERR_STREAM_PREMATURE_CLOSE';
 					errorCallback(err);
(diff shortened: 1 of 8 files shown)


test/test.js
```diff
@@ -1015,7 +1086,7 @@ describe('node-fetch', () => {
 		))
 			.to.eventually.be.fulfilled
 			.then((res) => {
-				res.body.on('error', (err) => {
+				res.body.once('error', (err) => {
```
achingbrain, line 1089: This is a bit weird - the error event gets emitted on the response body twice in node 14 (and 12 with this PR) - once with the expected `AbortError` and then again with `ERR_STREAM_PREMATURE_CLOSE`.

Both errors were still emitted even after changing how the `AbortError` is emitted (e.g. using `.destroy(err)` instead of `.emit('error', err)` so I'm open to suggestions on how to not need to change `.on` to `.once` here, if it's an issue.

achingbrain:
Ok, I hadn't realised quite how far back in time the supported node versions would go, will see if I can fix up some the oldies.

src/index.js
```diff
@@ -265,6 +294,34 @@ export default function fetch(url, opts) {
...
+	request.on('response', response => {
+		const {headers} = response;
+		if (headers['transfer-encoding'] === 'chunked' && !headers['content-length']) {
+			let properLastChunkReceived = false;
+
+			socket.on('data', buf => {
+				properLastChunkReceived = Buffer.compare(buf.slice(-3), LAST_CHUNK) === 0;
+			});
+
+			socket.addListener('close', () => {
```
achingbrain, line 314: In #1064 this was `socket.prependListener` but that's not supported in node 4.

src/index.js
```diff
@@ -98,6 +98,35 @@ export default function fetch(url, opts) {
...
+				response.body.end();
+			}
+		});
+
+		/* c8 ignore next 18 */
+		if (parseInt(process.version.substring(1)) < 14) {
+			// Before Node.js 14, pipeline() does not fully support async iterators and does not always
+			// properly handle when the socket close/end events are out of order.
+			req.on('socket', s => {
+				s.addListener('close', hadError => {
```
achingbrain, line 116: In #1064 this was `s.prependListener` but that's not supported in node 4.

src/index.js
```diff
@@ -98,6 +98,35 @@ export default function fetch(url, opts) {
...
+				response.body.destroy(err);
+			} else {
+				// node < 8
+				response.body.emit('error', err);
+				response.body.end();
+			}
+		});
+
+		/* c8 ignore next 18 */
+		if (parseInt(process.version.substring(1)) < 14) {
```
achingbrain, line 112: In #1064 this was `if (process.version < 'v14')` but it fails for node 8.

src/index.js
```diff
@@ -98,6 +98,35 @@ export default function fetch(url, opts) {
...
+			// Before Node.js 14, pipeline() does not fully support async iterators and does not always
+			// properly handle when the socket close/end events are out of order.
+			req.on('socket', s => {
+				s.addListener('close', hadError => {
+					// if a data listener is still present we didn't end cleanly
+					const hasDataListener = s.listenerCount('data') > 0
```
achingbrain, line 118: #1064 used the count of types of events with listeners registered on the socket to determine if the stream was closed prematurely but it's not reliable on node 8 and below, so instead if we've got a listener for `'data'` events we're still expecting data to arrive so use that to work out if we should emit the `ERR_STREAM_PREMATURE_CLOSE` error.

achingbrain:
I had to pin a few deps that have dropped compatibility with old node versions without releasing majors and I've added notes where this PR diverges from the fix in #1064 for the sake of compatibility with old node versions but this should be good for review.

test/test.js
gr2m: not sure about this, let's see what they other maintainers think, otherwise A++ pull request, thanks!

gr2m (approved):

test/test.js
achingbrain: I've refactored this away - now the implementation will check to see if the request has been aborted before emitting the `ERR_STREAM_PREMATURE_CLOSE` errors.

CI also now tests on all stable node versions 4-16 for a bit more confidence.

gr2m (approved):
thanks, looks good now!

I do not have access to the npm package myself, so someone else from @node-fetch/core needs to merge and publish the new version

src/index.js
```diff
@@ -265,6 +298,30 @@ export default function fetch(url, opts) {
...
+	request.on('socket', s => {
+		socket = s;
+	});
+
+	request.on('response', response => {
+		const {headers} = response;
+		if (headers['transfer-encoding'] === 'chunked' && !headers['content-length']) {
+			socket.addListener('close', () => {
+				// if a data listener is still present we didn't end cleanly
+				const hasDataListener = socket.listenerCount('data') > 0;
```
achingbrain, line 313: In #1064 this checks the incoming bytes to make sure it ends with `0\r\n` which signifies the last chunk is received.  I found this didn't work well against real-world servers as it assumes the `'data'` event would align with incoming chunks whereas I don't think there's any such guarantee.

I've reused the same check as the [patch for node 10-12 above](https://github.com/node-fetch/node-fetch/pull/1172/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R122) which seems to work well.

oed:
Hey @xxczaki @bitinn @TimothyGu! Can one of you please have a look an release this? My team is facing huge issues that should be fixed with this patch backported to 2.x.x. Should only take a few minutes of your time!

src/index.js
tekwiz: @achingbrain The data event doesn't have to align; this check is only applicable at the very end when the socket is closed and the last bytes of data are pushed out to the data event.  The only way for a chunked transfer to indicate "the end" is for `0\r\n` to be the last 3 bytes at this point, which is what the check was for.

glcohen:
bumping, @jimmywarting @achingbrain can this be merged?

tekwiz (approved):

tekwiz:
@jimmywarting I'm good with merging this into 2.x, but I feel it's significant enough to warrant a minor version bump to 2.7.0. I'll leave it to you to merge so you can make the determination on versioning.

IvanDeluxe:
Hey :)

Is there a way I could use this even though it's not merged .. or can we just merge it? 🔥

aphelionz:
friendly bump :)

achingbrain:
Thanks so much for merging this. I don't think it's been released yet, is there any chance of getting it out of the door?

gr2m:
@jimmywarting we don't have semantic-release setup in the `2.x` branch, we need to release manually, or setup semantic-release in there as well. Let me know if you'd like help with the latter

jimmywarting:
@gr2m if you could set it up, then it would be great

achingbrain:
@gr2m @jimmywarting just a gentle reminder this still hasn't made it into a release.  Please could the 2.x.x branch have a release with this in it?

achingbrain:
@gr2m @jimmywarting please, please, please can this be released?

gr2m:
hey sorry I'll look into it

gr2m:
there y'all go. Can you confirm it looks good?

gr2m:
@node-fetch/core I've setup automated releases in the 2.x branch, so backporting should be easy now, just send PRs against the 2.x branch and use [the commit conventions](https://github.com/node-fetch/node-fetch/blob/main/CONTRIBUTING.md#maintainers-only)

achingbrain:
Thankyouthankyouthankyouthankyou!

src/index.js
steveluscher: This has caused problems when reusing sockets for multiple requests, as when you use an `https.Agent` in ‘keep-alive’ mode.

https://github.com/node-fetch/node-fetch/issues/1767

PR #1197, opened 2021-06-27
Specify possible HTTP request methods in types

ImRodry:

- [X] Documentation update (improving types)
- [ ] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Added a "Methods" type to index.d.ts that covers all the HTTP request methods supported by node-fetch

There didn't seem to be any open issues regarding this

I know that the method parameter is case insensitive but it's common practice to write HTTP methods in upper case, which is why I only added those, let me know if you want me to change that though.

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -53,6 +53,8 @@ declare class Headers {
 	raw(): Record<string, string[]>;
 }
 
+type Methods = "GET" | "HEAD" | "POST" | "PUT" | "DELETE" | "OPTIONS" | "PATCH"
+
 interface RequestInit {
 	/**
 	 * A BodyInit object or null to set request's body.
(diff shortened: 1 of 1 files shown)


jimmywarting:
Any HTTP method is possible, it's not restricted to just a few subsets

ImRodry:
> Any HTTP method is possible, it's not restricted to just a few subsets

Then I can add more but the type string is objectively wrong because not any string is supported

jimmywarting:
> Then I can add more but the type string is objectively wrong because not any string is supported

That is where you are wrong... Spec says:

> There are no restrictions on methods. `CHICKEN` is perfectly acceptable (and not a misspelling of `CHECKIN`). Other than those that are normalized there are no casing restrictions either. `Egg` or `eGg` would be fine, though uppercase is encouraged for consistency.

https://fetch.spec.whatwg.org/#methods

jimmywarting:
There are api that uses SEARCH for rest APIs

jimmywarting:
There are only this  `CONNECT`, `TRACE`, or `TRACK` forbidden methods, but we live on the backend and isn't restricted of what browser puts on us

PR #1212, opened 2021-07-18
fix(Body): Discurage form-data and buffer()

### LINKED ISSUE #1167: Drop support for Form-Data?
[form-data](https://www.npmjs.com/package/form-data) isn't really spec compatible and it dose some abnormal stuff. I know many other libraries support it. But it increases the complexity to anyone who is building a http library if they decide to support it. They (we) need to learn how it works internally of how it can be read & streamed

I would like to emphasize that it's better to follow the standards instead. Any spec'ed FormData impl shouldn't need any documentation.

There are two other spec compatible FormData packages on npm that works grate on NodeJS
1. There is my own [formdata-polyfill](https://www.npmjs.com/package/formdata-polyfill)
2. And also [formdata-node](https://www.npmjs.com/package/formdata-node)

Both of wish can support appending Blob/Files backed up by the FileSystem (and also blob-like items). Unlike form-data that can only append a one-time readable streams. A FormData should be re-useable if you want to clone the body for whatever reason. (our tests cases could use the same FormData instances - but this dose not work with form-data)

This means that all other http libraries out there needs to serialize the body themself, But fear not, there are packages out there also that can convert them

- My formdata-polyfill have a utility function to convert any spec compatible FormData into a Blob so that can be posted instead `formDataToBlob(formdata)`
- octet-stream have built a [FormData Encoder](https://github.com/octet-stream/form-data-encoder) 
- And we also have a encoder built in now into node-fetch itself to convert any spec compatible FormData.

form-data have had a long time to address the read/modify parts but never got around to implement any those methods. That's b/c it's based on appending all entries directly into a readable stream, so they can't iterated over or modified them.

So my proposal really is to:
1. Stop supporting form-data
2. Throw a error or warning message that form-data is no longer supported and also mention alternative packages
3. Remove all hacks where we utilize  `form.getHeaders(), getBoundary(), getLength(), hasKnownLength()` - A FormData size should always be known/calculated. 

---

So what do you think? I would rather want this to be moved into a discussion but this repo haven't enabled it (and i can't do that)

jimmywarting:
## Purpose
- We do serialize spec compatible FormData ourself now. 
- We like to reduce the complexity of the core by removing support for form-data. 
- form-data have had a long time to become spec compliant but haven't done that
- using `res.buffer` is no good as it's a nodejs only feature, if ppl depend on this it only makes it more node-specific
we now strive to more spec compatible ways and more cross platform solutions. 

## Changes
This will output a one time warning when using form-data
image

___

- [x] Readme recommends 2 other spec compatible alternatives already
- [x] The onetime warning references the issue at hand to resolve this.

___

- Fixes #1167

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -6,7 +6,7 @@
  */
 
 import Stream, {PassThrough} from 'stream';
-import {types} from 'util';
+import {types, deprecate} from 'util';
 
 import Blob from 'fetch-blob';
 
@@ -140,6 +140,8 @@ export default class Body {
 	}
 }
 
+Body.prototype.buffer = deprecate(Body.prototype.buffer, 'Please use \'response.arrayBuffer()\' instead of \'response.buffer()\'', 'node-fetch#buffer');
+
 // In browsers, all properties are enumerable.
 Object.defineProperties(Body.prototype, {
 	body: {enumerable: true},
@@ -259,7 +261,11 @@ export const clone = (instance, highWaterMark) => {
 	return body;
 };
 
-let warnedUsingOldFormData = false;
+const getNonSpecFormDataBoundary = deprecate(
+	body => body.getBoundary(),
+	'form-data doesn\'t follow the spec and requires special treatment. Use alternative package',
+	'https://github.com/node-fetch/node-fetch/issues/1167'
+);
 
 /**
  * Performs the operation "extract a `Content-Type` value from |object|" as
@@ -303,17 +309,7 @@ export const extractContentType = (body, request) => {
 
 	// Detect form data input from form-data module
 	if (body && typeof body.getBoundary === 'function') {
-		if (!warnedUsingOldFormData) {
-			console.warn('\u001B[33m%s\u001B[0m', '[WARN]', `
-	You are probably using form-data.
-	form-data doesn't follow the spec and requires special treatment.
-	We will eventually remove support for form-data.
-	See alternatives https://github.com/node-fetch/node-fetch/issues/1167
-			`);
-			warnedUsingOldFormData = true;
-		}
-
-		return `multipart/form-data;boundary=${body.getBoundary()}`;
+		return `multipart/form-data;boundary=${getNonSpecFormDataBoundary(body)}`;
 	}
 
 	// Body is stream - can't really do much about this
--- a/src/body.js
+++ b/src/body.js
@@ -297,6 +297,11 @@ export const extractContentType = (body, request) => {
 		return null;
 	}
 
+
+	if (isFormData(body)) {
+		return `multipart/form-data; boundary=${request[INTERNALS].boundary}`;
+	}
+
 	// Detect form data input from form-data module
 	if (body && typeof body.getBoundary === 'function') {
 		if (!warnedUsingOldFormData) {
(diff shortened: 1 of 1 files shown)


jimmywarting:
Would like to add this message also if that is alright

image

NodeJS plans to impl fetch into core. guess they will be more spec compatible. this will just be a preparations for it + that it makes it more cross compatible

src/body.js
```diff
@@ -295,13 +297,23 @@ export const extractContentType = (body, request) => {
...
+	if (isFormData(body)) {
+		return `multipart/form-data; boundary=${request[INTERNALS].boundary}`;
+	}
+
 	// Detect form data input from form-data module
 	if (body && typeof body.getBoundary === 'function') {
-		return `multipart/form-data;boundary=${body.getBoundary()}`;
-	}
+		if (!warnedUsingOldFormData) {
+			console.warn('\u001B[33m%s\u001B[0m', '[WARN]', `
```
octet-stream, line 307: I think, you can use [`util.deprecate`](https://nodejs.org/dist/latest-v16.x/docs/api/util.html#util_util_deprecate_fn_msg_code) here if this is deprecation warning

jimmywarting: When i googled it it seem like they wanted to deprecate util.deprecate

octet-stream: It looks like they didn't. See: https://github.com/nodejs/node/pull/12631

jimmywarting: I saw that

Richienb:
Since node-fetch v3 bumps a major version, it would make sense to just remove things before a stable version ships if need be.

jimmywarting:
I'm fine with removing form-data and buffer right now before shipping v3 if anyone else agree

octet-stream:
I think that removing of the form-data package right away will have a huge impact and broke code for many developers. I think we need some transition period before its support will be gone for good.

jimmywarting:
> I think that removing of the form-data package right away will have a huge impact and broke code for many developers. Perhaps it would be preferable to have some transition period before its support will be gone for good.

Where thinking quite the same thing when i made this PR.

plans for v4 could be:
- impl whatwg streams body
- remove form-data
- remove `buffer()`

Richienb:
In which case, we could begin handling implementing new features in v3 and deprecate old ones to be removed in v4.

LinusU (approved):
👍

jimmywarting:
@gr2m can you review this also?

gr2m (approved):

kandaris:
So what would an alternative look like for the code below since form-data has been Deprecated?
code

jimmywarting:
`node-fetch` have opted to use [formdata-polyfill](https://www.npmjs.com/package/formdata-polyfill) to deserializer bodies to a formdata using `request.formData()` or `response.formData()`

dos i would recommend using that one as well to avoid any other similar pacakges `npm i formdata-polyfill fetch-blob`

```js
import { fileFromSync } from 'fetch-blob/from.js'
import { FormData } from 'formdata-polyfill/esm.min.js'

const fd = new FormData()
const file = fileFromSync('./package.json')

fd.append('upload', file)
```

There are a way to get it without installing it if you do:
```js
const FormData = (await new Response(new URLSearchParams).formData()).constructor
```
this way you will always depend on same version as we do.

Issue #1249, opened 2021-08-19
Fetch blocks Content-Length: 0 with GET request


bdw429s:
This is similar to https://github.com/node-fetch/node-fetch/pull/795 but different.  795 is asking to actually send a body with a GET request.  I've noticed that sending a GET request even with NO body is still blocked if there is a Content-Length header present, _even if the length is set to 0_.  i.e.
```
Content-Length: 0
```
I've come across a handful of older HTTP clients who do this just as a matter of course and it causes issues specifically in CloudFlare workers were developers are proxying the original request through the Fetch API.  The following error is thrown:
```
Request with a GET or HEAD method cannot have a body.
```
And just to be very clear-- there is _no actual body in my case_.  There is only a `Content-Length` header of 0.

https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2 says 

>  A user agent SHOULD NOT send a Content-Length header field when the request message does not contain  a payload body and the method semantics do not anticipate such a body.

The key words here are "SHOULD NOT".   Per https://datatracker.ietf.org/doc/html/rfc2119 "SHOULD NOT" means the behavior is not recommended, but is not striclty prohibited by the spec.  And since the spec does not say "MUST NOT", libraries such as the Ruby httprb Gem refuse to change the behavior of their code as they claim having a content-length header is not strictly forbidden in the spec.  See the thread in this ticket here:

https://github.com/httprb/http/issues/487

My suggestion is that an content-length of 0 should simply be ignored for GET requests as the mere inclusion of the header is not in violation of the spec.  I'm fine if actual request bodies are still rejected.

I am not a direct user of either of these libraries, but a stranded third party trying to work through this issue for context:

https://meta.discourse.org/t/link-preview-http-get-breaks-spec/200996

jimmywarting:
Hmm, I tried:
```js
fetch('https://httpbin.org/get', {
  headers: {
    'content-length': '2'
  }
}).then(r => r.json()).then(j => console.log(j.headers))
```

In Chrome, Safari & FF: no `content-length: 2` was sent
Deno & node-fetch: `content-length: 2` was sent

Not sure if undici is sending content-length also (they could have the same issue). 
looks almost like if undici may land into nodejs core one day as the default fetch api for NodeJS
can ping @ronag just in case if they have this issue.

Would be grate if there also was a WPT test case for this also so undici, node-fetch Deno and NodeJS can pick this up.
[...]

PR #1268, opened 2021-09-02
fix: Pass url string to http.request

### LINKED ISSUE #1267: IPv6 literals are broken
**Reproduction**

```
fetch('http://[::1]:8080')
  .then(res => console.debug('ok'))
  .catch(err => console.error('err', err));
```

```
getaddrinfo ENOTFOUND [::1]
```

**Expected behavior**

- If there is a server at `[::1]:8080`, it should not fail

**Cause**

- This was working fine in 2.x, but in 3.x the URL parser switched from Node's `url.Url` to WHATWG's `URL`

```
> u = url.parse('http://[::1]');
Url {
...
  host: '[::1]',
  hostname: '::1',
...
}
> u = new URL('http://[::1]')
URL {
...
  host: '[::1]',
  hostname: '[::1]',
...
}
```

- [offending line, 3.x](https://github.com/node-fetch/node-fetch/blob/914ce6be5ec67a8bab63d68510aabf07cb818b6d/src/request.js#L265) 
- [2.x](https://github.com/node-fetch/node-fetch/blob/b5e2e41b2b50bf2997720d6125accaf0dd68c0ab/src/request.js#L245) passes `url.Url` directly to `request()`

serverwentdown:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

The previous `url.parse` did remove brackets from hostname the `new URL` don't do it.

To fix this i Replaced `https.request(options[, callback])` with `https.request(url[, options][, callback])` and letting NodeJS take care of parsing the the url

- This strips square brackets ([]) off hostnames that have them
- <s>implicit allow basic auth `http://username:password@hostname` that wasn't working before</s>

Closes #1267 

## TODO:

 - [x] Throw if url includes credentials in the constructor<br>TypeError("Request cannot be constructed from a URL that includes credentials")
   - [x] Add test for this

### DIFF
--- a/docs/v3-LIMITS.md
+++ b/docs/v3-LIMITS.md
@@ -7,6 +7,8 @@ Known differences
 
 - On the upside, there are no forbidden headers.
 
+- Passing basic authentication credentials (`https://username:password@hostname`) in the URL is supported.
+
 - `res.url` contains the final url when following redirects.
 
 - For convenience, `res.body` is a Node.js [Readable stream][readable-stream], so decoding can be handled independently.
--- a/docs/v3-LIMITS.md
+++ b/docs/v3-LIMITS.md
@@ -7,8 +7,6 @@ Known differences
 
 - On the upside, there are no forbidden headers.
 
-- Passing basic authentication credentials (`https://username:password@hostname`) in the URL is supported.
-
 - `res.url` contains the final url when following redirects.
 
 - For convenience, `res.body` is a Node.js [Readable stream][readable-stream], so decoding can be handled independently.
--- a/src/request.js
+++ b/src/request.js
@@ -204,12 +204,14 @@ export const getNodeRequestOptions = request => {
 	// HTTP-network fetch step 4.2
 	// chunked encoding is handled by Node.js
 
-	// Overwrite search to force request() to retain trailing ? (issue #776)
 	const search = getSearch(parsedURL);
 
-	// Manually spread the URL object instead of spread syntax
+	// Pass the full URL directly to request(), but overwrite the following 
+	// options:
 	const options = {
+		// Overwrite search to retain trailing ? (issue #776)
 		path: parsedURL.pathname + search,
+		// The following options are not expressed in the URL
 		method: request.method,
 		headers: headers[Symbol.for('nodejs.util.inspect.custom')](),
 		insecureHTTPParser: request.insecureHTTPParser,
--- a/src/request.js
+++ b/src/request.js
@@ -206,7 +206,7 @@ export const getNodeRequestOptions = request => {
 
 	const search = getSearch(parsedURL);
 
-	// Pass the full URL directly to request(), but overwrite the following 
+	// Pass the full URL directly to request(), but overwrite the following
 	// options:
 	const options = {
 		// Overwrite search to retain trailing ? (issue #776)
--- a/src/request.js
+++ b/src/request.js
@@ -49,6 +49,10 @@ export default class Request extends Body {
 			input = {};
 		}
 
+		if (parsedURL.username !== '' || parsedURL.password !== '') {
+			throw new TypeError(`${parsedURL} is an url with embedded credentails.`);
+		}
+
 		let method = init.method || input.method || 'GET';
 		method = method.toUpperCase();
 
--- a/src/request.js
+++ b/src/request.js
@@ -204,24 +204,20 @@ export const getNodeRequestOptions = request => {
 	// HTTP-network fetch step 4.2
 	// chunked encoding is handled by Node.js
 
+	// Overwrite search to force request() to retain trailing ? (issue #776)
 	const search = getSearch(parsedURL);
 
 	// Manually spread the URL object instead of spread syntax
-	const requestOptions = {
+	const options = {
 		path: parsedURL.pathname + search,
-		pathname: parsedURL.pathname,
-		hostname: parsedURL.hostname,
-		protocol: parsedURL.protocol,
-		port: parsedURL.port,
-		hash: parsedURL.hash,
-		search: parsedURL.search,
-		query: parsedURL.query,
-		href: parsedURL.href,
 		method: request.method,
 		headers: headers[Symbol.for('nodejs.util.inspect.custom')](),
 		insecureHTTPParser: request.insecureHTTPParser,
 		agent
 	};
 
-	return requestOptions;
+	return {
+		parsedURL,
+		options
+	};
 };
--- a/src/index.js
+++ b/src/index.js
@@ -35,7 +35,7 @@ export default async function fetch(url, options_) {
 	return new Promise((resolve, reject) => {
 		// Build request object
 		const request = new Request(url, options_);
-		const { parsedURL, options } = getNodeRequestOptions(request);
+		const {parsedURL, options} = getNodeRequestOptions(request);
 		if (!supportedSchemas.has(parsedURL.protocol)) {
 			throw new TypeError(`node-fetch cannot load ${url}. URL scheme "${parsedURL.protocol.replace(/:$/, '')}" is not supported.`);
 		}
--- a/src/index.js
+++ b/src/index.js
@@ -35,12 +35,12 @@ export default async function fetch(url, options_) {
 	return new Promise((resolve, reject) => {
 		// Build request object
 		const request = new Request(url, options_);
-		const options = getNodeRequestOptions(request);
-		if (!supportedSchemas.has(options.protocol)) {
-			throw new TypeError(`node-fetch cannot load ${url}. URL scheme "${options.protocol.replace(/:$/, '')}" is not supported.`);
+		const { parsedURL, options } = getNodeRequestOptions(request);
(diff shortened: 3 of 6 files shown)


jimmywarting:
`[]` feels like a very specific usecase

what if we just took the hostname straight from `url.parse` instead? 

`hostname = url.parse(parsedURL.toString()).hostname`

I'm still wondering about just doing: `https.request(url[, options][, callback])` instead of spreading out every single option
and remove the `https.request(options[, callback])` option 

---

not fully rejecting this PR. but i like to have options before finalizing something and explore options

serverwentdown:
This is exactly the [implementation inside the Node.js standard library](https://github.com/nodejs/node/blob/0c35fe02d3b202290ab19d2fc72e914397fdba3d/lib/internal/url.js#L1391), so I'd say it is acceptable.

jimmywarting:
> This is exactly the [implementation inside the Node.js standard library](https://github.com/nodejs/node/blob/0c35fe02d3b202290ab19d2fc72e914397fdba3d/lib/internal/url.js#L1391), so I'd say it is acceptable.

Oh, interesting find.

serverwentdown:
Implementation of your proposal: https://github.com/node-fetch/node-fetch/compare/main...serverwentdown:fix-ipv6-literal-parsing-2

I also added tests for IPv6

jimmywarting:
> Implementation of your proposal: main...serverwentdown:fix-ipv6-literal-parsing-2, though it needs some refactoring (parsedURL and options aren't named well)

Do you want to apply that changes to this PR? 
( Also what do @Richienb, @LinusU, @gr2m (and anybody else) think? ) 

> I also added tests for IPv6 on both proposals.

Great! Needed that!

src/request.js
```diff
@@ -200,6 +200,13 @@ export const getNodeRequestOptions = request => {
 	if (!headers.has('Connection') && !agent) {
 		headers.set('Connection', 'close');
 	}
+	
+	// Because of the differences in paring between Node.js's standard library and WHATWG's spec, IPv6 literal hostnames ([::1]) are treated as domain names by Node.js's standard library.
+	// Strip off the square brackets in WHATWG's URL.hostname before handing it to Node.js.
+	let hostname = parsedURL.hostname;
+	if (hostname.startsWith('[') && hostname.endsWith(']')) {
+		hostname = hostname.substring(1, hostname.length - 1)
+	}
```
gr2m, line 209: Maybe add a reference to this pull request in the comment
```suggestion
	// Because of the differences in paring between Node.js's standard library and WHATWG's spec, IPv6 literal hostnames ([::1]) are treated as domain names by Node.js's standard library.
	// Strip off the square brackets in WHATWG's URL.hostname before handing it to Node.js.
	// See https://github.com/node-fetch/node-fetch/pull/1268
	let hostname = parsedURL.hostname;
	if (hostname.startsWith('[') && hostname.endsWith(']')) {
		hostname = hostname.substring(1, hostname.length - 1)
	}
```

gr2m (commented):
Change looks good to me! I'd approve the PR but not sure if there is anything else you want to add

src/request.js
jimmywarting: @gr2m what do you think about using `https.request(url[, options][, callback])` instead of `https.request(options[, callback])` as discussed earlier https://github.com/node-fetch/node-fetch/pull/1268#issuecomment-911880484?

if we do it the other way then this part of the code isn't needed at all and should be removed (replaced with https://github.com/node-fetch/node-fetch/compare/main...serverwentdown:fix-ipv6-literal-parsing-2)
Instead the test-case could references the PR

gr2m: Oh nice I like that much better!

jimmywarting: I'm also in fav. of `https.request(url[, options][, callback])` then we simply just let NodeJS core handle it internally

src/request.js
```diff
@@ -206,22 +206,20 @@ export const getNodeRequestOptions = request => {
...
+		// Overwrite search to retain trailing ? (issue #776)
 		path: parsedURL.pathname + search,
-		pathname: parsedURL.pathname,
-		hostname: parsedURL.hostname,
-		protocol: parsedURL.protocol,
-		port: parsedURL.port,
-		hash: parsedURL.hash,
-		search: parsedURL.search,
-		query: parsedURL.query,
-		href: parsedURL.href,
```
serverwentdown, line 219: For those who are curious why these can be removed: 
- Since we're passing the URL to `request()`, we are now using the Node.js standard library URL parser
- The standard library URL parser effectively [expands the URL into the same set of options](https://github.com/nodejs/node/blob/1c4df352b77514e93a46fb27e996755d45a0b530/lib/internal/url.js#L1388).

serverwentdown: Oh! This led me to a side effect of this PRs' changes: [Basic auth](https://github.com/nodejs/node/blob/1c4df352b77514e93a46fb27e996755d45a0b530/lib/internal/url.js#L1405) (`username:password@hostname`) is now supported implictly by `node-fetch`. 

Should this be a concern?

Seems to be allowed in the [`fetch()` spec, §4.2 step 8.19.2.3.](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch)

LinusU (approved):
I think that this seems like a great approach 👍

gr2m:
can you address the lint errors?

image


Running `npx xo --fix` should address them all

src/request.js
jimmywarting: 👍

serverwentdown:
Turns out the `fetch()` spec lists basic auth, but is not implemented by any browsers for security reasons. But, allowing basic auth should be useful for majority of applications.

jimmywarting (approved):

jimmywarting:
Started to get 2nd toughs on supporting basic auth...

Starting looking more into the `isAuthenticationFetch` in the spec and also trying it out in chromes devtool
image

Guess it have something to do with caching... it's bad to use [cache storage](https://developer.mozilla.org/en-US/docs/Web/API/Cache) and save the username & text in full plain text (url)

- the usr/psw should be converted to a basic auth request header to avoid saving the headers to the cache storage.
- It's also b/c the usr:psw should not be exposed in the `response.url` (for security reasons)

We could argue that we don't have the same level of restrictions and we don't have a cache storage.

I guess the `isAuthenticationFetch` flag is mostly set to true when working with mitm service workers and you get prompted with that builtin old basic auth login dialog in the browsers.

We also have a hole option with `credentials` that we haven't looked into much at all in node-fetch
`new Request(url, { credentials: 'include' })`

> credentials: The request credentials you want to use for the request: omit, same-origin, or include. The default is same-origin.

we don't have a concept of origin either.

Either way i still think this PR looks good question is wether or not we shoud:
- ...also throw an error when constructing the Request with a url that includes credentials
- or if we should auto "converted to an `Authorization` value" as mention in the spec also and strip away the usr/psw from the url
- or just simply allow it

we could also start having a look at what the credential option is

jimmywarting:
fyi. i tried deno's built in fetch api also... `fetch('https://foo:bar@vecka.nu', {credentials: 'include'})` worked out fine

jimmywarting:
@node-fetch/reviewers Should we throw an error for the time being on basic-auth-url? 
it's easier to lift the restriction rather than removing/deprecating/discourage it.

Our ambition is that the code you write should be working everywhere the same way in Deno, Node & Browsers.
And if non of the browsers allow it then everyone should be using the solution that works cross all platforms. Meaning: using btoa and appending basic auth headers & throwing an error like browser dose.

gr2m:
> Our ambition is that the code you write should be working everywhere the same way in Deno, Node & Browsers.
> And if non of the browsers allow it then everyone should be using the solution that works cross all platforms

I concur

docs/v3-LIMITS.md
```diff
@@ -7,6 +7,8 @@ Known differences
 
 - On the upside, there are no forbidden headers.
 
+- Passing basic authentication credentials (`https://username:password@hostname`) in the URL is supported.
```
jimmywarting, line 10: For now lets just throw a TypeError if url includes credentials in the Request constructor.
We can decide later what to do with it. Want to find out what the real reason is first.

jimmywarting (changes requested):

serverwentdown:
I also realised that the specification somewhat contradicts: when creating a `Request` with a URL that includes credentials, it [should throw a `TypeError`](https://fetch.spec.whatwg.org/#ref-for-include-credentials④).

jimmywarting (approved):

jimmywarting:
Thanks for your contribution to node fetch
It will be released in the next version.

I don't have the possibility to make a release yet until Sunday
I'm on a vacation now with no computer. Unless someone else dose it...

Would be nice to have that auto ci deployment now...

jimmywarting:
@serverwentdown If you would like to help out more with one other issue then i got something easy for you. (You don't have to if you don't want)

If a Request is constructed with method HEAD or GET then delete content-length request header #1249

lachesis:
This change to reject implicit basic-auth breaks our use case. We will not upgrade node-fetch past v2 until or unless this is fixed.

PR #1274, opened 2021-09-04
Fix(package.json): Corrected main file path in package.json

### LINKED ISSUE #1269: Version 2.6.1 has incorrect configuration for CJS in package.json
As file with `jest-fetch-mock`:
https://github.com/jefflau/jest-fetch-mock/issues/161#issuecomment-911921788

There is a mistake in the "main" entry in the package.json release in 2.6.1 that is causing import errors.

It currently reads:
```
"main": "lib/index",
```

It should be:
```
"main": "lib/index.js",
```

jimmywarting:
## Purpose
Fixes a problem when some tries to read the main file. 
It was a problem due to lack of file extension

## Changes

- Added file extension in package.json `"main":`
- I updated patch version in `package.json`
- I pinned codecov + teeny-request (it had breaking change)
- also changed the "external" function ( 2b0610d ) - rollup had some breaking changes also

## Additional

This is for v2

-------------------------------------------------------------------------------------------------

- [x] I updated changelog

-------------------------------------------------------------------------------------------------

- fixes #1269


jimmywarting:
Seems like teeny-request has a breaking change for an older node version where it uses spread `...` would it help to install it as a dep, pin it so codecov uses it instead?

jimmywarting:
Ironically codecov uses teeny-request
and teeny-request just wraps node-fetch v2.6.1

So we use our own fetch api to test our own api...

but at codecov i found out that:
> On February 1, 2022 this uploader will be completely deprecated and will no longer be able to upload coverage to Codecov.
> https://github.com/codecov/codecov-node/blob/master/README.md

jimmywarting:
... do rollup have some breaking change also?

```
src/index.js → lib/index.js, lib/index.es.js, lib/index.mjs...
[!] TypeError: Cannot read property 'src' of undefined
TypeError: Cannot read property 'src' of undefined
    at external (./node-fetch/rollup.config.js:47:52)
```

jimmywarting:
the external function did something that didn't work. 2b0610d
don't remember what we use this for...
either way... we don't have any dependencies in package.json so i just removed that part and it started working again.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ Changelog
 
 # 2.x release
 
+## v2.6.2
+
+- Fix: used full filename for main in package.json
+- Other: pinned all dev dependencies (codecov -> teeny-request had on breaking change)
```
gr2m, line 11: instead of pinning all dev dependencies, maybe we can just remove codecov from `2.x`?

jimmywarting: As long as codecov works then we could still use it? it's going to stop working Feb 1 2022... maybe remove it then?

I did a smaller pining earlier where i only pinned codecov (+ that i added & pinned teeny-request into devDep also) it seemed to work fine. Should we do this instead?

gr2m: I removed codecov a long time ago because it produced to many random failed builds due to 5xx errors. 

We can leave it in the `main` branch, I don't see much value for it in `2.x`. 

My recommendation would be to get to 100% coverage and then fail the tests if it drops below, no need for an external service. That's what I do with my projects anyway, no more random CI errors when the coverage service is down

> I did a smaller pining earlier where i only pinned codecov (+ that i added & pinned teeny-request into devDep also) it seemed to work fine. that's when i got past the error and starting getting errors with rollup instead
> Should we do this instead?

I would prefer that

jimmywarting: Right now in v3 we use c8 instead of codecov that uses nodes internal code coverage instead of wrapping all code to check if some stuff are executed + c8 runs locally with the CI workflow
So codecov is removed in v3.

> I would prefer that

Then I will do that.

jimmywarting: fixed

gr2m (approved):

LinusU (approved):
Nice 👍

PR #1287, opened 2021-09-13
chore: remove triple-slash directives from typings (#1285)

### LINKED ISSUE #1285: Typings polutes global space with DOM types
Same issue superagent had with their typings: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/41425#issuecomment-573385657

The `dom` reference triple-slash directive in `/@types/index.d.ts` causes all DOM types to be included in the global ambient declarations which breaks type checking.

https://github.com/node-fetch/node-fetch/blob/9cd2e43e6272a48d8971843bf93e0deb498664fa/%40types/index.d.ts#L2

See: https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-lib-

This means that Node.js projects that import `node-fetch` can reference DOM global types and values (`window`, `document`, `Element`, `ChildNode`, etc.) without typing errors.

This line don't seem to be needed as nothing from the DOM library is referenced. Can we just delete this line? I tried doing so, and TypeScript had no complaints.

Reproduction:

Simple TypeScript project that does NOT include DOM library.
```bash
echo '{
  "compilerOptions": {
    "lib": ["ES2015"]
  }
}' > tsconfig.json

npm i node-fetch @types/node
```

Attempting to access the window global throws a type error as expected.
```bash
echo "window.alert('hi');" > index.ts

tsc
# index.ts:1:1 - error TS2304: Cannot find name 'window'.
```

Same as above except we import node-fetch and the type error disappears unexpectedly.
``` bash
echo "import {} from 'node-fetch';
window.alert('hi');" > index.ts

tsc
# no errors
```

TypeScript reports the DOM library is being loaded due to a reference in node-fetch's type declaration file.
```bash
tsc --explainFiles
# ../../.nvm/versions/node/v14.17.6/lib/node_modules/typescript/lib/lib.dom.d.ts
#   Library referenced via 'dom' from file 'node_modules/node-fetch/@types/index.d.ts'
```

As you can see, simply importing `node-fetch` breaks type checking because `lib.dom.d.ts` is loaded due to the reference from `node-fetch/@types/index.d.ts`. This can also be observed in VSCode where DOM types are shown in autocomplete suggestions.

yo1dog:

- [ ] Documentation update
- [X] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Remove the `dom` reference triple-slash directive from the typings file which pollutes the global ambient declarations with DOM types.

- fix #1285

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -1,5 +1,4 @@
 /// <reference types="node" />
-/// <reference lib="dom" />
 
 import {Agent} from 'http';
 


carnesen:
@yo1dog Does removing `/// <reference lib="dom" />` break this package's types? I would guess it's there because this package references a type from the `dom` lib.

yo1dog:
#1285

> This line don't seem to be needed as nothing from the DOM library is referenced. Can we just delete this line? I tried doing so, and TypeScript had no complaints.

I removed the line there were no TypeScript errors in index.d.ts.

jimmywarting (approved):

joao-paulo-parity:
By the Blame feature (https://github.com/node-fetch/node-fetch/blame/6425e2021a7def096e13dbabcac2f10e6da83d11/%40types/index.d.ts) I found https://github.com/node-fetch/node-fetch/pull/1141/files#r671191147 which already more or less predicted that this `/// <reference lib="dom" />` directive would become a problem. I think that is enough motivation to go forward with this PR.

What is missing and how can we unblock it? Would it help to perhaps put the typechecking on CI so that we'll not need someone to manually verify that the types are still correct?

PeterF521 (approved):
I agree with this change, because adding a reference of DOM library pollutes `global` scope with useless DOM declarations, considering that `node-fetch` is only used in Node.js projects.

stephank:
I guess my project relied on these definitions, because it now fails with:

```
.yarn/cache/node-fetch-npm-3.2.9-37028f4521-ba421350b2.zip/node_modules/node-fetch/@types/index.d.ts:124:4 - error TS2749: 'FormData' refers to a value, but is being used as a type here. Did you mean 'typeof FormData'?

124  | FormData
       ~~~~~~~~

.yarn/cache/node-fetch-npm-3.2.9-37028f4521-ba421350b2.zip/node_modules/node-fetch/@types/index.d.ts:137:22 - error TS2749: 'FormData' refers to a value, but is being used as a type here. Did you mean 'typeof FormData'?

137  formData(): Promise<FormData>;
                         ~~~~~~~~
```

I think it's now trying to use this as a type: https://github.com/jimmywarting/FormData/blob/0fc449d79874bb4bcea45266080254f451b15a37/esm.min.d.ts#L1-L4

But error also pop up in dependencies. `formdata-polyfill` itself has similar errors when DOM definitions are disabled, and `fetch-blob` breaks because `Blob` and `File` are no longer defined.

The simple workaround is adding `"dom"` to `lib` in `tsconfig.json`, but guess I don't really understand how this change was intended to work. 🤷‍♂️ (But I do agree it should be done somehow!)

yo1dog:
> I guess my project relied on these definitions, because it now fails with:
> 
> ```
> .yarn/cache/node-fetch-npm-3.2.9-37028f4521-ba421350b2.zip/node_modules/node-fetch/@types/index.d.ts:124:4 - error TS2749: 'FormData' refers to a value, but is being used as a type here. Did you mean 'typeof FormData'?
> 
> 124  | FormData
>        ~~~~~~~~
> 
> .yarn/cache/node-fetch-npm-3.2.9-37028f4521-ba421350b2.zip/node_modules/node-fetch/@types/index.d.ts:137:22 - error TS2749: 'FormData' refers to a value, but is being used as a type here. Did you mean 'typeof FormData'?
> 
> 137  formData(): Promise<FormData>;
>                          ~~~~~~~~
> ```

I am unable to replicate. I just did a quick test:
```
mkdir test && cd test
npm i node-fetch @types/node typescript
echo 'import fetch from 'node-fetch'; console.log(fetch);' > test.ts
npx tsc test.ts
```

Node 16.14.2
TypeScript 4.7.4
node-fetch 3.2.9
formdata-polyfill 4.0.10

No errors. Can you share what version of Node and Typescript you are using? Also the installed versions of node-fetch and formdata-polyfill?

stephank:
I can reproduce it with your example if I instead do `npx tsc --lib ES2021 test.ts`. I guess DOM is a default include, and setting `lib` explicitely (via cli or tsconfig) removes it.

node.js v16.15.0
```
$ npm ls -a
nf@ /src/tests/nf
├── @types/node@18.0.6
├─┬ node-fetch@3.2.9
│ ├── data-uri-to-buffer@4.0.0
│ ├─┬ fetch-blob@3.2.0
│ │ ├── node-domexception@1.0.0
│ │ └── web-streams-polyfill@3.2.1
│ └─┬ formdata-polyfill@4.0.10
│   └── fetch-blob@3.2.0 deduped
└── typescript@4.7.4
```

Issue #1263, opened 2021-09-01
ESM only breaks everything


yinzara:
Following the conversation in:
https://github.com/sindresorhus/meta/discussions/15#discussioncomment-1256621

PLEASE PLEASE PLEASE, don't do this yet.

ESM is not ready to have a library as major as node-fetch go ESM only.  The workarounds you list in the README will not work in all cases.

node-fetch is one of the most mocked libraries in existence.  This move will cause MANY projects that's sole purpose is to mock node-fetch to cease to work with no real way to work around the problem.

PLEASE reconsider.

heath-freenome:
Or at least document how to get things working with jest and webpack like formatJS does: https://formatjs.io/docs/react-intl#jest

herberthobregon:
the solution is to add `"exports"` in the `package.json` for commonjs and esm (?)

Taelyn:
Honestly i find that a lazy/stupid/easy solution to the issue. "just stay on v2.6.1"

There was no need to break everyones app then the author thought it was smart to do so

For me dropping this moveing to another library. The owners might need to go learn some skills ;)
This is something you dont do with products.

LinusU:
> [...] , don't do this yet.

@yinzara we will continue to fix critical bugs and security issues in 2.x, so there should be nothing preventing anyone to stay on node-fetch 2.x until they are ready to switch to ESM.

> There was no need to break everyones app then the author thought it was smart to do so

@Taelyn I don't see how any app was broken since this was behind a major version bump. Did you have your version of `node-fetch` set to `"*"`? 🤔 

> It's a community splitting decision, like Python 3 vs 2.7, which has been an on-going battle for over a decade, I wouldn't be surprised to see the same happen here unless changes are made.

@nannal I think that this is why it's important for the ecosystem to migrate to EMS as soon as possible. The problem with Python 2/3 was that many major libraries kept supporting only Python 2, or kept supporting both and thus didn't encourage anyone to update.

yinzara:
> @yinzara we will continue to fix critical bugs and security issues in 2.x, so there should be nothing preventing anyone to stay on node-fetch 2.x until they are ready to switch to ESM.

Why do this at all yet? That means you now have to maintain two branches of code for all security issues and critical bugs. What if there are any official changes to the fetch API? What about a non-critical bug that still affects your users? 2.x users will then be SoL.

[...]

PR #1288, opened 2021-09-13
feat: convert to ESM/CJS hybrid package

yinzara:
As outlined in #1263, the NodeJS ecosystem is not yet prepared for a package as major as `node-fetch` being ESM only.

This PR adds cross transpilation of the 3.0.0 ESM version of `node-fetch` using `esbuild` as a `prepack` step before publishing and configuration in the package.json to allow for both styles of imports.

For ESM only packages that use `node-fetch`, they should be able to 'tree shake' the CJS files so they don't impact application size.

For CJS packages, they can continue to use `node-fetch` as they originally did with the understanding that in the future this support could be removed.

I have updated documentation to reflect the change.

This should allow abandonment of the 2.x branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# CJS generated output
+cjs
+
 # Sketch temporary file
 ~*.sketch
 
--- a/README.md
+++ b/README.md
@@ -126,14 +126,7 @@ if (!globalThis.fetch) {
 }
 ```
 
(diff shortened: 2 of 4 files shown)


jimmywarting:
Locking this as a duplicate of #1227 

There is no point to submit more variant of this solution unless you first can convince some of the maintainer that this is something we want to support.

It's not enough to argument why it should be supported and what all the benefits are. You also have to understand what the drawbacks of what this means to bundle all the things to cjs... Something which is not mentioned at all in this PR.

Issue #1290, opened 2021-09-14
v2: Better UTF-8 URL handling


jlowcs:
**Is your feature request related to a problem? Please describe.**

Since v3 is ESM only and your recommendation is to use v2 is ESM if this is blocking for us, could it be possible to port [this improvement](https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#better-utf-8-url-handling) to v2?

Indeed, we already had a few instances of a SSR issue in production where the fetched url contained UTF-8 characters and made the fetch fail server-side but not client-side. Those issues are pretty tricky to catch ahead of time if the developer forgets to take UTF-8 into account when building the url.

**Describe the solution you'd like**

Port [the UTF-8 improvement](https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#better-utf-8-url-handling) to v2.

LinusU:
1. Do you have an example of an url that is failing that we could use as a test case?
2. Would you be able to submit a PR for this?

jlowcs:
1. Here's a very simple test case:
```
require('node-fetch')('http://www.foo.com/ひらがな')
```
![image](https://user-images.githubusercontent.com/3777724/133255223-1734fe42-c4e8-4355-a032-5f331c602389.png)

2. Well, I would assume that you would have a pretty good idea on how to implement that fix since it was implemented in v3, while it will probably take me more time just because I'm not familiar with the codebase. That being said, if you give me the necessary pointers so I don't spend too much time on this, I suppose I could take care of that PR.

LinusU:
I've submitted a PR for this here: https://github.com/node-fetch/node-fetch/pull/1291

PR #701, opened 2019-12-02
refactor: Replace `url.parse` with `new URL()`

xxczaki:
I think this is the last thing we need to do in order to release the 3.x alpha release.

Replacing the `url.parse` API here doesn't break our tests:
https://github.com/bitinn/node-fetch/blob/3b452b8e1f280142d78d5a33abf945d77c25ea59/src/request.js#L51

However, trying to use the new WHATWG API here:
https://github.com/bitinn/node-fetch/blob/3b452b8e1f280142d78d5a33abf945d77c25ea59/src/request.js#L54

throws some errors:

```
> node-fetch@2.6.0 test /home/akepinski/dev/node-fetch
> cross-env BABEL_ENV=test mocha --require @babel/register --throw-deprecation test/test.js



  node-fetch
    ✓ should return a promise
    1) should return a promise
(node:11807) UnhandledPromiseRejectionWarning: FetchError: request to http://localhost:30001/hello failed, reason: connect ECONNREFUSED 127.0.0.1:80
    at ClientRequest.<anonymous> (/home/akepinski/dev/node-fetch/src/index.js:113:11)
    at ClientRequest.emit (events.js:210:5)
    at Socket.socketErrorListener (_http_client.js:406:9)
    at Socket.emit (events.js:210:5)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
(node:11807) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)


  1 passing (35ms)
  1 failing

  1) node-fetch
       should return a promise:
     Uncaught DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
      at emitDeprecationWarning (internal/process/promises.js:153:11)
      at processPromiseRejections (internal/process/promises.js:205:13)
      at processTicksAndRejections (internal/process/task_queues.js:94:32)



npm ERR! Test failed.  See above for more details.
```

Trying to use the new API [here](https://github.com/bitinn/node-fetch/blob/3b452b8e1f280142d78d5a33abf945d77c25ea59/src/request.js#L59) also breaks our tests, but with more precise errors:

```
175 passing (2s)
  29 failing

  1) node-fetch
[...]

PR #1291, opened 2021-09-14
fix: properly encode url with unicode characters

LinusU:

- Documentation update
- [x] Bug fix
- New feature
- Other, please explain:

I've added a call to `encodeURI` before parsing a passed in string as an url

#1290 

This is basically a backport of #701, but keeps the old code path for arbitrary URLs on the `Request` object which was supported in 2.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "node-fetch",
-    "version": "2.6.2",
+    "version": "2.6.3",
     "description": "A light-weight module that brings window.fetch to node.js",
     "main": "lib/index.js",
     "browser": "./browser.js",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,9 @@
         "url": "https://github.com/bitinn/node-fetch/issues"
     },
     "homepage": "https://github.com/bitinn/node-fetch",
+    "dependencies": {
+        "whatwg-url": "^5.0.0"
+    },
     "devDependencies": {
         "@ungap/url-search-params": "^0.1.2",
         "abort-controller": "^1.1.0",
@@ -60,7 +63,6 @@
         "rollup": "^0.63.4",
         "rollup-plugin-babel": "^3.0.7",
         "string-to-arraybuffer": "^1.0.2",
-        "teeny-request": "3.7.0",
-        "whatwg-url": "^5.0.0"
+        "teeny-request": "3.7.0"
     }
 }
--- a/src/request.js
+++ b/src/request.js
@@ -9,6 +9,7 @@
 
 import Url from 'url';
 import Stream from 'stream';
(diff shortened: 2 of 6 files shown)


LinusU:
>  The build canceled.

Hmmm 🤔 no idea why that happened...

src/request.js
```diff
@@ -62,7 +62,7 @@ export default class Request {
 				parsedURL = parse_url(input.href);
 			} else {
 				// coerce input to a string before attempting to parse
-				parsedURL = parse_url(`${input}`);
+				parsedURL = parse_url(encodeURI(`${input}`));
```
jlowcs, line 65: There's a possibility that the url is already encoded, in which case it would be encoded twice. Would that end up being an issue?

Adding a test case with an already-encoded url might be a good idea.

LinusU: Yeah, you're right. The `URL` constructor handles this so we must too 👍

LinusU: Fixed ✅

jimmywarting:
can just as well take the opportunity to bump pkg.version and update the changelog as well?

jimmywarting (changes requested):
can just as well take the opportunity to bump pkg.version and update the changelog as well?

LinusU:
@jimmywarting pushed a commit for that 👍

jimmywarting (approved):

LinusU:
@jimmywarting I don't have permission to publish to npm, would you be able to?

jimmywarting:
Yes i have permission

jimmywarting:
Released 2.6.3 to npm and github release.

dmichon-msft:
This PR caused a breaking change to the encoding behavior of URLs that breaks, for example, the proper formatting of SAS tokens in URLs in `@azure/core-http`.

LinusU:
@dmichon-msft sorry, I missed your comment here. As you are probably already aware this was fixed in `2.6.4`, and we've added a test for it so that it won't happen again ☺️

package.json
```diff
@@ -36,6 +36,9 @@
         "url": "https://github.com/bitinn/node-fetch/issues"
     },
     "homepage": "https://github.com/bitinn/node-fetch",
+    "dependencies": {
+        "whatwg-url": "^5.0.0"
```
SpiridonovAlexMoeDelo, line 40: why this version?
actual is 9.1.0

this breaks our code because we have dependency on whatwg-url v.6.5.0 already

LinusU: > why this version?
> actual is 9.1.0

This is the latest version that supports Node.js 4.x

> this breaks our code because we have dependency on whatwg-url v.6.5.0 already

In which way does your app break? 🤔

PR #1310, opened 2021-09-23
2.x: Specify encoding as an optional peer dependency in package.json

ciffelia:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

I installed node-fetch 2.x on my project with Yarn 2 (aka Yarn berry). When I call `require('node-fetch')`, I get this warning message:

```
(node:22404) [MODULE_NOT_FOUND] Error: node-fetch tried to access encoding, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

It seems that `encoding` package is optional and only used in `body.textConverted()` method (discussed in #412).

So I added the following to `package.json` of node-fetch:

```json
"peerDependencies": { 
    "encoding": "*"
},
"peerDependenciesMeta": {
    "encoding": {
        "optional": true
    }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,14 @@
     "dependencies": {
         "whatwg-url": "^5.0.0"
     },
+    "peerDependencies": { 
+        "encoding": "*"
+    },
+    "peerDependenciesMeta": {
+        "encoding": {
+            "optional": true
+        }
+    },
     "devDependencies": {
         "@ungap/url-search-params": "^0.1.2",
         "abort-controller": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
         "whatwg-url": "^5.0.0"
     },
     "peerDependencies": { 
-        "encoding": "*"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -39,6 +39,14 @@
     "dependencies": {
         "whatwg-url": "^5.0.0"
     },
+    "peerDependencies": { 
+        "encoding": "*"
```
LinusU, line 43: I think it's better to be explicit here? Since the package is still in `0.x` phase it seems very possible that there will be breaking changes ahead...

```suggestion
        "encoding": "^0.1.0"
```

ciffelia: Yes, I'll change that.

LinusU (approved):
Neat 👍

jimmywarting (approved):

ljharb:
Heads up that this was actually a breaking change within v2 - it being optional doesn't change that.

PR #1311, opened 2021-09-24
fix(Types): make `Body.json()` method generic

iShibi:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain:

Makes the `json` method generic. This will allow users to tell tsc the return type of the method. It's opt-in, which means if no type is provided, the return type will remain `unknown`

**What changes did you make?**

Turned `BodyMixin#json` into a generic method.
```ts
const a = await res.json(); // 'a' is unknown

interface User {
 name: string;
}
const b = await res.json<User>(); // 'b' is of type User
```

Lack of IntelliSense due to return type of `json` being `unknown`. It still defaults to `unknown` but the user can opt-in by providing a return type for the method.

### DIFF
--- a/@types/index.test-d.ts
+++ b/@types/index.test-d.ts
@@ -29,6 +29,14 @@ async function run() {
 
 	// Test JSON, returns unknown
 	expectType<unknown>(await getResponse.json());
+	// Test JSON with generics, returns T
+	interface T {
+		userId: number,
+		id: number,
+		title: string,
+		body: string
+	}
+	expectType<T>(await (await fetch('https://jsonplaceholder.typicode.com/posts/1')).json())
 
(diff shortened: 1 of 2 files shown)


@types/index.test-d.ts
```diff
@@ -29,6 +29,14 @@ async function run() {
 
 	// Test JSON, returns unknown
```
Kristrue, line 30: Kris

Kristrue (approved):

iShibi:
Closing this because using generics this way isn't preferred. See [Common Mistakes](https://github.com/definitelytyped/definitelytyped#common-mistakes)

Issue #1252, opened 2021-08-26
Use stream consumers


jimmywarting:
NodeJS released some [stream consumers](https://nodejs.org/api/webstreams.html#webstreams_utility_consumers) utility, so brought it to the [userland](https://www.npmjs.com/package/stream-consumers)

thought we could use it in both the body mixin 
Grate thing about it is that it operates on async iterator, so it can work with both async iterators, whatwg & node streams

PR #1314, opened 2021-09-26
feat(Body): Added support for `BodyMixin.formData()` and constructing bodies with FormData

### LINKED ISSUE #199: Implement body.formData

### LINKED ISSUE #1293: error TS2322: Type 'FormData' is not assignable to type 'BodyInit'
Using `formdata-polyfill` with `node-fetch@3.0.0` results in the following error:
```
error TS2322: Type 'FormData' is not assignable to type 'BodyInit'.
  Property 'sort' is missing in type 'FormData' but required in type 'URLSearchParams'.
```

jimmywarting:
## Purpose
- Add support for both `response.formData()` and `request.formData()`
- Support constructing not only `new Request(url, {body: formData})` but also `new Response(formData)` as it's supposed to be (only supported Request before...)
- It's also able to convert a `URLSearchParam` to a formData as it dose in browser/spec

This would be a great addition so you can avoid having to depend on some external multiparty, formidable, busboy or any other body parser packages. that all have some different (non-speced) apis that you would need to learn.
Instead you could just do:

```js
import { Response } from 'node-fetch'

http.createServer(async function (req, res) {
  const formData = await new Response(req, { headers: req.headers }).formData()
  const file = formData.get('avatar')
  
  var arrayBuffer = await file.arrayBuffer()
  var text = await file.text()
  var whatwgReadableStream = file.stream()
}) 
```

## Changes
- I changed the FormData test file to test more end to end instead.
- Added a fast algoritm to parse formData

## Additional information

Maybe thinking it would be nice to expose FormData, Blob and File from index.js so they are accessible, thoughts? 
Think we have had this discussion before with exposing Blob before... #392

___

- [x] I updated ./docs/CHANGELOG.md
- [x] I updated readme
- [x] I added unit test(s)

___

closes #199
closes #1118
closes #1293
closes #876


jimmywarting:
Ready to be reviewed...

package.json
```diff
@@ -91,6 +91,8 @@
       "unicorn/numeric-separators-style": 0,
       "unicorn/explicit-length-check": 0,
       "capitalized-comments": 0,
+      "node/no-unsupported-features/es-syntax": 0,
+      "no-fallthrough": 0,
```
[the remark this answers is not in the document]
jimmywarting: fixed

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,436 @@
...
+
+			parser.end();
+
+			return fd;
+		}
+
+		throw new TypeError('no or bad content-type header, no multipart boundary');
+	} else {
+		throw new TypeError('Failed to fetch');
+	}
```
LinusU, line 439: Instead of putting this at the end and having the function body inside the `if`, could we invert the if expression and throw early?

e.g.

```
if (!/multipart/i.test(ct)) {
  throw new TypeError('Failed to fetch');
}

const m = ct.match(/boundary=(?:"([^"]+)"|([^;]+))/i);

if (m == null) {
  throw new TypeError('no or bad content-type header, no multipart boundary');
}

// ...
```

I think that it's much easier to see which conditions the throws corresponds to then...

src/body.js
```diff
@@ -9,10 +9,10 @@ import Stream, {PassThrough} from 'stream';
 import {types, deprecate} from 'util';
 
 import Blob from 'fetch-blob';
+import {FormData, formDataToBlob} from 'formdata-polyfill/esm.min.js';
```
LinusU, line 12: Hmmm, will this break `instanceof` if someone else have imported the `FormData` from as so: `import { FormData } from 'formdata-polyfill'`? Maybe that doesn't matter since our public API doesn't specify how `instanceof` will work with `FormData` 🤔 

Yeah, it probably doesn't matter? 🤔

src/body.js
```diff
@@ -352,11 +370,6 @@ export const getTotalBytes = request => {
 		return body.hasKnownLength && body.hasKnownLength() ? body.getLengthSync() : null;
 	}
 
-	// Body is a spec-compliant FormData
-	if (isFormData(body)) {
-		return getFormDataLength(request[INTERNALS].boundary);
```
LinusU, line 357: Seems like this was very broken before? It doesn't pass the `form` argument...

src/body.js
jimmywarting: My FormData package uses duck type checking: [Symbol.hasInstance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance)

so as long as it has all the same properties then it can match any kind of instances

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,440 @@
...
+				this[markSymbol] = 0;
+			}
+		};
+
+		for (i = 0; i < length_; i++) {
+			c = ui8a[i];
+
+			switch (state) {
+				case S.PARSER_UNINITIALIZED:
+					return i;
```
LinusU, line 108: The `state` variable is never set to `PARSER_UNINITIALIZED`, in fact `PARSER_UNINITIALIZED` isn't used at all so I think that we should remove it

```suggestion
```

src/body.js
jimmywarting: yea, it was somewhat broken before, it only. It did what it was suppose to do but only for `Request`
What i do now instead is that i'm "normalizing" the FormData into a stream so we have less objects to work with...

did it in: https://github.com/node-fetch/node-fetch/blob/3a9c1bd54a9d2f85f028f7716238e3faf4c2909d/src/body.js#L53-L57

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,440 @@
...
+	START_BOUNDARY: s++,
+	HEADER_FIELD_START: s++,
+	HEADER_FIELD: s++,
+	HEADER_VALUE_START: s++,
+	HEADER_VALUE: s++,
+	HEADER_VALUE_ALMOST_DONE: s++,
+	HEADERS_ALMOST_DONE: s++,
+	PART_DATA_START: s++,
+	PART_DATA: s++,
+	PART_END: s++,
```
LinusU, line 17: This isn't used anywhere

```suggestion
```

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,440 @@
...
+						// reconsider the current character even so it interrupted the sequence
+						// it could be the beginning of a new sequence
+						i--;
+					}
+
+					break;
+				case S.END:
+					break;
+				default:
+					return i;
```
LinusU, line 298: From examining all the points where `state` is modified, it should never be able to hit this `default`? I think it would be better to throw here since this should never be able to happen:

```suggestion
				default:
					throw new Error(`Impossible state encountered: ${state}`)
```

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,440 @@
...
+			}
+		}
+
+		dataCallback('headerField');
+		dataCallback('headerValue');
+		dataCallback('partData');
+
+		this.index = index;
+		this.state = state;
+		this.flags = flags;
```
LinusU, line 308: What is the point of extracting these variables and then assigning them back? 🤔 

Is it to avoid exposing some kind of in-between state outwards? Maybe we should add a comment about it?

src/utils/multipart-parser.js
jimmywarting: Think I'm adding a comment instead... This isn't my own code I have written. It's fairly modified doe to strip out unnecessary parts

src/utils/multipart-parser.js
jimmywarting: I think this was a speed optimization provided by the original author.

Before entering the the hole for-loop and doing lots of chain lookup `this.state = x` then i guess it was faster to just do `state = x` and then update the property later.

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,440 @@
...
+			const callbackSymbol = 'on' + name.slice(0, 1).toUpperCase() + name.slice(1);
+			if (callbackSymbol in self) {
+				self[callbackSymbol]();
+			}
+		}
+
+		if ((this.state === S.HEADER_FIELD_START && this.index === 0) ||
+			(this.state === S.PART_DATA && this.index === this.boundary.length)) {
+			callback(this, 'partEnd');
+			callback(this, 'end');
```
LinusU, line 324: `callbackSymbol in self` will try to call `undefined` if the user sets e.g. `onPartEnd` to `undefined`, what do you think about checking `self[callbackSymbol] != null` instead (which would then allow `= undefined`, `= null`, or `delete` to clear it)?

-----

Generally, I think that doing dynamic lookups to make the call site cleaner increases the cognitive load of reading the code. Personally, the most straightforward thing that I would have liked would have been `this.onPartEnd?.()`, but since we are still supporting Node.js 12 we can't do that 😅. Still, I think that something like this would still make the code easier to read:

```suggestion
		function callback(fn) {
			if (fn != null) fn();
		}

		if ((this.state === S.HEADER_FIELD_START && this.index === 0) ||
			(this.state === S.PART_DATA && this.index === this.boundary.length)) {
			callback(this.onPartEnd);
			callback(this.onEnd);
```

and then maybe hoist up the function and call it something generic:

```js
function callIfNonNullish(fn, ...args) {
  if (fn != null) fn(...args)
}
```

I think that this avoids one level of indirection which might make it a bit harder to read. Specifically in this `end` function it isn't that much of a problem of course, but I had some hard times following the `callback` and `callbackData` in the other function (will put a comment there after I have read thru and understood all of that code, just started here ☺️)

src/utils/multipart-parser.js
```diff
@@ -0,0 +1,440 @@
...
+		} else if (this.state !== S.END) {
+			return new Error('MultipartParser.end(): stream ended unexpectedly');
```
LinusU, line 326: The return value of `end()` is never checked, I think that this was meant to `throw`?

```suggestion
			throw new Error('MultipartParser.end(): stream ended unexpectedly');
```

jimmywarting: fixed

src/utils/multipart-parser.js
jimmywarting: the hole multipart decoder is just a internal core part and never exposed to the end user, but i will take your suggestions

jimmywarting: > this.onPartEnd?.(), but since we are still supporting Node.js 12 we can't do that

yea, would also like to do that... for now i have done something like `x && x()` in other packages...

src/body.js
jimmywarting: I suppose we could replace this hole section https://github.com/node-fetch/node-fetch/blob/0f2f9bda685a0e259be50f4ba4aef9f40d38ab2b/src/utils/is.js#L48-L68 code and replace it with instance of FormData check instead...

src/body.js
```diff
@@ -302,7 +320,7 @@ export const extractContentType = (body, request) => {
 		return null;
 	}
 
-	if (isFormData(body)) {
+	if (body instanceof FormData) {
```
octet-stream, line 323: So, with that you will drop support for `formdata-node`? It seem that `formdata-polyfill/esm.min.js` does not recognize `formdata-node` as FormData instance.

<details>
<summary>Details</summary>

image

</details>

This change will break compatibility for those who already use `formdata-node` with `node-fetch` v3.

jimmywarting: Hmm, i thought i formdata-polyfill would use the symbol.hasInstance to accommodate for this.
maybe need to revert 405bcc0 after all

jimmywarting: Works now

jimmywarting: Re-added a test case for formdata-node

src/body.js
```diff
@@ -50,10 +50,10 @@ export default class Body {
...
 		} else if (body instanceof Stream) {
 			// Body is stream
-		} else if (isFormData(body)) {
-			// Body is an instance of formdata-node
-			boundary = `nodefetchformdataboundary${getBoundary()}`;
-			body = Stream.Readable.from(formDataIterator(body, boundary));
+		} else if (body instanceof FormData) {
+			// Body is an spec compatible FormData
+			body = formDataToBlob(body);
+			boundary = body.type.split('=')[1];
```
octet-stream, line 56: Do we even need a reference to `boundary`? Your `formDataToBlob` returns a `Blob` with type `multipart/form-data boundary=%boundary%`, so I think it can be just reused later.

jimmywarting: yea, kind of do... if the header class would be created first and passed down to the Body class (like how [github/fetch dose it](https://github.com/github/fetch/blob/master/fetch.js#L217-L259)) then i could assign the `content-type` but it's not possible the way it's structured right now.

both deno and undici do something like a util function that [extract body and contentType](https://fetch.spec.whatwg.org/#concept-bodyinit-extract) "10. Return body and Content-Type."
Unlike ourself where we extend a Body class. which is the "wrong thing" to do.

for instances `Response.prototype.arrayBuffer === Request.prototype.arrayBuffer` is only true in node-fetch 
not in undici, Deno or the browser is the function the same thing. Looking more into how [undici](https://github.com/nodejs/undici/blob/main/lib/fetch/body.js#L25) & [Deno](https://github.com/jimmywarting/deno/blob/2d519e9c5beb54e38c354ce65f22ecffbefdbc2f/extensions/fetch/22_body.js#L309) have implemented it more according to spec then i realize now that we have mostly done: "yea, we can do it this way and it will roughly do the same thing" so we have not followed the spec correctly

src/body.js
```diff
@@ -105,6 +105,24 @@ export default class Body {
 		return buffer.slice(byteOffset, byteOffset + byteLength);
 	}
 
+	async formData() {
+		const ct = this.headers.get('content-type');
+
+		if (ct.startsWith('application/x-www-form-urlencoded')) {
+			const fd = new FormData();
```
Richienb, line 112:
```suggestion
			const formData = new FormData();
```

Richienb (changes requested):

src/body.js
jimmywarting: Fixed

jimmywarting:
ping @LinusU @tekwiz @gr2m @Richienb if you could take a look at it then it would be grate

gr2m:
> What do you think about keeping this code in a separate package that both node-fetch and multer could depend on?

I think that'd be great, as long as it doesn't conflict with the standard compatibility of fetch

LinusU:
> Formidable have been tested with a variety of different multipart uploads and it seems pretty speedy at parsing the payload. If we keep changing it then i think we should be bringing in theirs test cases as well... I do not know how spec compatible it is.

I think that we are already working in a different way in the specific return statements that I talked about, because their `write` methods returns the number of bytes consumed, whereas our version does. So that's why I'm wary about that specific case.

Porting over the tests would be great, also adding some test case that we run both against our parser and the browser equivalent...

> > Also, I'm also maintaining Multer which parses multipart data, and I've been interested in switching out the parser to a bit more modern/simpler one. What do you think about keeping this code in a separate package that both node-fetch and multer could depend on?
> 
> I think this would be a grate fit for the [stream-consumers](https://github.com/jimmywarting/stream-consumers) it's mostly a straight out copy from undici BodyMixin that made it into node.js core. it already uses fetch-blob and is also written in ESM + that it's new so not so many breaking changes for some ppl It's something i have been wanting to bring into node-fetch anyway: #1252
> 
> It dose not have formData doe since FormData isn't implemented in NodeJs but it could be one day? there is issues about bringing in [FormData into node](https://github.com/nodejs/node/issues/38943) Also saw that @octet-stream linked undici to this PR: [nodejs/undici#974](https://github.com/nodejs/undici/issues/974)

For Multer I would probably need something a bit more low level, since we are doing field and size validation on the fly, and also stream the files to temporary files on disk 🤔 

Depends on how the API would look of course!

> Personally i would like to remove the `.buffer()` method in my stream consumer package also since i think it should be discouraged to use it anywhere outside of NodeJS. you can just as easily write `arrayBuffer(iterable).then(Buffer.from)` if you really need a `node:buffer`

👍

jimmywarting:
fyi, i 

> > Personally i would like to remove the `.buffer()` method in my stream consumer package also since i think it should be discouraged to use it anywhere outside of NodeJS. you can just as easily write `arrayBuffer(iterable).then(Buffer.from)` if you really need a `node:buffer`
> 
> 👍

I already went ahead and removed it... (put a warning sign on it)

LinusU:
@jimmywarting seems to be a conflict in `package.json`, could you fix that?

@Richienb would you be able to review this? ☺️

jacob-ebey:
Oh this is exciting. Can't wait to see this merged.

src/utils/multipart-parser.js
tunnckoCore: @LinusU yea, something like that. Every time i dig a bit deeper in Formidable's source i'm shocked what a parser that is. But well, it's here for 10 years and is working. Felix is.. ah.

tunnckoCore:
Hi there, one of the two `formidable` maintainers last years. @felixge, the author, passed the torch while back.
Happy to see his legacy merged into some even bigger module.

Still think that @jimmywarting should have added 1-2 lines of attribution and license, it's not that different than the current `latest` (v2) on npm. The [v1 one was pure and just the parser](https://github.com/node-formidable/formidable/blob/v1-legacy/lib/multipart_parser.js), then for v2 we made it Transform stream for some reason. But I like this clean Uin8Array variant more, so I'll really extract it as `formidable-multipart` or something like that, thought for decoupling Formidable for quite some time, and v3-v4 is where it can come.

jimmywarting:
> Still think that @jimmywarting should have added 1-2 lines of attribution and license,

yea, sorry about that. 
Can ofc add some 1-2 lines of attribution and license.

> @jimmywarting ha, didn't knew that node-fetch [is using formidable's multipart parser](https://github.com/node-fetch/node-fetch/blob/main/src/utils/multipart-parser.js), hehe.
>
> The only difference is that it's not Transform stream and uses uint8array. Seems like it also expose toFormData which is exactly what I thought in my previous comment LOL - thin layer on top of the parser.
>
> Do you mind to extract it as @formidable/multipart and we will both use it? Not sure if the namespace is free.

Yep, it was based on an older PR i made specifically for https://github.com/github/fetch a long time ago that dose not bring in `node:streams` or `node:buffer` as a dependency. it's was more cross compatible that way.

it kind of also has to work with the new whatwg streams, so that is why it's based on iterators instead. to support for both whatwg stream and node streams. 

----

That think i made in `fetch-blob` recently about writing files to a temporary location on the disk and then deleting it afterwards... i was think about using `createTempFile(iterable, name, opts)` instead of using `new File()`

---

Maybe like the idea of a separated module that is solo about parsing multipart payloads according to web specification. Would you like to create a new repo and maybe add me as a member?

Issue #1297, opened 2021-09-17
Happy Eyeballs


zwhitchcox:
**Is your feature request related to a problem? Please describe.**

DNS misconfigurations or server down time can cause `fetch` requests not to be fulfilled. In the browser, this is handled for you using the [Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs) algorithm.

This functionality is also implemented in curl, iOS, Android, etc..

With IPv6-only network connectivity becoming increasingly common, misconfigurations in DNS servers can cause requests to fail, even when connectivity is available through alternative servers.

Browsers typically try to create a connection, wait 300ms, then try the next available server provided by [`getaddrinfo`](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html) in the kernel.

Currently, `node-fetch` implements no failover for support for timed-out connections.

**Describe the solution you'd like**

I would propose `node-fetch` handle failover similarly to the browser. Or better yet, `curl`'s implementation.

`net.createConnection` attempts to create a connection with the default `dns.lookup` server. Btw, semi-related, the DNS lookup should be returning results with `verbatim=true`. Currently, due to a poor design decision by NodeJS developers, IPv4 addresses are preferred, even when this contradicts [RFC 6724](https://datatracker.ietf.org/doc/html/rfc6724). Note, that `verbatim=true` will probably become the default.

This can cause extremely confusing errors for developers using dual stack networks where IPv4 support periodically drops, because it's very hard to find what is causing the issue (I should know).

Ideally, `fetch` would send the first address first, wait 300ms, and if no connection was made, continue down the line of addresses returned by `dns.lookup`.

It could be a good idea to log a warning for the user that the first address sent by `getaddrinfo` was not available, as the 300ms timeout adds latency to every connection.

Still, this is better than dropping the request entirely.

In `curl`'s implementation, which I prefer, its starts with the first address, then tries the next address of a different family.

So, if you're on a IPv6-only network, you don't have to try every single IPv4 address before you get to the IPv6 address.

**Describe alternatives you've considered**

It would be great if node could implement this natively, but there are currently [no plans](https://github.com/nodejs/node/issues/6307) to do that.

Ultimately though, I think it would probably be better for the community to develop something like this and test it out before merging it into NodeJS anyway and see the response it gets. If it's extremely popular, NodeJS can merge it in, and the code can be removed from here.

zwhitchcox:
[...]

PR #1326, opened 2021-10-01
Implement Happy Eyeballs

zwhitchcox:

- [ ] Documentation update
- [x] Bug fix
- [x] New feature
- [ ] Other, please explain:

Implement happy eyeballs algorithm. For full description, see https://github.com/zwhitchcox/happy-eyeballs

#1297

The meat and potatoes of the algorithm is done here: https://github.com/zwhitchcox/happy-eyeballs/blob/master/src/happy-eyeballs.ts

#### Providing a Default User Agent

I think the cleanest way to implement the change is to provide a default agent.

That way, the existing user interface is preserved, and the user is still free to customize `fetch` requests any way they want.

This is also, the more loyal to the way the browser's `fetch` works, because happy eyeballs is implemented by the user agent, not through any other interface.

I did consider many other methods, but ultimately, I think this makes the most sense, and I think it will lead to the least confusion when debugging, as the inter-operability between various user agent methods can be complex.

This is also the most forward thinking in my opinion, because in future use cases, we will have a default user agent which can implement other desired behavior.

This also provides flexibility, all behavior is confined to on neat little package which can easily be overridden without fear of compatibility issues with other non-(NodeJS)-standard behavior.

#### Keeping Happy Eyeballs in Another NPM Package

I thought about copying the code from `happy-eyeballs` into this repository so it could have its own version of the code that it can easily change, but decided against this because

1. This isn't really a `node-fetch` specific feature, and if others want to use it outside of `node-fetch`, it creates a foundation that all other libraries can use and "speak the same language". In other words, developers won't have to fear compatibility issues with different implementations of `happy-eyeballs`, which I think provides a big benefit.
2. This will likely ultimately be a part of NodeJS (my next PR), so there's really no sense in adding all that code to this library only to remove it later. All that will need to be removed is two lines of code this way.
3. It's probably better to keep this type of thing external from `node-fetch` anyway, because it's kind of an extraneous detail, not really related to the core functionality of `node-fetch`, and I didn't want to add all that noise/bloat to the code base.

---

All that being said, I'm flexible and open to feedback/alterations that this team feels necessary, so please let me know any changes/improvements that need to be made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
   },
   "scripts": {
-    "test": "mocha",
+    "test": "mocha --async-stack-traces",
     "coverage": "c8 report --reporter=text-lcov | coveralls",
     "test-types": "tsd",
     "lint": "xo"
@@ -56,6 +56,7 @@
     "delay": "^5.0.0",
     "form-data": "^4.0.0",
     "formdata-node": "^3.5.4",
+    "happy-eyeballs": "^0.0.6",
     "mocha": "^8.3.2",
     "p-timeout": "^5.0.0",
     "tsd": "^0.14.0",
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "delay": "^5.0.0",
     "form-data": "^4.0.0",
     "formdata-node": "^3.5.4",
-    "happy-eyeballs": "^0.0.11",
     "mocha": "^8.3.2",
     "p-timeout": "^5.0.0",
     "tsd": "^0.14.0",
@@ -64,7 +63,8 @@
   },
   "dependencies": {
     "data-uri-to-buffer": "^3.0.1",
-    "fetch-blob": "^3.1.2"
+    "fetch-blob": "^3.1.2",
+    "happy-eyeballs": "^0.0.13"
   },
   "tsd": {
     "cwd": "@types",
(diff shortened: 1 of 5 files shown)


zwhitchcox:
Going to refactor this

tekwiz:
@zwhitchcox Thank you for writing this library. It looks like a very promising solution to a frustrating problem. However, I don't feel that it is wise to include this in node-fetch, particularly since the node maintainers made an active decision not to include this functionality.  That said, I think a reasonable compromise might be to include something in the README explaining the purpose of Happy Eyeballs and how to use your package.

@bitinn @jimmywarting I don't want to unilaterally make a rejection on this.  Would one of you be so kind to add your thoughts?

jimmywarting:
I'm a bit reluctant to this as well as it further complicates things and bringing in even more external dependencies
Skärmavbild 2022-01-15 kl  23 22 35

Need to review happy-eyeball for bugs and security stuff and being responsible for what goes into node-fetch

if this can be resolved via a custom user agent plugin and we can somehow mention how to use it in a readme then i would probably prefer that it stays as a external dependency. 

if we would use any happy-eyeballing stuff then i think the code should stay within node-fetch's own repository

NodeJS is also aiming on shipping fetch into core (I think/hope)

v4 is probably going to drop support for NodeJS v14.17 as v12.x will end it's LTS soon in April
So would prefer if it used built in AbortController and EventTarget that is used by NodeJS builtin instead...

PR #1333, opened 2021-10-09
Add req res data warn

### LINKED ISSUE #1000: Hint that using `data` is wrong
From time to time i see ppl using `data` instead of `body` which is wrong.
happened just recently in https://github.com/node-fetch/node-fetch/issues/481#issuecomment-717864624

how do you feel about adding a little warning message that can only appear once. (kinda like node reports that you are using experimental apis)

```js
if ('data' in bodyInit && notWarned) {
  console.warn('use body instead')
}
class Response {
  get data () {
    notWarned && console.warn('you want to use .json(), .text(), .arrayBuffer() or just .body')
  }
}
```

I don't think it is uncommon to see those mistake happening when switching from different request libraries `axios` use data instead of body for example. I can admit that i have used data as well and wasted minutes trying to figure out what's wrong

TheGLander:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Added a one-time warning for when .data is used in any way, as a way to help porting from other libraries for requests (such as `axios`).

Closes #1000


jimmywarting:
Thinking maybe if we could use https://nodejs.org/dist/latest-v16.x/docs/api/util.html#util_util_deprecate_fn_msg_code

It is a one time warning function...

jimmywarting:
I thought so too... But it's mere elegant in a way also. + We are already using it

jimmywarting:
we don't have a test for the other warning message about `res.buffer()` and the old `form-data`... but i think there is the `process.on('warning', fn)` that one could use to test something... 

https://www.geeksforgeeks.org/node-js-process-warning-event/

jimmywarting:
made some small changes to this in #1421

Issue #1335, opened 2021-10-10
follow = 0 not working


simllll:
**Reproduction**

1. 
2. 
3. 
4. 

**Expected behavior**

**Screenshots**

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       |
| node             |
| npm              |
| Operating System |

**Additional context**

PR #1336, opened 2021-10-12
update: when follow = 0 redirect manual is set

RafaelGSS:

- [ ] Documentation update
- [X] Bug fix
- [ ] New feature
- [ ] Other, please explain:

I see that it might be solved by either updating the documentation to explicitly say that `follow = 0` will throw an redirection error or make `follow = 0` not throw an error by using `redirect = manual`.

I'm comfortable with both approaches, however, the first one is smooth and is obviously is not a breaking change. I've opened this PR just to discuss the best approach. I'm open to work with the first one if needed.

#1335


jimmywarting:
If i where able to decide then the `follow` option would not be available for use anymore... 
if more granular control is needed then I think  a custom plugin should be using `manual`

so that the code can work the same way with other fetch implementations in deno, node and the browser...

PR #1346, opened 2021-10-21
refactor: use node: prefix for imports

dnalborczyk:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Other, please explain: uses node: prefix for node imports

depending on: https://github.com/node-fetch/node-fetch/pull/1348, although there's a bit more linting failing. I'll push it on top of the linting PR.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -2215,7 +2215,6 @@ describe('node-fetch', () => {
 		function lookupSpy(hostname, options, callback) {
 			called++;
 
-			// eslint-disable-next-line node/prefer-promises/dns
 			return lookup(hostname, options, callback);
 		}
 
@@ -2232,7 +2231,6 @@ describe('node-fetch', () => {
 		function lookupSpy(hostname, options, callback) {
 			families.push(options.family);
 
-			// eslint-disable-next-line node/prefer-promises/dns
 			return lookup(hostname, {}, callback);
 		}
 


mattfysh:
this appears to have broken `node-fetch` when running on AWS lambda, I don't believe they are currently running a version of node which supports the node: prefix

dnalborczyk:
> this appears to have broken `node-fetch` when running on AWS lambda, I don't believe they are currently running a version of node which supports the node: prefix

aws also does not yet support `esm` for that matter either. are you using a bundler (webpack, rollup etc.) and outputting to `commonjs`?

feross:
Question: What is the reason to do this? What does it improve? Asking because it broke my webpack build (which does not understand the `node:` prefix).

jimmywarting:
@feross 

It begun with `import streams from 'streams/web'`
... it was so new that their own build tool did not yet understand that this new module was coming from nodejs itself. it was resolved when we added a node prefix, so it did know when/if there was a core node module. https://github.com/node-fetch/fetch-blob/pull/122

So we added node prefix everywhere

v16.0.0, v14.18.0 | Added `node:` import support to require(...).
-- | --
v14.13.1, v12.20.0 | Added `node:` support: v14.13.1, v12.20.0

jimmywarting:
it looks like webpack has added support for node prefix https://github.com/webpack/webpack/pull/12693

PR #1378, opened 2021-11-11
fix(http.request): Cast URL to string before sending it to NodeJS core

### LINKED ISSUE #1376: Regression in node-fetch@3.1.0 + Electron renderer process. "TypeError: The "listener" argument must be of type function"
**Reproduction**

1. Use node-fetch@3.1 in Electron renderer process.

I've created a sample project: https://github.com/babatakao/node-fetch-3-electron-test . With this sample,

1. Clone the project.
2. `npm ci`
3. `npm run build`
4. `npm start`

Note that the project uses webpack since `node:` prefix of esm import, used by node-fetch, is not supported by Electron renderer.

**Expected behavior**

`done` alert is shown.

**Actual behavior**

image

```
TypeError: The "listener" argument must be of type function. Received an instance of Object
    at __node_internal_captureLargerStackTrace (node:internal/errors:463)
    at new NodeError (node:internal/errors:370)
    at __node_internal_ (node:internal/validators:229)
    at checkListener (node:events:131)
    at ClientRequest.once (node:events:533)
    at new ClientRequest (node:_http_client:209)
    at request (node:http:96)
    at renderer.js:111
    at new Promise (<anonymous>)
    at fetch (renderer.js:66)
    at renderer.js:7025
    at renderer.js:7030
    at renderer.js:7032
```

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.1.0 |
| node             | 16.13.0 (Electron 15 uses Node 16.5) |
| npm              | 7.20.6 |
| Operating System | Windows 10 x64 |

**Additional context**

The issue does not occur on node-fetch@3.0.0. Thus it seems a recent regression.

I've found the root cause. At [here](https://github.com/node-fetch/node-fetch/blob/v3.1.0/src/request.js#L51), `new URL` constructor references [NodeJS URL](https://nodejs.org/api/url.html) in most cases. However, it references [Chromium renderer native URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) in the Electron renderer process. This causes type mismatch: NodeJS http library checks that whether the URL has `searchParamsSymbol` or not at https://github.com/nodejs/node/blob/v16.13.0/lib/_http_client.js#L117.

I suspect that [this](https://github.com/node-fetch/node-fetch/blob/v3.1.0/src/request.js#L10) should be `import {URL, format as formatUrl} from 'node:url';` 🤔

jimmywarting:
## Purpose
In #1268 we started to just forward the URL to node's Request method instead of sending everything as an option, This broke something as we where not passing it as a string.

## Changes
I casted the URL into a string before handing the url to NodeJS http.request method
`send(parsedURL.toString(), options);`

___

- Fix #1376

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,7 +78,7 @@ export default async function fetch(url, options_) {
 		};
 
 		// Send request
-		const request_ = send(parsedURL, options);
+		const request_ = send(parsedURL.toString(), options);
 
 		if (signal) {
 			signal.addEventListener('abort', abortAndFinalize);
--- a/src/request.js
+++ b/src/request.js
@@ -1,4 +1,3 @@
-
 /**
  * Request.js
  *
@@ -21,7 +20,7 @@ const INTERNALS = Symbol('Request internals');
 /**
  * Check if `obj` is an instance of Request.
  *
- * @param  {*} obj
+ * @param  {*} object
  * @return {boolean}
  */
 const isRequest = object => {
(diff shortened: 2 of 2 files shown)


gr2m (approved):
Changes look good to me 👍🏼

LinusU (approved):
Neat 👍

PR #1387, opened 2021-11-18
fix(Redirect): Better handle wrong redirect header in a response

### LINKED ISSUE #1386: Uncatchable error when server redirects to invalid URL
**Reproduction**

1. Request to a valid URL to which the server responds with an invalid Location header
2. Server crashes despite `await fetch()` being wrapped in `try { } / catch () { }`

```
Uncaught TypeError [ERR_INVALID_URL]: Invalid URL: //super:invalid:url%/
      at onParseError (internal/url.js:279:9)
      at new URL (internal/url.js:355:5)
      at ClientRequest.<anonymous> (file:///home/tasinet/code/node-fetch/src/index.js:133:52)
      at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:647:27)
      at HTTPParser.parserOnHeadersComplete (_http_common.js:126:17)
      at Socket.socketOnData (_http_client.js:515:22)
      at addChunk (internal/streams/readable.js:290:12)
      at readableAddChunk (internal/streams/readable.js:265:9)
      at Socket.Readable.push (internal/streams/readable.js:204:10)
      at TCP.onStreamRead (internal/stream_base_commons.js:188:23)
```

An example URL with this behaviour is `http://vardakeio.gr`:
```
curl -v http://vardakeio.gr 2>&1 | grep Location
< Location: https://vardakeio.gr%/
```

**Expected behavior**

Invalid URL should be caught and an appropriate error returned

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.1.0
| node             | 14.17.6
| npm              | 6.14.15
| Operating System | Ubuntu 20.04.2 LTS

**Additional context**

This is because the location header is passed to `new URL()` without a try/catch block around it:

index.js:133
```
// HTTP fetch step 5.3
const locationURL = location === null ? null : new URL(location, request.url);
```

Raising a PR to address this.

tasinet:
## Purpose
Fixes crash when a Location header point to an invalid URL

## Changes
Just wrapping `new URL()` in a try/catch and returning a sensible (?) error message

## Additional information
Test included

___

- [x] I added unit test(s)

___

- Fixes #1386

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -133,7 +133,7 @@ export default async function fetch(url, options_) {
 				let locationURL;
 				try {
 					locationURL = location === null ? null : new URL(location, request.url);
-				} catch (_error) {
+				} catch {
 					// error can only be invalid URL
 					reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
 					finalize();
--- a/src/index.js
+++ b/src/index.js
@@ -133,7 +133,8 @@ export default async function fetch(url, options_) {
 				let locationURL;
 				try {
 					locationURL = location === null ? null : new URL(location, request.url);
-				} catch (e) {
+				} catch (error) {
+					// error can only be invalid URL
 					reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
 					finalize();
 					return;
--- a/src/index.js
+++ b/src/index.js
@@ -130,14 +130,18 @@ export default async function fetch(url, options_) {
 				const location = headers.get('Location');
 
 				// HTTP fetch step 5.3
-				let locationURL;
+				let locationURL = null;
 				try {
 					locationURL = location === null ? null : new URL(location, request.url);
 				} catch {
-					// error can only be invalid URL
-					reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
-					finalize();
-					return;
+					// error here can only be invalid URL in Location: header
+					// do not throw when options.redirect == manual
+					// let the user extract the errorneous redirect URL
+					if (request.redirect !== 'manual') {
+						reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
+						finalize();
+						return;
+					}
 				}
 
 				// HTTP fetch step 5.5
--- a/src/index.js
+++ b/src/index.js
@@ -133,7 +133,7 @@ export default async function fetch(url, options_) {
 				let locationURL;
 				try {
 					locationURL = location === null ? null : new URL(location, request.url);
-				} catch (error) {
+				} catch (_error) {
 					// error can only be invalid URL
 					reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
 					finalize();
--- a/src/index.js
+++ b/src/index.js
@@ -130,7 +130,14 @@ export default async function fetch(url, options_) {
 				const location = headers.get('Location');
 
 				// HTTP fetch step 5.3
-				const locationURL = location === null ? null : new URL(location, request.url);
+				let locationURL;
+				try {
+					locationURL = location === null ? null : new URL(location, request.url);
+				} catch (e) {
+					reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
+					finalize();
+					return;
+				}
 
 				// HTTP fetch step 5.5
 				switch (request.redirect) {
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -130,7 +130,15 @@ export default async function fetch(url, options_) {
 				const location = headers.get('Location');
 
 				// HTTP fetch step 5.3
-				const locationURL = location === null ? null : new URL(location, request.url);
+				let locationURL;
+				try {
+					locationURL = location === null ? null : new URL(location, request.url);
+				} catch (error) {
+					// error can only be invalid URL
+					reject(new FetchError(`uri requested responds with an invalid redirect URL: ${location}`, 'invalid-redirect'));
```
LinusU, line 138: Have we looked at what error browsers gives back here? 🤔

LinusU: Chrome:

Screenshot 2021-11-18 at 16 37 18

Safari:

Screenshot 2021-11-18 at 16 37 43

Right, browsers doesn't give any more info on errors 😅

LinusU (approved):
LGTM 👍

src/index.js
jimmywarting: Note of caution, Had some errors in the past where we don't return... 

```diff
- reject(...)
+ return reject(...)
```

jimmywarting: Thinking if we should throw a `TypeError` here as well to match what the browser gives back... but obvious with a better error message.

`new TypeError('Failed to fetch; Redirect url is not valid, Got: ' + location)`

LinusU: In that case I think it's true for every `FetchError`, no? 🤔

jimmywarting: > In that case I think it's true for every `FetchError`, no? 🤔

pretty much 😅

tasinet: From what I can tell, in node-fetch TypeErrors are mostly "programmer error" and FetchErrors are "remote/response error" 

Some TypeErrors:

node-fetch cannot load ${url}. URL scheme "${parsedURL.protocol.replace(/:$/, '')}" is not supported.');
Redirect option '${request.redirect}' is not a valid value of RequestRedirect'));
body used already for: ${data.url}');
Invalid referrerPolicy: ${referrerPolicy}');
Invalid referrerPolicy: ${policy}');
${parsedURL} is an url with embedded credentails.');
Request with GET/HEAD method cannot have body');
Expected signal to be an instanceof AbortSignal or EventTarget');
Header name must be a valid HTTP token [${name}]');
Invalid character in header content ["${name}"]');
Header pairs must be iterable');
Each header pair must be an iterable object');
Each header pair must be a name/value tuple');
Failed to construct \'Headers\': The provided value is not of type \'(sequence<sequence<ByteString>> or record<ByteString, ByteString>)');
Failed to fetch');
no or bad content-type header, no multipart boundary');

Some FetchErrors:

**Invalid response body while trying to fetch ${data.url}: ${error.message}**
Invalid response body while trying to fetch ${this.url}: ${error_.message}
content size at ${data.url} over limit: ${data.size}
Could not create Buffer from response body for ${data.url}: ${error.message}
Premature close of server response while trying to fetch ${data.url}

Whereas in the browser there only seem to be [AbortError and TypeErrors](https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions)

src/index.js
```diff
@@ -130,7 +130,15 @@ export default async function fetch(url, options_) {
 				const location = headers.get('Location');
 
 				// HTTP fetch step 5.3
-				const locationURL = location === null ? null : new URL(location, request.url);
+				let locationURL;
+				try {
+					locationURL = location === null ? null : new URL(location, request.url);
+				} catch (_error) {
```
LinusU, line 136: The catch binding is optional on all versions of Node.js that we support:

```suggestion
				} catch {
```

src/index.js
LinusU: I agree, but I think that we should fix that in a separate pull request? I think that this one is ready for merging as soon as the linting error is fixed?

jimmywarting:
Just thought of something... and i tested this in chrome...

if you have `redirect: 'manual'` then it shouldn't follow the redirect header and try to fetch it... then you may think that it should not throw any error for having invalid headers but it dose...

`fetch(/redirect/301/invalid', { redirect: 'manual' })`

so i tested it in firefox, and safari also

```js
// FF OK
fetch('http://localhost:58555/redirect/301/invalid', { redirect: 'manual' }).then(console.log, console.error)
// Safari OK
fetch('http://localhost:58555/redirect/301/invalid', { redirect: 'manual' }).then(console.log, console.error)
// Chrome Throws
fetch('http://localhost:58555/redirect/301/invalid', { redirect: 'manual' }).then(console.log, console.error)
```

We should have a test case for both.
would you mind
- Adding unit test for when `redirect` is set to `manual` and dose not throw
- Resolve the conflicts

Should also report this to Chrome / WPT

jimmywarting:
Reported it to chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1271920

jimmywarting (changes requested):
need a redirect=manual test that don't throw

tasinet:
Do you need anything else from my end here?

LinusU (approved):
LGTM

jimmywarting (approved):

PR #1392, opened 2021-11-18
fix: handle errors from the request body stream

mdmitry01:
## Purpose
`node-fetch` does not handle errors from the request body stream correctly. This PR fixes it.

## Changes
[`body.pipe(dest)`](https://github.com/node-fetch/node-fetch/blob/0284826de6e733c717447c6dfcddc5f0b538b254/src/body.js#L390) has been replaced with the `pipeline` function from the `stream` module.

___

- [x] I added unit test(s)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes will be recorded here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+* fix: handle errors from the request body stream by @mdmitry01 in https://github.com/node-fetch/node-fetch/pull/1392
+
 ## 3.1.0
 
 ## What's Changed
--- a/src/index.js
+++ b/src/index.js
@@ -291,7 +291,7 @@ export default async function fetch(url, options_) {
 			resolve(response);
 		});
 
-		// eslint-disable-next-line prefer-await-to-then
+		// eslint-disable-next-line promise/prefer-await-to-then
 		writeToStream(request_, request).catch(reject);
 	});
 }
--- a/src/index.js
+++ b/src/index.js
@@ -291,6 +291,7 @@ export default async function fetch(url, options_) {
 			resolve(response);
 		});
 
+		// eslint-disable-next-line prefer-await-to-then
 		writeToStream(request_, request).catch(reject);
 	});
 }
--- a/src/index.js
+++ b/src/index.js
@@ -291,7 +291,7 @@ export default async function fetch(url, options_) {
 			resolve(response);
 		});
 
-		writeToStream(request_, request);
+		writeToStream(request_, request).catch(reject);
 	});
 }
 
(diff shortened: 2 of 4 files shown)


test/main.js
```diff
@@ -1456,6 +1456,22 @@ describe('node-fetch', () => {
...
+			method: 'POST',
+			body: requestBody
+		};
+		const expectedError = new Error('an error');
+		setImmediate(() => {
+			requestBody.emit('error', expectedError);
+		});
+		return fetch(url, options).catch(error => {
+			expect(error).to.equal(expectedError);
+		});
```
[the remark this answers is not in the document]
mdmitry01: if `fetch` doesn't reject, the test fails with the following error:
`Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves`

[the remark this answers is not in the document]
LinusU: Are you sure about that? If `fetch` resolves instead of rejects then the test will be considered completed? Since `done` isn't explicitly called within the `catch` function...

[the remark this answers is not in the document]
mdmitry01: you are right, fixed

src/index.js
```diff
@@ -291,7 +291,7 @@ export default async function fetch(url, options_) {
 			resolve(response);
 		});
 
-		writeToStream(request_, request);
+		writeToStream(request_, request).catch(reject);
```
LinusU, line 294:
```suggestion
		// eslint-disable-next-line prefer-await-to-then
		writeToStream(request_, request).catch(reject);
```

test/main.js
```diff
@@ -1456,6 +1456,21 @@ describe('node-fetch', () => {
...
 	});
 
+	it('should reject if the request body stream emits an error', () => {
+		const url = `${base}inspect`;
+		const requestBody = new stream.PassThrough();
+		const options = {
+			method: 'POST',
+			body: requestBody
+		};
+		const errorMessage = 'request body stream error'
```
LinusU, line 1466:
```suggestion
		const errorMessage = 'request body stream error';
```

src/index.js
```diff
@@ -291,7 +291,8 @@ export default async function fetch(url, options_) {
 			resolve(response);
 		});
 
-		writeToStream(request_, request);
+		// eslint-disable-next-line prefer-await-to-then
```
LinusU, line 294:
```suggestion
		// eslint-disable-next-line promise/prefer-await-to-then
```

LinusU (approved):
Neat! 👍

jimmywarting:
hmm, this is something i know is going to change later at some point when we start to support whatwg streams...

jimmywarting (approved):

Yasin1413:
docs/CHANGELOG.md

docs/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes will be recorded here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
```
Yasin1413, line 7: #1392

Yasin1413 (commented):
> #1392

PR #1402, opened 2021-11-30
core: Don't use buffer to make a blob

### LINKED ISSUE #1401: blob() method shouldn't use buffer() method
body.blob() use body.buffer() inside so I see the this error log even if I didn't use body.buffer().
I think it should be replaced to consumeBody() like other method (body.text(), body.json())
```
[node-fetch#buffer] DeprecationWarning: Please use 'response.arrayBuffer()' instead of 'response.buffer()'
```

jimmywarting:
## Purpose
It was logging a warning for using deprecated `body.buffer()` implicit when calling `body.blob()`

## Changes
construct blob using a arrayBuffer instead

___

- Fix #1401


LinusU (approved):
👍

PR #1405, opened 2021-12-05
update readme for TS @types/node-fetch

### LINKED ISSUE #1396: Documentation change suggestion: @types/node-fetch deprecated, delivers no types. Suggest "@types/node-fetch@2.5.12" in docs
**Is your feature request related to a problem? Please describe.**

This would be a request to change the documentation. 

`@types/node-fetch` is now (understandably) deprecated as they're wrapped up in this project.

However, when using older versions of `node-fetch` (`^2.6.6`), also trying to use the latest release of `@types/node-fetch` (`v3.0.2`), we're presented with the problem of not actually being delivered any types! 

Going back to `v3.0.0` of `@types/node-fetch` all of the `types` are `null`, so it would have to explicitly reference the last version known to provide types: `v2.5.12`

The error this produces in TypeScript is:

```
Could not find a declaration file for module 'node-fetch'. 
'/<project path>/node_modules/node-fetch/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/node-fetch` if it exists or add a new declaration (.d.ts) file containing `declare module 'node-fetch';`ts(7016)
```

which happens even when `@types/node-fetch` is installed

**Describe the solution you'd like**

- Current Documentation that would change:
https://github.com/node-fetch/node-fetch#typescript

changing this:

```sh
npm install --save-dev @types/node-fetch
```

to:

```sh
npm install --save-dev @types/node-fetch@2.5.12
```

**Additional context**

- I am in an environment where upgrading to the current version of `node-fetch` isn't currently viable.
- Additionally, I use `"noImplicitAny": true,` in my .tsconfig which may also be a factor here as per the err above
  - Just for funsies I disabled `noImplicitAny` and the same issue still arises

Thank you for your time considering this change; if you'd like me to submit a PR upon approval of this change please let me know.

adamellsworth:
## Purpose

Closes #1396

## Changes

Updates Readme such that the `@types/node-fetch` refs the last known version with valid types (`@types/node-fetch@2.x`); targets users who cannot upgrade to `node-fetch@3.x`

## Additional information

n/a
___

- [x] I updated readme

> __Note to reviewers of this PR:__
> because of the minuscule nature of this change I have not updated the Changelog – I feel such an addition would add a bit of unnecessary "noise" to the real additions to the codebase.
> also, ty for writing such an amazing lib <3

### DIFF
--- a/README.md
+++ b/README.md
@@ -750,7 +750,7 @@ An Error thrown when the request is aborted in response to an `AbortSignal`'s `a
 For older versions please use the type definitions from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped):
 
 ```sh
-npm install --save-dev @types/node-fetch
+npm install --save-dev @types/node-fetch@2.x
 ```
 
 ## Acknowledgement


jimmywarting (approved):

LinusU (approved):
👍

Issue #1279, opened 2021-09-08
Loading node-fetch@3 in cjs/esm


jimmywarting:
# Pure ESM package

**node-fetch** is an ESM-only module - you are not able to import it with require. You don't necessary have to convert your hole project to ESM just b/c we did it. There are ways to load ESM modules in cjs package too (more about this in FAQ). You can also stay with the v2 branch - **We will keep updating v2 with bug/security issues**. But not so much with new features...

This means you have the following choices:

1. Use ESM yourself. **(preferred)**\
  Use `import fetch from 'node-fetch'` instead of `const fetch = require('node-fetch')` to import the package. You also need to put `"type": "module"` in your package.json and more. Follow the below guide.
2. you could use async [`import(…)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports) from CommonJS instead of `require(…)`. This is supported in CommonJS also
3. Stay on the existing version of the package until you can move to ESM.

**You also need to make sure you're on the latest minor version of Node.js. At minimum Node.js 12.20, 14.14, or 16.0.**

I would strongly recommend moving to ESM. ESM can still import CommonJS packages, but CommonJS packages cannot import ESM packages synchronously.

ESM is natively supported by Node.js 12.17 and later.

### How can I move my CommonJS project to ESM?

- Add `"type": "module"` to your package.json.
- Replace `"main": "index.js"` with `"exports": "./index.js"` in your package.json.
- Update the `"engines"` field in package.json to Node.js 12: `"node": "^12.20.0 || >=14.13.1"`.
- Remove `'use strict';` from all JavaScript files.
- Replace all `require()`/`module.export` with `import`/`export`.
- Use only full relative file paths for imports: `import x from '.';` → `import x from './index.js';`.
- If you have a TypeScript type definition (for example, `index.d.ts`), update it to use ESM imports/exports.
- Optional but recommended, use the [`node:` protocol](https://nodejs.org/api/esm.html#esm_node_imports) for imports.

### Can I import ESM packages in my CJS project?

Yes, but only asynchronous. The `import('node-fetch')` function is supported in CJS contexts too.
if you would like you can create a wrapper function:
```js
const fetch = (...args) => import('node-fetch').then(({default: fetch}) => fetch(...args));
```
(the next request won't re-import node-fetch cuz modules gets cached)
This can make the Node process boot up faster and only lazy loads the node-fetch when it's needed

[...]

PR #1412, opened 2021-12-12
chore: use Symbol.toStringTag in isAbortSignal check

### LINKED ISSUE #751: fetch doesn't allow for AbortSignal custom impl with constructor name other then exactly AbortSignal

JohanWillfred:
## Purpose
2.x.x version `isAbortSignal` is not always working correctly with code processed by webpack, and for implementations  [with a constructors custom name](https://github.com/node-fetch/node-fetch/issues/751).  This minor change should be useful for people who cannot migrate to 3.x.x version right now.

## Changes
This PR basically extends `isAbortSignal` with the same function's code from 3-rd version of the project.

## Additional information
I tried to cover it with tests, but I failed :( 

___

- [ ] I added unit test(s)

___

- Fixes #751

___

Authors, maintainers, and contributors of the project, thank you. You are awesome ❤️

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -15,6 +15,7 @@ import Body, { clone, extractContentType, getTotalBytes } from './body';
 
 const INTERNALS = Symbol('Request internals');
 const URL = Url.URL || whatwgUrl.URL;
+const NAME = Symbol.toStringTag;
 
 // fix an issue where "format", "parse" aren't a named export for node <10
 const parse_url = Url.parse;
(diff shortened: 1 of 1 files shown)


LinusU (approved):
LGTM 👍

jimmywarting:
is this necessary?

You could either:
- Update to node-fetch@3
- opt out of not touching the node_modules with webpack and just compile your own code.
- use built in global AbortController (and signal) got introduced into node v15.0.0 instead of some polyfill
- you could also opt to not mangle certain names

JohanWillfred:
> is this necessary?

It definitely is not necessary. For my problem, I disabled webpack module concatenation, and it works fine now. (I just stuck with node 12 for some time, so I had to use the polyfill). I will completely understand if you'd decide to close this PR.
I was mainly thinking that these minor changes can save someone a couple of hours of investigations / fixes.

Also, thank you for putting all the migration guidance and instructions together. #1279 helped me and many other developers a lot.

PR #1457, opened 2022-01-16
test: Update major busboy version

jimmywarting:
## Purpose
dependabot wasn't able to update busboy cuz it had breaking changes...

## Changes
Only affect the test server...

## Additional information
I was thinking maybe we don't really need the busboy...? we could just simply echo everything back and utilize node-fetch new `body.formData()` parsers.
and then do something like:
```js
fetch('/multipart', { method: 'post', body: formData })
  .then(res => res.formData())
  .then(formData => {
    // expect responded formData to match the posted formData
  })
```
Dos removing the dependency on busboy... 
___

- [ ] Decide if we want to keep busboy as a dependency...  if not: 
  - [ ] use `res.formData()` to decode it.

___

- closes #1437

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "devDependencies": {
     "abort-controller": "^3.0.0",
     "abortcontroller-polyfill": "^1.7.1",
-    "busboy": "^0.3.1",
+    "busboy": "^1.4.0",
     "c8": "^7.7.2",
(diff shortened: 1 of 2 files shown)


gr2m (approved):

LinusU (approved):
Neat 👍

PR #1463, opened 2022-01-18
feat: export Blob, File and FormData + utilities

### LINKED ISSUE #1454: Expose more stuff (Blob, File & FormData)
In other help section i have mention ways of posting a FormData, Blob, and a File

Some solutions involves adding same extra dependency on the same thing that node-fetch also depends on and other solutions involves more hacky stuff to get the internals of what node-fetch is using internally.

like: 
```js
import fetch, { Headers, Request, Response } from 'node-fetch'

// Get the internals fetch depends on
const { constructor: Blob } = await new Response().blob()
const { constructor: FormData } = await new Response(new URLSearchParams).formData()
const fd = new FormData(); fd.set('a', new Blob())
const { constructor: File } = fd.get('a')

const file = new File(['hello world'], 'sample.txt', { type: 'text/plain' })
fd.set('a', file)
fetch(url, { method: 'POST', body: fd })
```

it feels very hacky and unnecessary to go around the problem of using the same dependency as node-fetch is using. 
It also requires top level await to work...

I think that we should expose this 3 classes as well so you can just do:

```js
import fetch, { Headers, Request, Response, FormData, Blob, File } from 'node-fetch'

const file = new File(['hello world'], 'sample.txt', { type: 'text/plain' })
const fd = new FormData()
fd.set('a', file)
fetch(url, { method: 'POST', body: fd })
```
it would also be worth exposing the `fetch-blob/from.js` utility functions that turns a file into a Blob or a File without actually reading the hole content into memory... 

i think this would simplify the developers experiences and also making it easier to keep things up to date and avoiding dedupe and using different versions

---

For the v4 milestone, would it also be worth exposing the whatwg Readable/Wrtiable Stream as well?
(So that we take care of exposing nodes builtin whatwg stream whenever possible and falling back to web-streams-polyfill otherwise)

```js
import { stream } from 'node-fetch'
const { TransformStream, ReadableStream, WritableStream, ... } = stream
```
maybe it's too much exposure?
it is also possible to do the same hack here (but you only get ReadableStream)
```js
const ReadableStream = new Response('').body.constructor
const rs = new ReadableStream({ /* ... */ })
fetch(url, { method: 'POST', body: rs })
```
it would be nice to have the same instances of streams so they can be piped to...
piping from a native ReadableStream to a polyfilled WritableStream don't work well together...

jimmywarting:
## Purpose
Making it easier to import and use `FormData, Blob & File`. Installing same dependency as `node-fetch` or otherwise getting them in a round-about way was hacky.

## Changes
This re-export the stuff so it is available to use

___

- [ ] I updated readme
- [x] I added unit tests (for FormData, Blob, and fileFrom only) 

___

- fix #1454

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,10 +21,12 @@ import Request, {getNodeRequestOptions} from './request.js';
 import {FetchError} from './errors/fetch-error.js';
 import {AbortError} from './errors/abort-error.js';
 import {isRedirect} from './utils/is-redirect.js';
+import {FormData} from 'formdata-polyfill/esm.min.js';
 import {isDomainOrSubdomain} from './utils/is.js';
 import {parseReferrerPolicyFromHeader} from './utils/referrer.js';
 
-export {Headers, Request, Response, FetchError, AbortError, isRedirect};
+export {FormData, Headers, Request, Response, FetchError, AbortError, isRedirect};
+export * from 'fetch-blob/from.js';
 
 const supportedSchemas = new Set(['data:', 'http:', 'https:']);
 
--- a/src/index.js
+++ b/src/index.js
@@ -24,9 +24,17 @@ import {isRedirect} from './utils/is-redirect.js';
 import {FormData} from 'formdata-polyfill/esm.min.js';
 import {isDomainOrSubdomain} from './utils/is.js';
 import {parseReferrerPolicyFromHeader} from './utils/referrer.js';
+import {
+	Blob,
+	File,
+	fileFromSync,
+	fileFrom,
+	blobFromSync,
+	blobFrom
+} from 'fetch-blob/from.js';
 
 export {FormData, Headers, Request, Response, FetchError, AbortError, isRedirect};
-export * from 'fetch-blob/from.js';
+export {Blob, File, fileFromSync, fileFrom, blobFromSync, blobFrom}
 
 const supportedSchemas = new Set(['data:', 'http:', 'https:']);
 
--- a/src/index.js
+++ b/src/index.js
@@ -34,7 +34,7 @@ import {
 } from 'fetch-blob/from.js';
 
 export {FormData, Headers, Request, Response, FetchError, AbortError, isRedirect};
-export {Blob, File, fileFromSync, fileFrom, blobFromSync, blobFrom}
+export {Blob, File, fileFromSync, fileFrom, blobFromSync, blobFrom};
 
 const supportedSchemas = new Set(['data:', 'http:', 'https:']);
 
--- a/test/form-data.js
+++ b/test/form-data.js
@@ -1,8 +1,6 @@
 import {FormData as FormDataNode} from 'formdata-node';
-import {FormData} from 'formdata-polyfill/esm.min.js';
-import {Blob} from 'fetch-blob/from.js';
 import chai from 'chai';
-import {Request, Response} from '../src/index.js';
+import {Request, Response, FormData, Blob} from '../src/index.js';
 
 const {expect} = chai;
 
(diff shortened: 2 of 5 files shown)


src/index.js
```diff
@@ -21,10 +21,12 @@ import Request, {getNodeRequestOptions} from './request.js';
 import {FetchError} from './errors/fetch-error.js';
 import {AbortError} from './errors/abort-error.js';
 import {isRedirect} from './utils/is-redirect.js';
+import {FormData} from 'formdata-polyfill/esm.min.js';
 import {isDomainOrSubdomain} from './utils/is.js';
 import {parseReferrerPolicyFromHeader} from './utils/referrer.js';
 
-export {Headers, Request, Response, FetchError, AbortError, isRedirect};
+export {FormData, Headers, Request, Response, FetchError, AbortError, isRedirect};
+export * from 'fetch-blob/from.js';
```
jimmywarting, line 29: I'm a bit unsure abobut this wildcard export doe... what do u think?

LinusU: Hmm, yeah, I would _probably_ prefer to explicitly list them

test/form-data.js
```diff
@@ -1,8 +1,6 @@
 import {FormData as FormDataNode} from 'formdata-node';
-import {FormData} from 'formdata-polyfill/esm.min.js';
-import {Blob} from 'fetch-blob/from.js';
 import chai from 'chai';
-import {Request, Response} from '../src/index.js';
+import {Request, Response, FormData, Blob} from '../src/index.js';
```
LinusU, line 3: Not really related, but I think that this can import from `node-fetch` with ESM, which will import the local package:

```suggestion
import {Request, Response, FormData, Blob} from 'node-fetch';
```

LinusU (approved):
Neat 👍

test/form-data.js
jimmywarting: umm, i think i rather prefer relative import
Thinking that it's easier for the path resolver to quicker lookup the file/module rather than going via some package.json lookup

LinusU: I can't imagine that would even make a milliseconds difference? The nice thing about using `from 'node-fetch'` is that you also test that the public importable API is how you expect it. E.g. it tests that your `exports` is setup correctly ☺️

jimmywarting: "många bäckar små" as we say in Sweden... for one file it might not make a differences. but when when you have many dependencies like [this](https://npmgraph.js.org/?q=xo) then it adds up. I bet if every file actually used the extension in each and every file then it would not have to 2nd guess what you meant to import.

relative import is also more remote friendlier as it would not have access to the filesystem to scan for things. then you would need import-maps

https://nodejs.org/dist/latest-v17.x/docs/api/modules.html#all-together
https://github.com/nodejs/node/blob/34e3dd50348dd824c881c183824237532c7721c5/lib/internal/modules/cjs/loader.js#L346

I know that unix is quite good at handling many small files and stat-ing each and every thing relative fast but windows might struggle with [the cost of many small modules](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/)

jimmywarting:
☝️  Neat

gulshan:
Hi, vscode/typescript cannot see these newly exported types-
image

Can you have a look?

jimmywarting:
@gulshan Already got it covered in https://github.com/node-fetch/node-fetch/pull/1465
i thought i would document it in the next release when it was released so that the readme would not show something that haven't been shipped yet. But now that we have auto deployment with auto releasees then i think things will change around here

PR #1464, opened 2022-01-20
docs: remove the changelog

### LINKED ISSUE #1372: Delete the CHANGELOG.md
@node-fetch/core what do you think about ditching the CHANGELOG.md?

Everytime we release a new tag, then we have the possibility to auto generate a changelog from all PR/commits that where added. 
that way we don't have to manually keep the changelog up to date.

jimmywarting:
## Purpose
with semantic release bot that can auto generate release changes then we no longer need this changelog.
also update the PR template to reflect this, `Fixes` wasn't a valid command to auto close issues... 

___

- fix #1372


Narretz:
Imo that's a bad idea. Release notes directly in github are fine, but a single changelog file is easier to search and also available outside of github.

jimmywarting:
I think it's pretty easy to see all the changes in https://github.com/node-fetch/node-fetch/releases devided into a own section

Narretz:
True, the release page looks pretty good. I hadn't checked them in a while and Github has improved them. The other point still stands that the release notes are not available when I have the repository checked out.

LinusU (approved):
> The other point still stands that the release notes are not available when I have the repository checked out.

Since the changelog is generated from the Git log one could argue that the changelog is available by running `git log`.

In any way, I don't think that having a changelog file available when checking out the repo is enough upside to warrant all the extra work that it puts on us...

stevenschmatz:
Funnily enough, I just had to search the changelog today due to dependabot alerts, and it was a harder experience with releases. For example, with a full changelog open I can easily see chronological order whereas its harder in releases. Also, I tried searching `2.6.1` which for some reason had no results, even though v2.6.1 was on the page.

Narretz:
Can't the semantic-release-bot update a file with the changes? It's touching the repo anyway, right?

jimmywarting:
> Can't the semantic-release-bot update a file with the changes? It's touching the repo anyway, right?

Maybe @gr2m can answer that question.
Maybe Github should have a better search feature in Releases.

gr2m:
the nice thing about using github tags and releases is that `semantic-release` does not need to push any own commits. 

If you want a changelog file, there is a plugin for that:
https://github.com/semantic-release/changelog

I agree the best would be a better search for releases built into github.com

@Belco90 built https://octoclairvoyant.vercel.app/comparator which is pretty great! I use it for that purpose sometimes

jimmywarting:
> @Belco90 built https://octoclairvoyant.vercel.app/comparator 

Cool!

trygve-aaberge-adsk:
Uh, how am I supposed to see which changes happened in 3.0.0 and what I have to do to upgrade with this deleted? The 3.0.0 and 3.0.0-beta releases mostly just links to the deleted changelog.

PR #1465, opened 2022-01-20
docs: update formdata example

jimmywarting:
## Purpose
Now that we are exporting FormData, Blob & Files then i thought the code examples needed to be updated.

___

- [x] I updated readme
- [x] I added the exports to `.d.ts`


LinusU (approved):
👍

PR #1471, opened 2022-01-23
chore: move support and features to discussion

jimmywarting:
## Purpose
Got tierd of looking trough all issue and trying to have some kind of organisation as to what can be worked on. Many things are just question. 
I want `Issues` to be well... actual issues that effects the node-fetch code base.

So i wanted `question` to be move into the Q&A section in discussion.
While at it i thought that `Feature request` could also be moved to `Ideas` in the Discussion forum as well. And when a idea becomes something that we agree on implementing and we have an actual plan on how to implement something then it can be turned into a issue that we can be assigned to someone to work on...

## Changes
- `Question` & `Feature request` are moved to Discussion from here on...


LinusU (approved):
Great 👍

PR #1473, opened 2022-01-23
fix: Use `keep-alive` by default for `Connection` header.

regseb:
## Purpose

Use `keep-alive` by default for `Connection` header.

## Changes

Replace `close` by `keep-alive`.

## Additional information

Chromium and Firefox use `keep-alive` by default for `Connection` header.
[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection) specifies `close` is the default value for _HTTP 1.0_. And usually, it's `keep-alive` for _HTTP 1.1_ (version used by [Node.js](https://github.com/nodejs/node/blob/master/lib/_http_client.js#L289)).

___

- [ ] I updated ./docs/v3-UPGRADE-GUIDE
- [x] I updated readme
- [ ] I added unit test(s)

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,7 +539,7 @@ If no values are set, the following request headers will be sent automatically:
 | ------------------- | ------------------------------------------------------ |
 | `Accept-Encoding`   | `gzip,deflate,br` _(when `options.compress === true`)_ |
 | `Accept`            | `*/*`                                                  |
-| `Connection`        | `close` _(when no `options.agent` is present)_         |
+| `Connection`        | `keep-alive` _(when no `options.agent` is present)_         |
 | `Content-Length`    | _(automatically calculated, if possible)_              |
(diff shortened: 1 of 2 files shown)


AlttiRi:
However, will it have any effect on the TCP connection?

I think just setting the `Connection` header to `keep-alive` will not have real effect **on the connection** — it's not enough to just set the header for using the same connection for different `fetch` calls).

[README.md#default-headers](https://github.com/node-fetch/node-fetch/blob/6875205c1a6eff1ce89ae762b1636cf3cd9fb957/README.md#default-headers)
[README.md#custom-agent](https://github.com/node-fetch/node-fetch/blob/6875205c1a6eff1ce89ae762b1636cf3cd9fb957/README.md#custom-agent)

dhedey:
I suggest this should be closed, as it has been obsoleted by #1736 👍

PR #1484, opened 2022-01-27
chore: replace headers class

### LINKED ISSUE #1119: The new Headers implementation should be reverted
Hi, I am a former maintainer of node-fetch. I noticed that #834 (merged last year) changed the implementation of the Headers class so that it is based on Proxy and URLSearchParams. It does not advertise to fix any bugs, nor does it add any features – readability improvements being the main motivation of the PR.

In the aftermath, we found that the code in #834 introduced various bugs like #917 and #1072 – the former being especially heinous as it depends on an undocumented bug in Node.js' URLSearchParams implementation. Additionally, it introduces a raft of behaviors that are incompatible with browsers' implementation of the class:

```js
{
  const h = new Headers();
  h.get = () => { console.log('yes!'); };
  h.set = () => { console.log('yes!'); };
  h.get('abc')    // prints yes!
  h.set('abc', 1) // does not print yes!
}

console.log(new Headers() instanceof URLSearchParams);
  // prints true; should be false
console.log(Headers.prototype.values instanceof (function*(){}).constructor);
  // prints true; should be false
console.log(new Headers().values()[Symbol.toStringTag]);
  // prints Generator; should be "Headers Iterator"
  // note that Chrome implements this wrongly but Firefox and Safari both do the right thing
  // see https://crbug.com/793204
console.log(new Headers().sort);
  // prints [Function: sort]; should be undefined
console.log(Headers.prototype.get.call(1, 'content-type'));
  // prints undefined; should throw a TypeError
```

The two biggest readability improvements #834 brings are a) the simplification using URLSearchParams and b) generator functions. However, the first is a stretch – there isn't really anything similar between Headers and URLSearchParams other than the multimap interface they expose – and the second is simply against the [Web IDL](https://heycam.github.io/webidl/#define-the-iteration-methods) standard that defines how the JavaScript interface of web APIs. On the other hand, #834 starts using ES2015 `Proxy`, a huge hammer with lots of nuances for something that's really quite simple. (Note that I am also a current editor of the Web IDL spec, and authored much of the Node.js URLSearchParams implementation.)

It appears to me that a cost–benefit analysis would weigh against the recent rewrite of the `Headers` class. I ask the current maintainers to consider reverting the new implementation.

jimmywarting:
## Purpose
Replace the header class for something that follows the spec a little more correctly

## Changes
`getAll()` & `raw()` and all of the URLSearchParams methods are removed

## Additional information
https://github.com/node-fetch/node-fetch/issues/1119#issuecomment-1002220499
___

- [x] I updated & removed some unit test(s)

___

- fix #1119


gr2m (commented):
> `getAll()` & `raw()` and all of the URLSearchParams methods are removed

Make sure to add `BREAKING CHANGE:` statements in the commit body when you merge. You can add multiple, e.g.

```
refactor: replace headers class for better compatibility with the spec

BREAKING CHANGE: `Headers.getAll()` has been removed

BREAKING CHANGE: `Headers.raw()` has been removed
```

Macias131987:
jimmywarting:replace-headers-class

PR #1485, opened 2022-01-27
chore: completely remove `.buffer()`

jimmywarting:
## Purpose
time to remove deprecated node-fetch only method `.buffer()`

## Additional information

___

- [ ] I updated readme
- [x] I added & removed some unit test(s)


gr2m (approved):

PR #1486, opened 2022-01-27
chore: completely remove support for form-data package

### LINKED ISSUE #1167: Drop support for Form-Data?

jimmywarting:
## Purpose
Pull the plug on deprecated form-data package

___

- [ ] I updated readme
- [x] I removed some unit test(s)

___

- fix #1167

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "chai-iterator": "^3.0.2",
     "chai-string": "^1.5.0",
     "coveralls": "^3.1.0",
-    "form-data": "^4.0.0",
     "formdata-node": "^4.2.4",
     "mocha": "^9.1.3",
     "p-timeout": "^5.0.0",
--- a/src/body.js
+++ b/src/body.js
@@ -267,9 +267,7 @@ export const clone = (instance, highWaterMark) => {
 		throw new Error('cannot clone body after it is used');
 	}
 
-	// Check that body is a stream and not form-data object
-	// note: we can't clone the form-data object without having it as a dependency
-	if ((body instanceof Stream) && (typeof body.getBoundary !== 'function')) {
+	if (body instanceof Stream) {
 		// Tee instance body
 		p1 = new PassThrough({highWaterMark});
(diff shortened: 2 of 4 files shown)


gr2m (approved):

PR #1490, opened 2022-01-27
chore: remove EventTarget from the isAbortSignal

### LINKED ISSUE #1469: Don't check if AbortSignals tag name match EventTarget
For the v4 milestone: Revert #1001 about checking that if the tag name match EventTarget. 

_Originally posted by @TimothyGu in https://github.com/node-fetch/node-fetch/pull/1001#discussion_r550736382_

jimmywarting:
## Purpose
should not check if a AbortSignal is of type EventTarget

## Additional information

___

- [ ] I updated readme
- [x] I changed unit test description

___

- fix #1469

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -49,12 +49,7 @@ export const isBlob = object => {
  * @return {boolean}
  */
 export const isAbortSignal = object => {
-	return (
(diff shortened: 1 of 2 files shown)


gr2m (approved):

PR #1508, opened 2022-02-19
fix: possible `Cannot read property 'startsWith' of null`

UltraCakeBakery:
## Purpose
If the header is not set this function throws the error in the title of this pull request.

> I ran into this issue in `svelte-kit`: https://discord.com/channels/457912077277855764/939868205869072444/944634140811218975

## Changes
Added a question mark (optional chaining) to the if statement

## Additional information
Not sure if this is the correct fix for this problem. It 

___

- [ ] I updated readme
- [ ] I added unit test(s)

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -110,7 +110,7 @@ export default class Body {
 	async formData() {
 		const ct = this.headers.get('content-type');
 
-		if (ct.startsWith('application/x-www-form-urlencoded')) {
+		if (ct?.startsWith('application/x-www-form-urlencoded')) {
 			const formData = new FormData();
 			const parameters = new URLSearchParams(await this.text());
 


UltraCakeBakery:
Not a fan of my fix here to be honest. According to https://developer.mozilla.org/en-US/docs/Web/API/Request/formData this should always resolve / return an FormData object... Should I change the entire formData() function to always return a formData?

7nik:
@UltraCakeBakery MDN is a nice place to read what does a function/class/etc. but you need to look to the specs to understand how it does it. [The specs ](https://fetch.spec.whatwg.org/#ref-for-dom-body-formdata%E2%91%A0) says following for the method `formData()`:
> FormData
> If mimeType’s essence is "multipart/form-data", then:
> ...
> Otherwise, if mimeType’s essence is "application/x-www-form-urlencoded", then:
> ...
> Otherwise, throw a TypeError.

So, for unsupported content types, it should throw a TypeError.

The node-fetch's `formData()` itself handles the second case, two others - by a parser called via `toFormData(this.body, ct)`.

P.S.: I wouldn't say that parser's ` throw new TypeError('Failed to fetch');` for unsupported content type is an informative error.

UltraCakeBakery:
> @UltraCakeBakery MDN is a nice place to read what does a function/class/etc. but you need to look to the specs to understand how it does it. [The specs ](https://fetch.spec.whatwg.org/#ref-for-dom-body-formdata%E2%91%A0) says following for the method `formData()`:
> 
> > FormData
> > If mimeType’s essence is "multipart/form-data", then:
> > ...
> > Otherwise, if mimeType’s essence is "application/x-www-form-urlencoded", then:
> > ...
> > Otherwise, throw a TypeError.
> 
> So, for unsupported content types, it should throw a TypeError.
> 
> The node-fetch's `formData()` itself handles the second case, two others - by a parser called via `toFormData(this.body, ct)`.
> 
> P.S.: I wouldn't say that parser's ` throw new TypeError('Failed to fetch');` for unsupported content type is an informative error.

Gotcha. I didn't know where to find the specification so I assumed MDN was a good second choice. It would be kinda neat imo if formData is always available but empty instead of it throwing an error. But that is ofc not the spec so yeah.

Is the change in my pull request valid? If so, I would like to see it merged ASAP as it is a blocking issue for me and other users of svelte-kit.

jimmywarting:
@7nik is right, it should throw an error one way or the other. 
basically using `body.formdata()` requires/demand that that there is a `content-type` that it should be able to read from in order to properly deserialize it back into a FormData

it should rather be a "Failed to fetch" type error if anything.

---

now i notice that we are not canceling/aborting the response if formdata fails for some reason (which we should be doing)

jimmywarting:
beside, we currently use node v12.20 and that have no support for optional changing
we might drop support for 12.x in april when it goes EOL

PR #1513, opened 2022-02-25
fix: cancel request example import

SoftwareAndOutsourcing:
Fixes cancel request example import.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,7 +470,7 @@ You may cancel requests with `AbortController`. A suggested implementation is [`
 An example of timing out a request after 150ms could be achieved as the following:
 
 ```js
-import fetch, { AbortError} from 'node-fetch';
+import fetch, { AbortError } from 'node-fetch';
 
 // AbortController was added in node v14.17.0 globally
 const AbortController = globalThis.AbortController || await import('abort-controller')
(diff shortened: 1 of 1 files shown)


jimmywarting:
Hmm, wondering if not we should try to fix the real problem that is that we return a custom AbortError message, it should rather be a `DOMException` 

```js
// Browser test:
ctrl = new AbortController() 
ctrl.abort()
fetch('/', ctrl).catch(console.log) 
// DOMException { 
//   name: AbortError,
//   code: 20,
//   message: "Failed to execute 'fetch' on 'Window': The user aborted a request." 
// }
```

We could optionally start to depend on [node-domexception](https://www.npmjs.com/package/node-domexception). it's already included by another sub dependency of ours (fetch-blob).

README.md
```diff
@@ -470,7 +470,7 @@ You may cancel requests with `AbortController`. A suggested implementation is [`
 An example of timing out a request after 150ms could be achieved as the following:
 
 ```js
-import fetch from 'node-fetch';
+import fetch, { AbortError} from 'node-fetch';
```
LinusU, line 473:
```suggestion
import fetch, { AbortError } from 'node-fetch';
```

LinusU (approved):
LGTM 👍

LinusU:
@jimmywarting I think that would be nice, but I still think that we should merge this one in the mean time 👍

jimmywarting:
Alright. maybe change it in next major release to avoid breakage?

jimmywarting:
should maybe have done `docs:` instead of `fix:` to avoid bumping an unnecessary release

PR #1521, opened 2022-03-13
fix: Add `ArrayBuffer` and views to the allowed `BodyInit` types

cprecioso:
## Purpose

The previous type definitions produced an error when assigning an `ArrayBuffer` or an `ArrayBufferView` to the `Request` body. However, [these cases are actually handled in the code](https://github.com/node-fetch/node-fetch/blob/6425e2021a7def096e13dbabcac2f10e6da83d11/src/body.js#L47-L52), so they should not produce an error.

## Changes

Add `ArrayBufferLike` and `ArrayBufferView` to the `BodyInit` union.

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -121,6 +121,8 @@ export interface ResponseInit {
 export type BodyInit =
 	| Blob
 	| Buffer
+	| ArrayBufferLike
+	| ArrayBufferView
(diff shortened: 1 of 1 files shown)


jimmywarting (approved):

LinusU (approved):

cprecioso:
Closing due to old age

PR #1535, opened 2022-04-14
docs: Fix min nodejs version needed for AbortController

pietrzakacper:
After testing that and checking the docs it seems that AbortController was introduced in v15.0.0 not in v14.17.0
https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,7 +472,7 @@ An example of timing out a request after 150ms could be achieved as the followin
 ```js
 import fetch, { AbortError } from 'node-fetch';
 
-// AbortController was added in node v14.17.0 globally
+// AbortController was added in node v15.0.0 globally
 const AbortController = globalThis.AbortController || await import('abort-controller')
 
 const controller = new AbortController();


jimmywarting:
From theirs docs it says

<summary style="box-sizing: border-box; margin: 0.5rem 0px; padding: 0.5rem 0px; cursor: pointer;">History</summary><slot id="details-content" style="color: rgb(255, 255, 255); font-family: Lato, &quot;Lucida Grande&quot;, &quot;Lucida Sans Unicode&quot;, &quot;Lucida Sans&quot;, Verdana, Tahoma, sans-serif; font-size: 13.6px; font-style: normal; font-variant-ligatures: none; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(9, 12, 21); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">

Version | Changes
-- | --
v15.4.0 | No longer experimental.
v15.0.0, v14.17.0 | Added in: v15.0.0, v14.17.0

</slot>

but that did require some experimental flags to be enabled...

pietrzakacper:
I see, thanks @jimmywarting for giving more context :) In my tests, using 14.17.0 the `AbortController` was not globally available when running node without any flags. In 15.0.0 it is globally available but shows warnings:
```
(node:273417) ExperimentalWarning: AbortController is an experimental feature. This feature could change at any time
```
From the user perspective, I expected my code that runs on 14.17.0 to be able to just use this feature without any additional hassle. I think it should be either communicated that it requires additional flags on runtime or the min supported version should be the one that doesn't require any runtime flags, which is 15.0.0

alexanderramos89:
test/test.js

PR #1557, opened 2022-05-19
fix: destroy socket if body is aborted

MoonBall:
## Purpose

Fix: the socket is hangup although the body is aborted.

## Changes

Destroy socket if body is aborted.

## Additional information

see [demo](https://github.com/MoonBall/node-fetch-hangup-demo) for reproduction.

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix #000


MisterD123:
I just encountered an equivalent issue in a test in our codebase and my fix looks almost identical, except I have the destroy error forwarding go in the opposite direction. Should the introduced `pipeAndDestroy` method do both `dst.on('error', err => src.destroy(err));` and `src.on('error', err => dst.destroy(err));` ?

PR #1599, opened 2022-07-07
fix(Headers): don't forward secure headers on protocol change

max-stytch:
## Purpose
Resolves https://www.huntr.dev/bounties/db31e05b-ff10-4057-81a3-37445bf161cd/ by validating that the URL protocol remains the same when determining whether to send secure headers on a redirect.
This prevents MITM attacks from sniffing secure headers when a redirect downgrades a `https://` to a `http://`

## Changes
Adds an additional check to the redirect `follow` step to determine whether to send secure headers or not.

## Additional information

___

- [ ] I updated readme
- [x] I added unit test(s)

___

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,7 +22,7 @@ import {FetchError} from './errors/fetch-error.js';
 import {AbortError} from './errors/abort-error.js';
 import {isRedirect} from './utils/is-redirect.js';
 import {FormData} from 'formdata-polyfill/esm.min.js';
-import {isDomainOrSubdomain} from './utils/is.js';
+import {isDomainOrSubdomain, isSameProtocol} from './utils/is.js';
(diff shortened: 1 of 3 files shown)


jimmywarting (approved):

gr2m (approved):
great PR 👍🏼

victal:
Sorry if this is the wrong place to ask, but since this PR is a fix for a security issue, will (or should) it be backported to the 2.x branch as it was done for #1449?

Is this done automatically or should I (or someone else interested in the fix) open another PR targeting the 2.x branch for that?

jimmywarting:
if you @victal could create a PR to the `v2` branch then that would be grate!

victal:
Just created #1605 for it, thanks!

PR #1605, opened 2022-07-19
fix(headers): don't forward secure headers on protocol change

victal:
## Purpose

Backport of #1599 to the 2.x branch as it's a security fix (Resolves the issue at https://www.huntr.dev/bounties/db31e05b-ff10-4057-81a3-37445bf161cd/)

## Additional information

___

- [ ] I updated readme
- [x] I added unit test(s)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,6 +36,20 @@ const isDomainOrSubdomain = (destination, original) => {
 	);
 };
 
+/**
+ * isSameProtocol reports whether the two provided URLs use the same protocol.
+ *
+ * Both domains must already be in canonical form.
+ * @param {string|URL} original
+ * @param {string|URL} destination
+ */
+const isSameProtocol = (destination, original) => {
+       const orig = new URL(original).protocol;
+       const dest = new URL(destination).protocol;
+
+       return orig === dest;
+};
+
 
 /**
  * Fetch function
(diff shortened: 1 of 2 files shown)


jimmywarting (approved):

victal:
I forgot to ask earlier, but are 2.x versions released automatically when a PR is closed, the same as new 3.x versions?

If not, how/when can we have a new release published with this fix, and can we help with that in any way if needed?

sseide:
apparently the release is not done automatically as there is no v2.6.8 by now while these changes are merged for some days and the newer 3.x release with this fix is available for some days too.

@jimmywarting @cvalb Can someone please publish a new 2.x release to fix this security issue for all projects depending on this older version?

Many thanks in Advance,
S. Seide

jimmywarting:
It should make an release automatically but i think it failed to publish at that one point for some reason.
@gr2m what would be the best way to re-trigger a release with semantic-release

dgmike:
Hello @jimmywarting @gr2m, a new version will rolled up to 2.x.x on npm?

sseide:
@gr2m or @jimmywarting - any news about a new 2.x npm release? It did not happen until now and looking at the npm download statistic in comparison to the 3.x release its much needed...

Thanks

spacenate:
Thanks @victal for back-porting this! @jimmywarting @gr2m it would be awesome if a new 2.x version could be released on npm 🙏

maralfol:
Hello,

Any update on 2.x release @jimmywarting @gr2m? I've been checking the releases and I don't think it was even triggered, or at least, it does not appear on the release history. It would be really helpful if anyone could trigger a new release 🙏🙏.
[Apollo-core](https://github.com/apollographql/apollo-server/blob/main/packages/server/package.json) is using 2.6.7 in all versions and triggering it would fix it.
Thanks!

iamart:
This seemed to be a breaking change for us; after dep. update a server w/ 302 protocol redirect response makes fetch to return 401 error instead of following up with redirect and returning 200.

Issue #1609, opened 2022-07-26
Found potential security but got no response


vovikhangcdv:
## Issue

Hi there,

I have found a potential security issue in [node-fetch](https://github.com/node-fetch/node-fetch). I had tried to connect to SECURITY.md email but got no response too. I don't create a full disclosure issue there cause it could affect the users. So please get in touch with me through email [doublevkay@gmail.com](mailto:doublevkay@gmail.com) or just access the report on https://huntr.dev/bounties/a7e6a136-0a4b-46c4-ad20-802f1dd60bf7 (only private accessible by maintainers).

vovikhangcdv:
Hi @NotMoni and @jimmywarting, can you verify [the report on Huntr](https://huntr.dev/bounties/a7e6a136-0a4b-46c4-ad20-802f1dd60bf7)?

titivuk:
Hello,
What about version 2.x?

vovikhangcdv:
> Hello, What about version 2.x?

Hi, Version 2.x is not affected by this vulnerability.

titivuk:
> > Hello, What about version 2.x?
> 
> Hi, Version 2.x is not affected by this vulnerability.

Hi, looks like Snyk issue, I did not code causing the issue in v2 but Snyk says v2 is affected as well.
Thank you for confirmation

PR #1611, opened 2022-07-26
fix: ReDoS referrer

vovikhangcdv:
## Purpose
Fix Inefficient Regular Expression Complexity (ReDoS) in the referrer regex.
> Inefficient regular expression complexity regex when trying to match Potentially Trustworthy could lead to a denial of service attack. With a formed payload 'http://' + 'a.a.'.repeat(i) + 'a', 76 characters payload could take 42642 ms time execution. 

## Changes
- Change `Potentially Trustworthy` regex matcher.

## Additional information
- [Report on Huntr](https://huntr.dev/bounties/a7e6a136-0a4b-46c4-ad20-802f1dd60bf7/)

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix [#1609](https://github.com/node-fetch/node-fetch/issues/1609)

### DIFF
--- a/src/utils/referrer.js
+++ b/src/utils/referrer.js
@@ -119,7 +119,7 @@ export function isOriginPotentiallyTrustworthy(url) {
 	// 5. If origin's host component is "localhost" or falls within ".localhost", and the user agent conforms to the name resolution rules in [let-localhost-be-localhost], return "Potentially Trustworthy".
 	// We are returning FALSE here because we cannot ensure conformance to
 	// let-localhost-be-loalhost (https://tools.ietf.org/html/draft-west-let-localhost-be-localhost)
-	if (/^(.+)\.localhost$/.test(url.host)) {
+	if (url.host === 'localhost' || url.host.endsWith('.localhost')) {
 		return false;
 	}
 
--- a/src/utils/referrer.js
+++ b/src/utils/referrer.js
@@ -119,7 +119,7 @@ export function isOriginPotentiallyTrustworthy(url) {
 	// 5. If origin's host component is "localhost" or falls within ".localhost", and the user agent conforms to the name resolution rules in [let-localhost-be-localhost], return "Potentially Trustworthy".
 	// We are returning FALSE here because we cannot ensure conformance to
 	// let-localhost-be-loalhost (https://tools.ietf.org/html/draft-west-let-localhost-be-localhost)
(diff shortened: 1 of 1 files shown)


jimmywarting (dismissed):
LGTM

src/utils/referrer.js
```diff
@@ -119,7 +119,7 @@ export function isOriginPotentiallyTrustworthy(url) {
 	// 5. If origin's host component is "localhost" or falls within ".localhost", and the user agent conforms to the name resolution rules in [let-localhost-be-localhost], return "Potentially Trustworthy".
 	// We are returning FALSE here because we cannot ensure conformance to
 	// let-localhost-be-loalhost (https://tools.ietf.org/html/draft-west-let-localhost-be-localhost)
-	if (/^(.+\.)*localhost$/.test(url.host)) {
+	if (/^(.+)\.localhost$/.test(url.host)) {
```
LinusU, line 122: I don't think that this will match just `localhost` anymore?

How about:

```suggestion
	if (url.host === 'localhost' || url.host.endsWith('.localhost')) {
```

jimmywarting: you are right... `/^(.+)\.localhost$/.test('localhost') === false`

jimmywarting (approved):

FlameFractal:
Can we add this to v2 (CommonJS) as well?

([ref](https://github.com/BindrUK/bindr/pull/2056))

LinusU:
> Can we add this to v2 (CommonJS) as well?

We would be happy to accept a PR for this

FlameFractal:
I went through 2.x code and I could not find this issue there. 2.x doesn't even implement `referrer`, and there's no check for `localhost` anywhere else either. Moreover, there's only 10 regex tests and I think none contain this vulnerability (maybe @vovikhangcdv could verify this).

We got a Snyk-bot notification about this issue, but I think 2.x is not affected by this CVE.

vovikhangcdv:
Hi there, I confirm that version 2.x is not affected by this vulnerability.

vovikhangcdv:
I suggest we could add a Security Advisory and specify the affected version for this issue.

JamieSlome:
Just for clarity, we have updated the CVE here:

https://github.com/CVEProject/cvelist/pull/6757/files

PR #1625, opened 2022-08-15
Remove unnecessary await import() call

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

## Purpose
I have a need to use webpack to bundle node-fetch targeting "electron-renderer" and it would be very useful to only have a single file as output. However, because of this `await import()` call, webpack generates two files.

## Changes
- Change an `await import()` to `import { ... } from "..."`

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -15,6 +15,7 @@ import {FormData, formDataToBlob} from 'formdata-polyfill/esm.min.js';
 import {FetchError} from './errors/fetch-error.js';
 import {FetchBaseError} from './errors/base.js';
 import {isBlob, isURLSearchParameters} from './utils/is.js';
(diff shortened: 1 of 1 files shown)


jimmywarting:
I purposely did a lazy async import b/c formdata parser isn't always needed. it's a bit wasteful to load files that aren't needed. beside this is server side... so there shouldn't be any big reason to have to bundle everything into one single file?

even less so now that node v18 have fetch built in right now.

Nokel81:
Sure, but if you read the purpose of this change then you would see that I am not quite using a node server.

Furthermore, one small file isn't saving that much memory.

Nokel81:
@jimmywarting I was able to work around this, so I will close this PR.

Issue #1569, opened 2022-05-31
Keep headers order while sending them


AlttiRi:
### Bug
It sorts user set headers.

### Expected behavior

It should **not** sort the headers while sending them (while performing a network request).

### Example

```js
import nodeFetch from "node-fetch";

const url = "http://127.0.0.1:3000/";
let headers = {
    "xxx-3": "3",
    "xxx-2": "2",
    "xxx-1": "1",
}
console.log(Object.keys(headers));
console.log(await (await nodeFetch(url, {headers})).text());
```
It sends _(You can check it with a HTTP server that logs the input request headers)_
```
{
  accept: '*/*',
  'accept-encoding': 'gzip,deflate,br',
  connection: 'close',
  'user-agent': 'node-fetch',
  'xxx-1': '1',
  'xxx-2': '2',
  'xxx-3': '3',
  host: '127.0.0.1:3000'
}
```

My `xxx-*` headers were sorted. 
The default header were not sorted. (They keep the append order.)

### Additional comment

If you want to simulate `Headers` object behavior you should sort the headers only when a user iterates through them, but not when you send the request.

https://github.com/nodejs/undici/pull/1309#issuecomment-1082040985

More over, it does not sort the default headers. So this sorting just makes no sense.

### Info

- node-fetch v3.2.3

AlttiRi:
Also it think it should send `"accept-encoding": "gzip, deflate, br"` instead of `"accept-encoding": "gzip,deflate,br"` like it browsers do.

They separate encoding with `", "`, not just with `","`.

AlttiRi:
Yes, when a user code iterates a `Header` object the iterator returns the headers in sorted order.
But the inner state keeps the order in which the headers were added. In this order the headers are send in TCP socket.

Keeping the headers order is required when a user want to make exactly the same requests as it do other libs, browsers.

PR #1662, opened 2022-10-17
Remove header sorting when iterates over it

JVBorges:
## Purpose

Keep headers order while sending them.

## Changes

Remove sorting function when iterates over it.
___

- [x] I updated readme
- [x] I added unit test(s)

___

- fix [1569](https://github.com/node-fetch/node-fetch/issues/1569)

### DIFF
--- a/src/headers.js
+++ b/src/headers.js
@@ -105,7 +105,7 @@ export default class Headers extends URLSearchParams {
 
 		super(result);
 
-		// Returning a Proxy that will lowercase key names, validate parameters and sort keys
+		// Returning a Proxy that will lowercase key names, validate parameters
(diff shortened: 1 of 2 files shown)


paulodiovani:
This is a small _fix_ that does not change the behavior of the library, but since sorting the headers does not help with anything, it is a welcome change anyway, IMO.

Note: the `Host` header is included by node itself (https://github.com/nodejs/node/blob/6adaf23c20acc48b836134ea660dc8c16f028117/lib/_http_client.js#L283-L300) (pointed by @alexandresaura).

Somberor (approved):
Please accept this merge very needed.

PR #1670, opened 2022-11-08
feat: add static Response.json

### LINKED ISSUE #1669: Impl "Response.json"
**Reproduction**

1. `new Response.json({ foo: 'bar' })`

KhafraDev:
## Purpose

Implements static `Response.json` from https://github.com/whatwg/fetch/pull/1392

## Changes

Adds static Response.json

## Additional information

___

- [ ] I updated readme
- [x] I added unit test(s)

___

- Fixes #1669

### DIFF
--- a/src/response.js
+++ b/src/response.js
@@ -124,6 +124,25 @@ export default class Response extends Body {
 		return response;
 	}
 
+	static json(data = undefined, init = {}) {
+		const body = JSON.stringify(data);
+
+		if (body === undefined) {
+			throw new TypeError('data is not JSON serializable');
+		}
+
+		return new Response(body, {
+			...init,
+			headers: {
+				// If body’s type is non-null and response’s header list does not contain
+				// `Content-Type`, then append (`Content-Type`, body’s type) to response’s
+				// header list.
+				'Content-Type': 'application/json',
+				...init.headers
+			}
+		});
+	}
+
 	get [Symbol.toStringTag]() {
 		return 'Response';
 	}
--- a/src/response.js
+++ b/src/response.js
@@ -131,15 +131,15 @@ export default class Response extends Body {
 			throw new TypeError('data is not JSON serializable');
 		}
 
+		const headers = new Headers(init && init.headers);
+
+		if (!headers.has('content-type')) {
+			headers.set('content-type', 'application/json');
+		}
+
 		return new Response(body, {
 			...init,
-			headers: {
-				// If body’s type is non-null and response’s header list does not contain
-				// `Content-Type`, then append (`Content-Type`, body’s type) to response’s
-				// header list.
-				'Content-Type': 'application/json',
-				...init.headers
-			}
+			headers
 		});
 	}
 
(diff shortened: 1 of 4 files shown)


src/response.js
```diff
@@ -124,6 +124,25 @@ export default class Response extends Body {
 		return response;
 	}
 
+	static json(data = undefined, init = {}) {
```
LinusU, line 127: I think that using `init = {}` and assuming `init` is an object further down in the function will make the following code crash:

```js
Response.json({ hej: 1 }, null)
```

But in Chrome that works:

Screenshot 2022-11-09 at 09 22 54

src/response.js
```diff
@@ -124,6 +124,25 @@ export default class Response extends Body {
...
+		}
+
+		return new Response(body, {
+			...init,
+			headers: {
+				// If body’s type is non-null and response’s header list does not contain
+				// `Content-Type`, then append (`Content-Type`, body’s type) to response’s
+				// header list.
+				'Content-Type': 'application/json',
+				...init.headers
```
LinusU, line 141: What happens if `init.headers` contains `content-type` or `cOnTeNt-TyPe`?

src/response.js
LinusU, line 127: Not sure if this is in the spec, but Chrome throws different errors if called without `data` vs. explicitly passing `undefined`:

Screenshot 2022-11-09 at 09 29 56

Not sure we need to follow this though, but I think that the `= undefined` should be removed? It communicates that the argument is optional, but it is in fact required...

```suggestion
	static json(data, init = {}) {
```

KhafraDev: `any` webidl types mean that the type is optional and Response.json.length should be 0

src/response.js
KhafraDev: I have a test for this

src/response.js
KhafraDev: node-fetch doesn't seem to do much argument validation, I followed other areas of the code.

For example when constructing a new response:
```js
new Response('', null)
```

throws an error when it shouldn't

```js
new Response('', 3)
```
works when it shouldn't (same with booleans, etc)

KhafraDev: see https://github.com/nodejs/undici/pull/1723

jimmywarting:
I'm okey with some small deviations, we don't have all the webidl stuff undici have and don't follow the spec at 100%. it was set up to be "good enough" and solve the most common use cases.
We didn't have all the web parts in NodeJS that builds up fetch for what it is today in the beginning when node-fetch was first built, like web streams, blob, formdata, AbortController, etc. so we did compromise a little bit. 

it's going to be replaced by NodeJS in the newer versions anyway at some point. node-fetch is just a polyfill at this stage. 
it at least gets the job done and i'm sure ppl will get it right 99% of the times.

src/response.js
LinusU: That's cool, today I learned!

LinusU (approved):
Looks great! 👍

jimmywarting (approved):

PR #1686, opened 2022-12-02
Response.bodyUsed should be false when there is no body

### LINKED ISSUE #1684: Response.bodyUsed should be false when there is no body
**Reproduction**

```TypeScript
const response = new Response(); // <=== Response does not have a body
expect(response.bodyUsed).toEqual(false); // OK

expect(await response.text()).toEqual('');
expect(response.bodyUsed).toEqual(false); // KO: node-fetch returns true
```

**Expected behavior**

Under Chrome 107:

```
<< const response = new Response()
<< response.bodyUsed
>> false
<< await response.text()
<< response.bodyUsed
>> false
```

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.3.0
| node             | 18.12.1
| npm              | 8.19.2
| Operating System | macOS Monterey

Related issue: https://github.com/nodejs/undici/issues/1794

tkrotoff:
Fix #1684


test/response.js
```diff
@@ -264,4 +264,26 @@ describe('Response', () => {
 
 		new Response('a').data;
 	}));
+
+	it('should set bodyUsed to true if body has been read', async () => {
+		for (const toCheck of ['a', '']) {
+			const res = new Response(toCheck);
+			expect(res.bodyUsed).to.equal(false);
+
+			/* eslint-disable-next-line no-await-in-loop */
```
tkrotoff, line 273: Don't think this lint rule matters in context of unit test

src/body.js
```diff
@@ -209,6 +207,8 @@ async function consumeBody(data) {
 		return Buffer.alloc(0);
 	}
 
+	data[INTERNALS].disturbed = true;
+
 	/* c8 ignore next 3 */
 	if (!(body instanceof Stream)) {
```
tkrotoff, line 213: Don't know if `disturbed = true` should be done after this line

PR #1687, opened 2022-12-05
[#1576] Fixes unexpected premature close on chunked responses

### LINKED ISSUE #1576: Getting `Premature close` errors unexpectedly but very reliably in specific circumstances
Hi,

Thanks for this great library!

We have been getting `Premature close` errors on our production server when calling API endpoints.

These errors are entirely repeatable for the same API request with a given input and output – for example looking up details for a specific user – while API requests to the same endpoint with different inputs/outputs work as expected.

The errors occur only on our production Linux server (of course) and are not reproducible on our development machines (macOS), at least not for the specific users that trigger the error on the server.

I have not dug deeply enough to be sure, but it smells to me like faulty handling of chunked responses that happen to trigger a boundary condition due to the packet size or chunk read size on our server.

I am suspicious of the code here https://github.com/node-fetch/node-fetch/blob/043a5fcfc82884155bcb2afba2a82333b2f29c02/src/index.js#L391-L399 for recognising the terminating bytes of a chunked response. It seems to assume that the final 5 bytes will always be either contained in a single chunk, or be at a chunk boundary exactly 3 bytes into the sequence. Should/could this check instead look at the last 5 bytes received in all cases, no matter which chunk contains those bytes?

The same API requests that trigger `Premature close` errors work on the production server using node-fetch succeed if we use the Axios HTTP client library instead, and if we make the requests on the command line with `curl`.

**Reproduction**

1. Run on a Linux server (Ubuntu 20.04 LTS)
2. Make GET request to an API endpoint that works in 99% of cases
3. Get a response that happens to trigger the issue, e.g. in our case with a Content-Length of 9171 or 12816
4. Request fails every time with the `Premature close` error

**Expected behavior**

Complete the request and return the response body.

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | Confirmed issue with versions 3.2.1 and 3.2.5
| node             | 16.14.0 and 16.15.1
| npm              | 8.3.1 and 8.11.0
| Operating System | Ubuntu 20.04.4 LTS

**Additional context**

Requests to the same API endpoints and data request/responses work correctly on dev machines (macOS) and with alternative HTTP clients (Axios, curl)

Treverix:
## Purpose
Fixes #1576 (unexpected premature close errors on chunked responses)

## Changes

It corrects a false assumption about split of the 'LAST CHUNK' marker between the previous and current buffer. The new solution concatenates last bytes of the previous buffer with all bytes from the current buffer (if that holds less then 5 bytes) and does the check on that.

## Additional information

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix #1576

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -397,10 +397,10 @@ function fixResponseChunkedTransferBadEnding(request, errorCallback) {
 
 			// Sometimes final 0-length chunk and end of message code are in separate packets
 			if (!properLastChunkReceived && previousChunk) {
-				properLastChunkReceived = (
-					Buffer.compare(previousChunk.slice(-3), LAST_CHUNK.slice(0, 3)) === 0 &&
-					Buffer.compare(buf.slice(-2), LAST_CHUNK.slice(3)) === 0
-				);
+				if (buf.length < 5) {
+					properLastChunkReceived = Buffer.compare(
+						Buffer.from([...previousChunk.slice(-5), ...buf]).slice(-5), LAST_CHUNK) === 0;
+				}
 			}
 
 			previousChunk = buf;


kasperpeulen:
I would love to test out this PR, we suffer quite a lot from a premature close issue in node-fetch in the storybook pipeline:

* https://app.circleci.com/pipelines/github/storybookjs/storybook/33011/workflows/b3ef4e9b-4a44-46e6-8c9a-d2075054d683/jobs/444562/parallel-runs/5
* https://app.circleci.com/pipelines/github/storybookjs/storybook/32962/workflows/e512522a-96ed-4a4d-aef9-e1770dc08076/jobs/444380/parallel-runs/2
* https://app.circleci.com/pipelines/github/storybookjs/storybook/32892/workflows/67f913a0-332d-4382-8966-c902c2c3c4b1/jobs/444017/parallel-runs/2
* https://app.circleci.com/pipelines/github/storybookjs/storybook/32690/workflows/a2cfb5d9-79ce-488d-af9a-e282a96310a7/jobs/443167/parallel-runs/2
* https://app.circleci.com/pipelines/github/storybookjs/storybook/35150/workflows/a3f13eb6-bf9b-4e3b-9d9e-d6594277eaac/jobs/454236/parallel-runs/5

Not sure if it is related, but we have no clue how to fix it.

xfournet:
This PR should be merged. We suffer the same problem, and this patch fixed it. It must be released in npm (3.x.x and 2.x.x), thanks!

marcesengel:
@LinusU would you maybe be so kind to assign other reviewers in case you don't have the time right now?

TatyanaA-PH:
+1

newhouse:
bump...kinda  a big deal, this problem

Issue #1657, opened 2022-09-30
'FormData' refers to a value, but is being used as a type here


IlyaSavichItr:
**Reproduction**

Hi, I've got some issues with node-fetch v3.2.10 while trying to upgrade from v2.6.7 (tbf these issues appeared starting from v3.0.0).

When trying to compile typescript getting
```
node_modules/formdata-polyfill/esm.min.d.ts:2:11 - error TS2749: 'FormData' refers to a value, but is being used as a type here. Did you mean 'typeof FormData'?

2   new (): FormData;
            ~~~~~~~~
```
Or
```
node_modules/node-fetch/@types/index.d.ts:137:22 - error TS2749: 'FormData' refers to a value, but is being used as a type here. Did you mean 'typeof FormData'?

137  formData(): Promise<FormData>;
                         ~~~~~~~~
```

**Expected behavior**

I expect not to get any issue

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.2.10
| node             | 18.1.0
| yarn              | 3.2.0
| Operating System | WSL2.0 Ubuntu 20.04, Windows 10

Tried typescript v.4.7.4 and v4.8.4, both gets the same result

ritschwumm:
i'm not quite sure why this doesn't work any more. i'd really like to know, so i'd be glad if someone could help me find out.
to me it seems as if the real culprit is formdata-polyfill which does not provide any types for FormData except the declaration that it exists and can be constructed.

i noticed that adding the "dom" lib to tsconfig.json makes the compile error go away. as a workaround this might be good enough, but i'd very much prefer not pretending my node app has a full browser dom available...

qtiki:
> i noticed that adding the "dom" lib to tsconfig.json makes the compile error go away. as a workaround this might be good enough, but i'd very much prefer not pretending my node app has a full browser dom available...

Trying to use `@types/node` and `@types/web` at the same time often leads to conflicts in the types. I tried with a clean project and just adding `node-fetch` with both types I get a compiler error from `AbortSignal` type mismatch:
```
node_modules/@types/web/index.d.ts:2048:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; timeout(milliseconds: number): AbortSignal; }'.

2048 declare var AbortSignal: {
[...]

PR #1696, opened 2023-01-05
Allow URL class object as an argument for fetch()

### LINKED ISSUE #1261: URL is missing in typescript definitions for RequestInfo
node-fetch: 3.0.0

Note, jsdoc for [fetch](https://github.com/node-fetch/node-fetch/blob/main/src/index.js#L30) lists URL as supported input.

Maxim-Mazurok:
## Purpose

Fixes https://github.com/node-fetch/node-fetch/issues/1261

## Changes

Added `URL` to `fetch` and `Request` constructor

## Additional information

Inspired by: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a9f480c82b3b356990e5dd574bf2c1ba9c2fdfb5/types/node-fetch/index.d.ts#L213 and https://github.com/microsoft/TypeScript/blob/f43cd0accac3e5033820fd27035731ed88f54938/lib/lib.dom.d.ts#L18281
___

- [x] I added unit test

___

- fix #1261

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -145,9 +145,9 @@ export interface Body extends Pick<BodyMixin, keyof BodyMixin> {}
 
 export type RequestRedirect = 'error' | 'follow' | 'manual';
 export type ReferrerPolicy = '' | 'no-referrer' | 'no-referrer-when-downgrade' | 'same-origin' | 'origin' | 'strict-origin' | 'origin-when-cross-origin' | 'strict-origin-when-cross-origin' | 'unsafe-url';
-export type RequestInfo = string | URL | Request;
+export type RequestInfo = string | Request;
 export class Request extends BodyMixin {
-	constructor(input: RequestInfo, init?: RequestInit);
+	constructor(input: RequestInfo | URL, init?: RequestInit);
 
 	/**
 	 * Returns a Headers object consisting of the headers associated with request. Note that headers added in the network layer by the user agent will not be accounted for in this object, e.g., the "Host" header.
@@ -216,4 +216,4 @@ export class AbortError extends Error {
 }
 
 export function isRedirect(code: number): boolean;
-export default function fetch(url: RequestInfo, init?: RequestInit): Promise<Response>;
+export default function fetch(url: RequestInfo | URL, init?: RequestInit): Promise<Response>;
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -145,7 +145,7 @@ export interface Body extends Pick<BodyMixin, keyof BodyMixin> {}
 
 export type RequestRedirect = 'error' | 'follow' | 'manual';
 export type ReferrerPolicy = '' | 'no-referrer' | 'no-referrer-when-downgrade' | 'same-origin' | 'origin' | 'strict-origin' | 'origin-when-cross-origin' | 'strict-origin-when-cross-origin' | 'unsafe-url';
-export type RequestInfo = string | Request;
+export type RequestInfo = string | URL | Request;
 export class Request extends BodyMixin {
 	constructor(input: RequestInfo, init?: RequestInit);
 
--- a/@types/index.test-d.ts
+++ b/@types/index.test-d.ts
@@ -37,6 +37,7 @@ async function run() {
 	// Post
 	try {
 		const request = new Request('http://byjka.com/buka');
+		new Request(new URL('http://byjka.com/buka'));
 		expectType<string>(request.url);
 		expectType<Headers>(request.headers);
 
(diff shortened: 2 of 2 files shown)


@types/index.d.ts
```diff
@@ -145,7 +145,7 @@ export interface Body extends Pick<BodyMixin, keyof BodyMixin> {}
 
 export type RequestRedirect = 'error' | 'follow' | 'manual';
 export type ReferrerPolicy = '' | 'no-referrer' | 'no-referrer-when-downgrade' | 'same-origin' | 'origin' | 'strict-origin' | 'origin-when-cross-origin' | 'strict-origin-when-cross-origin' | 'unsafe-url';
-export type RequestInfo = string | Request;
+export type RequestInfo = string | URL | Request;
```
LinusU, line 148: The official TypeScript types have added `URL` as an acceptable parameter to `fetch`, but `RequestInfo` still doesn't include `URL`. 

https://github.com/microsoft/TypeScript/blob/f43cd0accac3e5033820fd27035731ed88f54938/lib/lib.dom.d.ts#L18281

Is there a reason why we are doing it differently here?

Maxim-Mazurok: That's a good question. It seems that in `lib.dom.d.ts` it's always used like this: `RequestInfo | URL`, except for `Cache.addAll`. However, I just tried in Chrome and it seems to accept `URL` just fine as well. So moving `URL` into `RequestInfo` seems like a pretty good optimization to me, I'll be opening such PR to TS shortly.

Maxim-Mazurok: Ah, never mind, seems like `| URL` is being added conditionally here: https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/dbe454edfe3804b0bbfa1ee1a0a88e9c0a1d4092/src/build/emitter.ts#L679
So I can't see an easy fix, even though semantically `RequestInfo = string | URL | Request` probably would make more sense.
Probably better to stick to what they have to avoid conflicts, updating now.

Maxim-Mazurok: Also added `URL` to `Request` constructor, as in TS: https://github.com/microsoft/TypeScript/blob/f43cd0accac3e5033820fd27035731ed88f54938/lib/lib.dom.d.ts#L11733

LinusU: Awesome! 🎉

LinusU (approved):
👍

Maxim-Mazurok:
CI fails for reasons unrelated to the change, it fails with the same errors on the main branch as well.
Following advice from https://github.com/node-fetch/node-fetch/issues/1657#issuecomment-1294221033 adding `tsconfig.json` with the following contents:
```json
{
	"compilerOptions": {
		"lib": ["DOM"]
	}
}
```
resolves the issue, however, I'm not sure if that's the correct way to do it.

If it was up to me - I'd force-merge this PR and address the `FormData` issues separately:
```
npm run test-types

> node-fetch@3.1.1 test-types
> tsd


  node_modules/formdata-polyfill/esm.min.d.ts:2:10
  ✖    2:10  FormData refers to a value, but is being used as a type here. Did you mean typeof FormData?  
  ✖    3:13  FormData refers to a value, but is being used as a type here. Did you mean typeof FormData?  
  ✖    5:49  FormData refers to a value, but is being used as a type here. Did you mean typeof FormData?  

  @types/index.d.ts:124:3
  ✖  124:3   FormData refers to a value, but is being used as a type here. Did you mean typeof FormData?  
  ✖  137:21  FormData refers to a value, but is being used as a type here. Did you mean typeof FormData?  

  node_modules/fetch-blob/from.d.ts:20:63
  ✖   20:63  File refers to a value, but is being used as a type here. Did you mean typeof File?          
  ✖   25:59  File refers to a value, but is being used as a type here. Did you mean typeof File?          

  node_modules/fetch-blob/file.d.ts:1:75
  ✖    1:75  Element implicitly has an any type because type typeof globalThis has no index signature.    

  8 errors
``` 

But happy to add the above mentioned fix if that sounds like a good idea?

Maxim-Mazurok:
Looking forward to having this published on npm, thanks!

Maxim-Mazurok:
@jimmywarting, I'm not familiar with the release cycle of this library, but I'm still waiting for this to be published in v3, thank you!

jimmywarting:
Kind of release a new version every time a PR with the commit `fix:` or `feat:` is merged. docs and d.ts is rarely never released

Maxim-Mazurok:
Ah, I wasn't aware of that... This probably should've been a `fix:`...
Perhaps adding commitizen or some pre-commit hook to validate commit messages would help new contributors?
I'm going to create a dummy PR with `fix:` to trigger release now.

PR #1704, opened 2023-01-18
Fix "global is not defined"

segevfiner:
## Purpose
I think this was missed in https://github.com/node-fetch/node-fetch/pull/1534

## Changes
Replace `global` -> `globalObject`

## Additional information
I get a `global is not defined` error otherwise. (octokit is still using this instead of isomorphic-fetch/cross-fetch)

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -17,7 +17,7 @@ module.exports = exports = globalObject.fetch;
 
 // Needed for TypeScript and Webpack.
 if (globalObject.fetch) {
-	exports.default = globalObject.fetch.bind(global);
+	exports.default = globalObject.fetch.bind(globalObject);
 }
 
 exports.Headers = globalObject.Headers;


LinusU (approved):
👍

jimmywarting:
I think we should remove getGlobal and just use `globalThis` instead. and do we really need to use bind?

All in all:
```js
"use strict";

const { fetch, Headers, Request, Response } = globalThis;

module.exports = exports = fetch;

// Needed for TypeScript and Webpack.
if (fetch) {
	exports.default = fetch;
}

exports.Headers = Headers;
exports.Request = Request;
exports.Response = Response;
```

NewEraCracker (approved):

jimmywarting:
sure

jimmywarting (approved):

Brynor (approved):

PR #1706, opened 2023-01-25
Changes in Response Class contructor for status range and null body status check

asingh04:
## Additional information

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix #000

### DIFF
--- a/src/response.js
+++ b/src/response.js
@@ -11,6 +11,22 @@ import {isRedirect} from './utils/is-redirect.js';
 const INTERNALS = Symbol('Response internals');
 
 /**
+ * Checks the provided 'status' property to the Response constructor
+ * @param {*} status
+ * @throws RangeError if the status variable is not in the acceptable range of [200, 599]
+ * @returns The provided status value if it is valid
+ */
+function validateStatusValue(status) {
+	const statusValue = Number.parseInt(status, 10);
+
+	if (Number.isNaN(statusValue) || statusValue < 200 || statusValue > 599) {
+		throw new RangeError(`Failed to construct 'Response': The status provided (${status}) is outside the range [200, 599].`);
+	}
+
+	return statusValue;
+};
+
+/**
  * Response class
  *
  * Ref: https://fetch.spec.whatwg.org/#response-class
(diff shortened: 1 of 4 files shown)


src/response.js
```diff
@@ -23,12 +39,18 @@ export default class Response extends Body {
...
-		const status = options.status != null ? options.status : 200;
+		// if status is not provided or is null, let's use default 200 value
+		// otherwise, lets check if the value is in correct range
+		const status =	(typeof options.status === 'undefined' || options.status === null) ? 200 :
+			validateStatusValue(options.status);
 
 		const headers = new Headers(options.headers);
 
-		if (body !== null && !headers.has('Content-Type')) {
+		if (status === 204 && typeof body !== 'undefined' && body !== null) {
```
jazelly, line 49: I think we also need to add `101` and `103`, due to [this note]( https://fetch.spec.whatwg.org/#dom-response-error:~:text=101%20and%20103%20are%20included%20in%20null%20body%20status%20due%20to%20their%20use%20elsewhere.%20They%20do%20not%20affect%20this%20step.)

src/response.js
```diff
@@ -119,7 +141,8 @@ export default class Response extends Body {
 	}
 
 	static error() {
-		const response = new Response(null, {status: 0, statusText: ''});
+		// default error status code = 400
+		const response = new Response(null, {status: 400, statusText: ''});
```
jazelly, line 145: I would argue for a default 500, as 400 normally indicates errors done by users. 

Imagine this, if a client forgets to define their status code and uses a default one, their users should be seeing a 500 as it's typically a mistake from server side.

Another note is I believe logical changes like this should be reflected in doc.

src/response.js
jazelly, line 49: I think we also need `205` and `304`, as per [the standard](https://fetch.spec.whatwg.org/#null-body-status:~:text=A%20null%20body%20status%20is%20a%20status%20that%20is%20101%2C%20103%2C%20204%2C%20205%2C%20or%20304.)

src/response.js
```diff
@@ -23,12 +39,18 @@ export default class Response extends Body {
 	constructor(body = null, options = {}) {
 		super(body, options);
 
-		// eslint-disable-next-line no-eq-null, eqeqeq, no-negated-condition
-		const status = options.status != null ? options.status : 200;
```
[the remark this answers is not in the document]
asingh04: @jazelly in this code
```javascript
const status = options.status != null ? validateStatusValue(options.status) : 200;
```

if the `status` is undefined( when the client/application hasnt prvovided the value), the default value `200` should be assumed, but the function `validateStatusValue` will be throwing an error. That's why added the check for `undefined` as well.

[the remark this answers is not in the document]
jazelly: Yes, but notice it's a `!=`. When `options.status` is `undefined`, `options.status != null` is falsy, and will set a default 200 to the status.

asingh04:
Hi @jazelly i will be re-working on this PR as I have mis-implemented the `Response` class and `status` property which does not match with the `Response` class available in the browsers

PR #1707, opened 2023-01-30
docs: readme - non ESM example

Michal-Miky-Jankovsky:
## Purpose
Fix old example in the README.md file.

## Changes
require -> import

## Additional information
according to different part of the same document
 > node-fetch from v3 is an ESM-only module - you are not able to import it with require()


jimmywarting (approved):

LinusU (approved):
👍

PR #1736, opened 2023-04-21
fix: Remove the default connection close header.

### LINKED ISSUE #1735: "socket hang up" / ECONNRESET on consecutive requests with Node.js 19 and Node.js 20
# Overview

When using `node-fetch` on Node.js 19+ to make two consecutive requests to the same web server, we always get the error: `FetchError: request to <URL> failed, reason: socket hang up`

## Details

This error only arises with `node-fetch` and Node.js 19+ (tested using 19.9.0 and 20.0.0), and the web server we were using was Axum. We didn't see this issue when using a node-based proxy to help with debugging - I imagine it depends slightly on the order of TCP packet delivery and latency and the internal state machines of the HTTP servers - but it seems like many HTTP servers would see this issue.

The error appears to only occur where the **second request to the same server in the same round of the event loop as the first request was awaited**, which gives the error:

```
FetchError: request to <URL> failed, reason: socket hang up
```

# Explanation 
## Root cause

We believe the following explains what we're seeing, and is backed up by logs at the end of this post:

* `fetch` sends a `connection: close` header in its first HTTP request...
  * The server responds with a http response, which `fetch` hands back to the caller.
  * The server actions the close connection request by following up the http response with a TCP packet with the `FIN` flag. This marks the end of the connection from the server's point of view. At this point, it does not expect to receive any further requests.
  
* Still in the same event loop, the caller calls fetch again to make its second HTTP request to the same origin.
  * Node.js attempts to send the next request to the host using _the same connection_
  * But the TCP connection is closed at the server's end, so returns a `conn reset` at the TCP layer
  * Resulting in the `socket hang up`

And I believe this may be the cause:

* The first response from the server is handed back to the caller likely in the [poll phase of the event loop round](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick#phases-overview)).
* In the mean time, the server has responded to close the TCP connection - but the event loop doesn't get round to handling it (not least because it's not yet in the `close callbacks` phase)
* The caller then attempts to send a new request. But as far as the event loop is concerned, we're still at the `poll` stage, and haven't yet handled the socket close event...
* So I would guess that Node.js internally has yet to close the connection (because it's waiting for the `close callbacks` stage?), so attempts to send the HTTP request on the same (as yet unclosed) connection.
* The server receives this request but has already closed the connection - so responds with connection reset message... which results in `fetch` raising the `socket hang` error.

Basically it's a race condition between the server closing the connection and the client sending the second request... but it happens reproducibly 100% of the time, and is caused by buggy behaviour, possibly in the Node.js agent?

I would imagine that Node.js 19 changed some internal handling of socket closes which delayed the handling of the close when a connection close http request is sent, and introduced - but I'm not sure if it's Node.js or `node-fetch` at fault.

# Actions
## Workarounds

The following both appeared to work - but for what it's worth, I don't think either of these is a "fix" because it's no longer an easy isomorphic drop in for the browser fetch.

* It appears that inserting the line `await new Promise((resolve) => setTimeout(resolve, 0))` before every API call to try to ensure any connections are closed fixes the problem.
* Using a [custom http / https agent](https://github.com/node-fetch/node-fetch#custom-agent) with `keepalive`, eg:
```ts
import http from 'node:http';
import https from 'node:https';

const agentSelector = function(_parsedURL: any) {
    if (_parsedURL.protocol == 'http:') {
        return httpAgent;
    } else {
        return httpsAgent;
    }
}

// ...

const response = await fetch(url, { agent: agentSelector });
```

## Suggestions for fixes

> TLDR: [These lines are incorrect in node-fetch](https://github.com/node-fetch/node-fetch/blob/bcfb71c7d10da252280d13818daab6925e12c368/src/request.js#L291)

There is arguablably issues in both `node-fetch` and Node.js itself.

However this is done, there are race conditions where a server can close a connection at the same time as you try to send another message on the connection - and this needs to be handled - I guess normally this is where applications implement retry logic (although it's an argument that libraries should have this - although maybe not `node-fetch`!)

BUT there is a perfect storm that makes this race condition happen every time (100% reproducible for me and my colleagues):
* It looks like Node.js 19 [changed the default agent to have keepalive true](https://github.com/nodejs/node/commit/4267b92604ad78584244488e7f7508a690cb80d0) - previously the default didn't use keep alive.
* But `node-fetch` is [adding a header to close connections if no agent is specified](https://github.com/node-fetch/node-fetch/blob/bcfb71c7d10da252280d13818daab6925e12c368/src/request.js#L292).

Previously, sending the `Connection: close` header explicitly didn't do anything wrong (even if it wasn't strictly needed). BUT sending a connection close down a keepalive connection isn't very nice behaviour - and trips up node's handling of the connection.

The following are three key fixes that could be made - all of which would fix the issue to various degrees:
* `node-fetch` should not default to sending a `Connection: Close` header if the default agent is set up to use keepalive... Maybe it shouldn't send the connection close header at all and should let Node.js handle it? (raised in #1736 )
* `Node.js` could fix itself so that if it receives a `FIN` it immediately doesn't attempt to send more requests on the same connection. (mentioned in this comment: https://github.com/nodejs/node/issues/47130 as point 1)
* `Node.js` could better pre-emptively handle the scenario where a caller sends a `Connection: close` header (mentioned in this comment: https://github.com/nodejs/node/issues/47130 as point 2)

### Related to point the node-fetch point:

The [node docs on http](https://nodejs.org/api/http.html) say this about the `keepalive` option on the agent:

> `keepAlive` [boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type) - Keep sockets around even when there are no outstanding requests, so they can be used for future requests without having to reestablish a TCP connection. Not to be confused with the `keep-alive` value of the `Connection` header. The `Connection: keep-alive` header is always sent when using an agent except when the Connection header is explicitly specified or when the `keepAlive` and `maxSockets` options are respectively set to `false` and `Infinity`, in which case `Connection: close` will be used. Default: `false`.

This suggests to me that Node.js has decent default handling, and we should be letting Node.js handle this and not take matters into our own hands by sending a confusing `Connection: close` header. By sending the header, we could be interfering with Node.js's ability to keep the socket alive if (for example) the `maxSockets` are reached.

Also - the fact we only set `Connection: close` on the request when a user doesn't provide an agent feels a little weird - either it's worth doing when the agent is set to keepalive or not - regardless of whether we use the default agent or not.

-Finally - perhaps `node-fetch` could consider consider respecting the `keepalive` option of `fetch` itself and create the correct agent - with a cache - to prevent it from using the `default` agent - or not even use the `default` agent at all.-
EDIT: Sorry - I'm wrong. The `keepalive` option on fetch is an entirely separate concept, to do with keeping the connection alive after the page ends in the browser (eg a `navigator.sendBeacon` replacement) and not related to http keepalive - which makes sense - fetch is more high-level than that.

# Logs

## TCP Dump
From running `tcpdump -A -ni lo tcp port 3333` to monitor the TCP messages between the server and the `node-fetch` client.

The key parts are ([see reference](https://gist.github.com/tuxfight3r/9ac030cb0d707bb446c7)):
* A 47630->3333 http connection:close request
* Server responds with the http response, followed by a Flags [F.] `FIN` packet (seq
* Another request attempt on the same connection (same port and seq number continues)
* Server response with the reset flag (Flags [R]) `RST` - reset connection packet.

```
19:28:09.351131 IP 127.0.0.1.47630 > 127.0.0.1.3333: Flags [P.], seq 1:225, ack 1, win 512, options [nop,nop,TS val 109329269 ecr 109329241], length 224

POST <PATH1> HTTP/1.1
Connection: close
Host: 127.0.0.1:3333

19:28:09.351166 IP 127.0.0.1.3333 > 127.0.0.1.47630: Flags [.], ack 225, win 510, options [nop,nop,TS val 109329269 ecr 109329269], length 0
19:28:09.352154 IP 127.0.0.1.3333 > 127.0.0.1.47630: Flags [P.], seq 1:2730, ack 225, win 512, options [nop,nop,TS val 109329270 ecr 109329269], length 2729
HTTP/1.1 200 OK
content-type: application/json
content-length: 2619
date: Wed, 19 Apr 2023 17:28:09 GMT

<<response-payload>>
19:28:09.352163 IP 127.0.0.1.47630 > 127.0.0.1.3333: Flags [.], ack 2730, win 497, options [nop,nop,TS val 109329270 ecr 109329270], length 0
19:28:09.352175 IP 127.0.0.1.3333 > 127.0.0.1.47630: Flags [F.], seq 2730, ack 225, win 512, options [nop,nop,TS val 109329270 ecr 109329270], length 0

// Second request

19:28:09.365378 IP 127.0.0.1.47630 > 127.0.0.1.3333: Flags [P.], seq 225:506, ack 2731, win 512, options [nop,nop,TS val 109329283 ecr 109329270], length 281
POST <PATH2> HTTP/1.1
Content-Type: application/json
Accept: */*
Content-Length: 24
User-Agent: node-fetch/1.0 (+https://github.com/bitinn/node-fetch)
Accept-Encoding: gzip,deflate
Connection: close
Host: 127.0.0.1:3333

<REQUEST_BODY_2>
19:28:09.365401 IP 127.0.0.1.3333 > 127.0.0.1.47630: Flags [R], seq 594996662, win 0, length 0
```

## `NODE_DEBUG=net` logs

From running `NODE_DEBUG=net yarn test`

```
/// FIRST REQUEST

NET 54331: createConnection [
  [Object: null prototype] {
    protocol: 'http:',
    slashes: true,
    auth: null,
    host: '127.0.0.1',
    port: '3333',
    hostname: '127.0.0.1',
    hash: null,
    search: null,
    query: null,
    pathname: '<PATH>',
    path: null,
    href: 'http://127.0.0.1:3333/<URL>',
    method: 'POST',
    headers: [Object: null prototype] {
      Accept: [Array],
      'Content-Length': [Array],
      'User-Agent': [Array],
      'Accept-Encoding': [Array],
      Connection: [Array]
    },
    agent: undefined,
    keepAlive: true,
    scheduling: 'lifo',
    timeout: 5000,
    noDelay: true,
    servername: '',
    _agentKey: '127.0.0.1:3333:',
    encoding: null,
    keepAliveInitialDelay: 1000
  },
  [Function (anonymous)],
  [Symbol(normalizedArgs)]: true
]
NET 54331: pipe false null
NET 54331: connect: attempting to connect to 127.0.0.1:3333 (addressType: 4)
NET 54331: _read - n 16384 isConnecting? true hasHandle? true
NET 54331: _read wait for connection
NET 54331: afterConnect
NET 54331: _read - n 16384 isConnecting? false hasHandle? true
NET 54331: Socket._handle.readStart
NET 54331: _read - n 16384 isConnecting? false hasHandle? true

/// SECOND REQUEST

NET 54331: destroy
NET 54331: close
NET 54331: close handle
```

dhedey:
## Purpose
This fixes #1735 and likely replaces #1473.

## Changes

Removes the behaviour to add a `Connection: close` header. This was causing issues in Node.js 19+ (see #1735), but possibly subtle issues in earlier Node.js too. Note that we still expect some users to use `node-fetch` in Node.js 19+ because enabling fetch requires an `--experimental-fetch` command line parameter passed to Node.js and some users (such as people consuming my library which requires a `fetch` implementation) require it.

Instead, we rely on the underlying http implementation in Node.js to handle this, as per the documentation at
https://nodejs.org/api/http.html#new-agentoptions

The [original change introducing this](https://github.com/node-fetch/node-fetch/commit/af21ae6c1c964cd40e48e974af56ea2e1361304f) provided no clear motivation for providing the header, and [the implementation has since been changed to disable this header when an agent is provided](https://github.com/node-fetch/node-fetch/commit/7f68577de44c7d1efe7009b212f7c54a0b4a709b), so I think there is sufficient evidence that removing this header is the correct behaviour.  

If this change is accepted, I'm happy to back-port this into the 2.x branch, as I know some users of my library use 2.x because they can't use ESM.

## Additional information
See  #1735 for background.

___

- [X] I updated the readme
(I didn't add unit tests because the behaviour now depends the presence of the header now depends on the default http agent behaviour, and therefore on the Node.js version being used and I didn't want to introduce that dependency into the tests)

___

- fix #1735 
- replaces #1473

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,7 +539,6 @@ If no values are set, the following request headers will be sent automatically:
 | ------------------- | ------------------------------------------------------ |
 | `Accept-Encoding`   | `gzip, deflate, br` (when `options.compress === true`) |
 | `Accept`            | `*/*`                                                  |
-| `Connection`        | `close` _(when no `options.agent` is present)_         |
 | `Content-Length`    | _(automatically calculated, if possible)_              |
 | `Host`              | _(host and port information from the target URI)_      |
 | `Transfer-Encoding` | `chunked` _(when `req.body` is a stream)_              |
@@ -557,6 +556,8 @@ The `agent` option allows you to specify networking related options which are ou
 - Custom DNS Lookup
 
 See [`http.Agent`](https://nodejs.org/api/http.html#http_new_agent_options) for more information.
+
+If no agent is specified, the default agent provided by Node.js is used. Note that [this changed in Node.js 19](https://github.com/nodejs/node/blob/4267b92604ad78584244488e7f7508a690cb80d0/lib/_http_agent.js#L564) to have `keepalive` true by default. If you wish to enable `keepalive` in an earlier version of Node.js, you can override the agent as per the following code sample. 
 
 In addition, the `agent` option accepts a function that returns `http`(s)`.Agent` instance given current [URL](https://nodejs.org/api/url.html), this is useful during a redirection chain across HTTP and HTTPS protocol.
 
--- a/src/request.js
+++ b/src/request.js
@@ -288,10 +288,6 @@ export const getNodeRequestOptions = request => {
 		agent = agent(parsedURL);
 	}
 
-	if (!headers.has('Connection') && !agent) {
-		headers.set('Connection', 'close');
-	}
-
 	// HTTP-network fetch step 4.2
 	// chunked encoding is handled by Node.js
 


LinusU (approved):
Since this should only affect newer versions of Node.js which previously didn't work correctly, I don't think that this would constitute a major version bump.

LGTM 👍

dhedey:
Thanks @LinusU :).

Is there anything more I can do to help get this merged? Would it help if I raised a separate PR to backport this to the 2.x branch now?

matthewkeil:
Hi there.  We are end users through `cross-fetch` having issues upgrading our system from Node 18 -> 20.  This PR will ultimately help us get updated.  We are curious if there is anything we can assist with to help @dhedey get this across the finish line.  He did an excellent job researching and we are super grateful for his help!!!  Please let us know as we eagerly await this merge

matthewkeil:
Thank you @jimmywarting and @dhedey  🎉

nflaig:
Awesome work guys! It has been really insightful to follow the discussions here and on the node issue. I am wondering if you also plan to backport this to 2.x branch, this would be really appreciated. I guess with more people updating to node 20 it is also quite important to do.

jimmywarting:
Dont know about backporting to v2 is really necessary. It is mostly frezed and dose not really get any new feature updates. Only critical bug fixes and security related stuff.
Tried fixing a BOM text issue by using same text decoder in v2 but it broke som things cuz it was not really well backward compatible for older NodeJS versions. So it has to be reverted quickly. 

Cjs user can still import esm version if they use async import instead. 

const fetch = (...args) => import('node-fetch').then(({default: fetch}) => fetch(...args))

Or update to a node version that has fetch built in

nflaig:
I understand your concerns and might be good to not touch v2, assuming those users are on older node version anyways. The problem is rather with cross-fetch not updating to node-fetch v3, but this has been proposed a while ago already https://github.com/lquixada/cross-fetch/issues/146.

dhedey:
@jimmywarting

Can I start this by saying - I've used node-fetch a lot over the years, and really appreciate the hard work and stewardship of you and the other contributors to this project 👍.

This is a critical bug in my opinion, simply breaking many integrations when using `node-fetch` with more recent versions of node, so in my opinion is really deserving of a fix to the version 2 branch.

I believe this change is low-risk, for a couple of reasons:
* Note that the Connection Close isn't included if an `agent` is provided, which has seen no reported issues.
* I don't believe `Connection: Close` was ever needed. Even on eg Node v4 docs for the default agent, it doesn't mention sending a `Connection: Close` header: https://nodejs.org/docs/latest-v4.x/api/http.html#http_http_globalagent.

I'm happy to put in that PR to version 2 if that's a help.

To add some flavour regarding attention to version 2 - the reality is that:
* Libraries like ours choose `node-fetch` as a polyfill or fallback (or, more specifically, we provide documentation on how our users can use `node-fetch`), and such libraries can't choose to upgrade to v3 without forcing all of _our_ users off cjs / onto ESM, which simply isn't something we are able or willing to do.
* Looking at the version history, across v2 versions, v2 has over 10x the downloads of v3 in the last week: https://www.npmjs.com/package/node-fetch?activeTab=versions - [2.6.12](https://www.npmjs.com/package/node-fetch/v/2.6.12) got 10,648,684
* Why?
  * One of the primary reasons for still using node-fetch now is to support these older applications, or, importantly for building libraries which need to be compatible with different nodejs versions / build / packaging paradigms :). This tail-end of older applications (or, importantly, libraries!) needing compatibility is the main `node-fetch` user base (based on the numbers above). And libraries often can't control which versions of NodeJS they're run under.
  * I would imagine v3 hasn't really taken off because if people are able to use ES modules or using async import of `fetch` then in most cases, they will likely be building an application themselves (not a library) and can just use node's built in fetch. 

It's worth noting that I tried just using `const fetch = (...args) => import('node-fetch').then(({default: fetch}) => fetch(...args))` but importing an ESM module didn't work when I needed to support CJS builds for upstream users. Maybe I did something wrong - but I don't believe broadly that this approach works for libraries.

Don't get me wrong, I'm not trying to criticise v3 at all! But I just wanted to emphasise, that the main user base is still on v2, and in my opinion, almost certainly will always be - because only v2 really captures the USP of "why do I want to use `node-fetch`?" - which (in my experience) is either:
* (A) To provide a `fetch` implementation for my application built on a legacy stack running an old version of node, for reasons outside of my control (grr managers / legacy systems). Such applications are unlikely to be able to support ESM. The workaround above might be possible with some installations/bundlers, but is sufficiently hard to find/understand that many devs I know would just stick on v2 which "seems to work out of the box". OR
* (B) To provide a universally runnable `fetch` implementation for a library capable of running in different node/build configurations, many of which don't support ESM or have fetch enabled

And v3 doesn't work so well for either (A) or (B), so most projects are still running v2 (or would migrate off to `node`'s built in fetch if they upgraded).

jimmywarting:
If you feel like backporting it then go for it

dhedey:
Fantastic, thanks @jimmywarting - Appreciate your support on this.

I've raised the PR here: https://github.com/node-fetch/node-fetch/pull/1765 - over to you to approve and release.

Thanks again!

README.md
```diff
@@ -558,6 +557,8 @@ The `agent` option allows you to specify networking related options which are ou
 
 See [`http.Agent`](https://nodejs.org/api/http.html#http_new_agent_options) for more information.
```
vahid543, line 558: ¹¹

vahid543 (commented):
- [ ]

PR #1482, opened 2022-01-25
fix: handle bom in text and json

jimmywarting:
## Purpose
long over due problem to an easy fix... `.text()` & `.json()` didn't handle BOM correctly

## Changes
switch `buffer.toString()` to `TextDecoder().decode()`

___

- [x] I added unit test(s)

___

- fix #541

PR #1739, opened 2023-05-01
fix: handle bom in text and json

aclarembeau:
## Purpose

Bring the fix https://github.com/node-fetch/node-fetch/pull/1482/commits from 3.X to 2.X

## Changes

Allow parsing results with BOM encoding

## Additional information

It's just a copy paste of the fix of https://github.com/node-fetch/node-fetch/pull/1482/commits for node-fetch 2.X, for the people who can't use 3.X
___

- [ ] I updated readme
- [x] I added unit test(s)

___

- fix #000

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "dependencies": {
         "whatwg-url": "^5.0.0"
     },
-    "peerDependencies": { 
+    "peerDependencies": {
         "encoding": "^0.1.0"
     },
     "peerDependenciesMeta": {
--- a/src/body.js
+++ b/src/body.js
@@ -114,12 +114,8 @@ Body.prototype = {
 	 * @return  Promise
 	 */
 	json() {
-		return consumeBody.call(this).then((buffer) => {
-			try {
-				return JSON.parse(buffer.toString());
-			} catch (err) {
-				return Body.Promise.reject(new FetchError(`invalid json response body at ${this.url} reason: ${err.message}`, 'invalid-json'));
-			}
+		return this.text.then((text) => {
+			return JSON.parse(text); 
 		})
 	},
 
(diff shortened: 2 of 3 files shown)


jimmywarting (approved):

smeet666:
I would really love to see this change accepted, I'm in the exact same case where I can't upgrade to 3.0 **now** but I don't like too much to patch it on my side.

jimmywarting:
Normally the process have been that at least 2 core members needs to review and accept, if at least 2 agrees then we will merge. 
But if it may take to long or nobody have time to review then 1 can be enough.

we also don't do so much in the v2 branch anymore unless it's a security issue.

ping @LinusU 

---

> I can't upgrade to 3.0

any particular reason why you can't update?
there isn't any problem loading ESM module from CJS
you just have to load it lazily with `import('node-fetch').then(...)` read more [here](https://github.com/node-fetch/node-fetch/issues/1279) of how you can load v3

LinusU (approved):
Neat! 👍

jon-esperanza:
This has deprecated Node.js 10 for this node-fetch version due to TextDecoder being used as a global object.

TextDecoder became a global object in Node.js 11.
Screenshot 2023-05-08 at 4 44 44 PM

LinusU:
@jon-esperanza the breakage was not intentional, we are reverting this now. Thank you for bringing this to our attention.

revert progress: #1741

smeet666:
Is there any status on this one, then? Is it abandoned? It's marked as merged, but seems to be reverted.

jimmywarting:
Yes. it's reverted b/c it broke backward comp. 
`TextDecoder` was not available in those older NodeJS 

Must come up with a own solution that strips BOM manually without using it.

PR #1757, opened 2023-06-18
feat: optional get raw headers as string

mstruebing:
## Purpose

Most libraries expect raw headers to be a string instead of an array  so I added an optional argument which will return the raw headers as a string rather than a `string[]`.

## Changes

Added an optional argument to the `headers.raw()` function to return the header values as a string in case this is set to `true.

## Additional information

I could also create a second function like: `headers.rawAsString()` instead of adding an argument to the existing function.

___

- [ ] I updated readme
- [x] I added unit test(s)

___

Doesn't fix but is related to #783

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -65,7 +65,7 @@ export class Headers {
 	values(): IterableIterator<string>;
 
 	/** Node-fetch extension */
-	raw(): Record<string, string[]>;
+	raw(): Record<string, string[] | string>;
 }
 
 export interface RequestInit {
--- a/src/headers.js
+++ b/src/headers.js
@@ -197,11 +197,18 @@ export default class Headers extends URLSearchParams {
 	/**
 	 * Node-fetch non-spec method
 	 * returning all headers and their values as array
-	 * @returns {Record<string, string[]>}
+	 * @returns {Record<string, string[] | string>}
 	 */
-	raw() {
+	raw(asString = false) {
+		if (asString)  {
+
+		}
 		return [...this.keys()].reduce((result, key) => {
 			result[key] = this.getAll(key);
+
+			if (asString) {
+				result[key] = result[key].join(', ');
+			}
 			return result;
 		}, {});
 	}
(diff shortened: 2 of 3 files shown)


jimmywarting:
ideally ppl should just stop depending on raw headers all together.
the hope is for ppl to just stop depending node-fetch all together and use built in fetch when it's available. 

The reason ppl wanted to have raw in the first place was cuz the fetch spec decided that they should normalize the headers to be joined by `,` if one header is appended more than once.
some ppl where not happy about this cuz it made reading set-cookies hard that could include `,` in the value.

so ppl requested the `raw` and we implemented it...

now there is a new spec that mention iterating over headers that dose not combine the set-cookies but instead yields the everything.

ppl should really be using `[...response.headers]` or `Object.fromEntires(response.headers)` or even `Headers.prototype.getSetCookie`

that is what every other environment have agree upon of how it should work. 
it essentially match our raw method.

I think that out raw method should be deprecated and ppl should stop using it and strive more towards cross compatible solutions with other env / fetch implementations. 

so i'm generally -1 on this.

jimmywarting:
there are some new Header methods / spec things we need to fix...

mstruebing:
Okay, I'm using `Object.fromEntires(response.headers)` now but I thought that it would be useful to other people too who get into the same situation.

Does this mean I should just close this PR?

jimmywarting:
i can close it for you.

if you like you could make a suggestion in the readme that the best way to access the headers would be to either use:
`[...response.headers]` or `Object.fromEntires(response.headers)` instead of using the legacy/non-standard raw method

and perhaps maybe even put a deprecation warning on using the `raw` option. 

i did work on a more spec'ed Header classed here https://www.npmjs.com/package/fetch-headers the idea was for node-fetch to adapt it and let headers live on a separate place. as it has very little to do with fetch/request/response and is mostly isolated from everything else. then node-fetch would extend it to include the `raw` option (just for backwards comp. reasons)

if you feel like you are willing to bring in fetch-headers into node-fetch and replace our own than that would be appreciated as well. 
`fetch-headers` runs it own test against WPT so some test in node-fetch could be removed.

the missing thing is `getSetCookies`

PR #1763, opened 2023-07-02
remove async from json parse function

israelCohen90:
## Additional information

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix #000

### DIFF
--- a/src/body.js
+++ b/src/body.js
@@ -145,7 +145,7 @@ export default class Body {
 	 * @return  Promise
 	 */
 	async json() {
-		const text = await this.text();
+		const text = this.text();
 		return JSON.parse(text);
 	}
 
@@ -155,7 +155,7 @@ export default class Body {
 	 * @return  Promise
 	 */
 	async text() {
-		const buffer = await consumeBody(this);
+		const buffer = consumeBody(this);
 		return new TextDecoder().decode(buffer);
 	}
 
@@ -220,7 +220,7 @@ async function consumeBody(data) {
 	let accumBytes = 0;
 
 	try {
-		for await (const chunk of body) {
+		for (const chunk of body) {
 			if (data.size > 0 && accumBytes + chunk.length > data.size) {
 				const error = new FetchError(`content size at ${data.url} over limit: ${data.size}`, 'max-size');
 				body.destroy(error);


jimmywarting:
This can't be done.
getting the text of a response is not sync so it needs to "await" the response

LinusU:
@israelCohen90 the parsing needs to be async because it first needs to wait for the text to arrive before it can parse it as JSON...

jimmywarting:
cuz consuming / fetching a response from a API or anything is being done done asynchronous. fetching something is non-blocking

The `body` can't be read in a synchronous manner cuz it could be a readable stream that do have `Symbol.asyncIterator` 
https://github.com/node-fetch/node-fetch/blob/7b86e946b02dfdd28f4f8fca3d73a022cbb5ca1e/src/body.js#L217-L223

So you can't do a sync loop to consume it cuz you don't have a the data yet.

So we need to use a `for..await` loop to get all the chunks from the stream. and the `consumeBody` function is also async.
https://github.com/node-fetch/node-fetch/blob/7b86e946b02dfdd28f4f8fca3d73a022cbb5ca1e/src/body.js#L194

So when we call `consumeBody()` it dose actually return a `Promise`.

So it would mean that this:
```js
	async text() {
		const buffer = consumeBody(this);
		return new TextDecoder().decode(buffer);
	}
```
would cause a error when trying to decode it:
```js
	async text() {
		const promise = consumeBody(this)
    // Uncaught TypeError: Failed to execute 'decode' on 'TextDecoder':
    // The provided value is not of type '(ArrayBuffer or ArrayBufferView)'.
    // (received promise)
		return new TextDecoder().decode(promise)
	}
```

---

a simple demostration why it can't be async would be this:

```js
class F {
  async text() {
    return 'true'
  }
  async json() {
    return JSON.parse(this.text())
  }
}

new F().json()
Promise {<rejected>: SyntaxError: Unexpected token 'o', "[object Promise]" is not valid JSON
    at JSON.parse (<anonymo…}
```
you would have to remove async from the text function as well. 

```js
class F {
  text() {
    return 'true'
  }
  async json() {
    return JSON.parse(this.text())
  }
}

await new F().json() // true
```

PR #1774, opened 2023-09-07
docs(README): when patching global context, patch TextDecoder and TextEncoder too

jtomaszewski:
After following the README on patching the global context, node-fetch fails to me in jest jsdom environment on node v16.16.0 with following error:

```
TextDecoder is not defined

      Cause: ReferenceError: TextDecoder is not defined
      at Response.TextDecoder (node_modules/node-fetch/src/body.js:159:14)
      at call (node_modules/@babel/runtime/helpers/regeneratorRuntime.js:44:17)
      at Generator.tryCatch (node_modules/@babel/runtime/helpers/regeneratorRuntime.js:125:22)
      at Generator._invoke [as next] (node_modules/@babel/runtime/helpers/regeneratorRuntime.js:69:21)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:9)
```

That's probably because node-fetch in https://github.com/node-fetch/node-fetch/blob/main/src/body.js#L159 refers to `TextDecoder` using a global reference, which doesn't exist in my environment.

## Changes

In recommended script in README that patches the global context, add `TextDecoder` and `TextEncoder` to the global context too.

I also removed a few unnecessary imports from it that weren't used at all.

## Additional information

- [x] I updated readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,26 +141,28 @@ To use `fetch()` without importing it, you can patch the `global` object in node
 // fetch-polyfill.js
 import fetch, {
   Blob,
-  blobFrom,
-  blobFromSync,
   File,
-  fileFrom,
-  fileFromSync,
   FormData,
   Headers,
   Request,
   Response,
-} from 'node-fetch'
+} from 'node-fetch';
+import { TextDecoder, TextEncoder } from 'util';
 
 if (!globalThis.fetch) {
-  globalThis.fetch = fetch
-  globalThis.Headers = Headers
-  globalThis.Request = Request
-  globalThis.Response = Response
+  globalThis.fetch = fetch;
+  globalThis.Blob = Blob;
+  globalThis.File = File;
+  globalThis.FormData = FormData;
+  globalThis.Headers = Headers;
+  globalThis.Request = Request;
+  globalThis.Response = Response;
+  globalThis.TextDecoder = TextDecoder;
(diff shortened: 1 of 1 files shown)


jimmywarting:
Since this is just readme and no breaking changes, then i would now have suggested another approach.

using conditional lazy top level `import()` and maybe [Nullish coalescing assignment (??=)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) in order to not load it in vain

So maybe something like this:
```js
// fetch-polyfill.js
globalThis.fetch || await import('node-fetch').then(mod => {
  globalThis.fetch = mod.default
  globalThis.Blob ??= mod.Blob
  globalThis.File ??= mod.File
  globalThis.FormData ??= mod.FormData
  globalThis.Headers = mod.Headers
  globalThis.Request = mod.Request
  globalThis.Response = mod.Response
})

globalThis.TextDecoder || import('node:util').then(util => {
  globalThis.TextDecoder ??= util.TextDecoder
  globalThis.TextEncoder ??= util.TextEncoder
})

// index.js
(globalThis.TextDecoder && globalThis.fetch) || await import('./fetch-polyfill.js')
```

Issue #1519, opened 2022-03-09
aborting should throw a doeexception


jimmywarting:
Hmm, wondering if not we should try to fix the real problem that is that we return a custom AbortError message, it should rather be a `DOMException` 

```js
// Browser test:
ctrl = new AbortController() 
ctrl.abort()
fetch('/', ctrl).catch(console.log) 
// DOMException { 
//   name: AbortError,
//   code: 20,
//   message: "Failed to execute 'fetch' on 'Window': The user aborted a request." 
// }
```

We could optionally start to depend on [node-domexception](https://www.npmjs.com/package/node-domexception). it's already included by another sub dependency of ours (fetch-blob).

_Originally posted by @jimmywarting in https://github.com/node-fetch/node-fetch/issues/1513#issuecomment-1051110449_

PR #1775, opened 2023-09-08
support `AbortSignal.reason`

### LINKED ISSUE #1462: AbortController abort reason Parameter
## Description

In the definition of the [AbortController on MDN](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) it is stated, that it is possible to pass a parameter to `AbortController.abort()` that is thrown as Error if the AbortController is used in a fetch call.

## Current

`AbortController.abort()` exists, but does not expect a parameter.

## Wanted

Like done in the [@types/mongodb] implementation ([see here](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44549)), the optional reason parameter is also needed in the [@types/node] implementation. And i think before that makes sense, it has to be implemented here in node-fetch

## Version

Im using node-fetch as a transitive dependency on version `2.6.6`.
I Set up a project using [next.js](https://github.com/vercel/next.js/) using `^12.0.7` and [ts-node](https://github.com/TypeStrong/ts-node) using `^10.4.0`

gomain:
In enviroments that implement it.

When an `AbortSignal#reason` has a value, i.e. in nodejs >= v17.2.0, it is added on the returned `AbortError`. `AbortError#reason` would be `undefined` otherwise.

Contrary to the specification, when `AbortController#abort` is called, with or without a reason, `fetch` _does not_ throw the `AbortSignal#reason`. This is to allow more fine grained error handling. Had the `reason` was thrown, the general information that an abort happened is lost and it becomes sole responsibility of the reason provider.

## Purpose

Allow detecting the _reason_ fetch was aborted.

## Changes

- Extend `AbortError` to have field `reason`. It's constructor takes this additional argument.
- The `AbortSignal#reason` is given when constructing `AbortError` instance. Could possibly be `undefined` in environments/libraries/polyfills that do not support it.

## Additional information

The actual value (even existence) of `AbortSignal#reason` is external to this library. We make no assumptions about it. We simply forward it.

This goes against #1519. There is no inherent reason to follow the specifications to the dot. `node-fetch` is for node, specifically, not the web. TBD.

___

- [ ] I updated readme
- [x] I added unit test(s)

___

- fix #1462

### DIFF
--- a/src/errors/abort-error.js
+++ b/src/errors/abort-error.js
@@ -4,7 +4,9 @@ import {FetchBaseError} from './base.js';
  * AbortError interface for cancelled requests
  */
 export class AbortError extends FetchBaseError {
-	constructor(message, type = 'aborted') {
-		super(message, type);
+	constructor(message, reason) {
+		super(message, 'aborted');
+
+		this.reason = reason;
 	}
 }
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -13,6 +13,7 @@ import {
 
 type AbortSignal = {
 	readonly aborted: boolean;
+	readonly reason: unknown;
 
 	addEventListener: (type: 'abort', listener: (this: AbortSignal) => void) => void;
 	removeEventListener: (type: 'abort', listener: (this: AbortSignal) => void) => void;
@@ -212,6 +213,7 @@ export class FetchError extends Error {
 export class AbortError extends Error {
 	type: string;
 	name: 'AbortError';
+	reason: DOMException | unknown;
 	[Symbol.toStringTag]: 'AbortError';
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -67,7 +67,7 @@ export default async function fetch(url, options_) {
 		let response = null;
 
 		const abort = () => {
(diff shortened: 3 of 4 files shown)


src/errors/abort-error.js
```diff
@@ -4,7 +4,9 @@ import {FetchBaseError} from './base.js';
  * AbortError interface for cancelled requests
  */
 export class AbortError extends FetchBaseError {
-	constructor(message, type = 'aborted') {
```
LinusU, line 7: Is this a breaking change?

`AbortError` is exported publicly from the package...

gomain: Yeah, wasn't aware the constructor was actually exported. It just seemed odd that the `type` would be explicitly given, i.e. what else could it be.

We basically need to thread the `reason` into the thrown `AbortError`. To not break its constructor, the only sane option would be to extend it:
```js
  // no change to existing `AbortError`
  class AbortWithReasonError extends AbortError {
    constructor(message, reason) {
      super(message);
      this.reason = reason;
    }
  }
```

PR #1793, opened 2023-12-04
chore: avoid deprecation warning

homura:
## Purpose

NodeJS 21 warns that the `punycode` is deprecated. The `whatwg-url@5.x` dependent by `node-fetch` depends on the deprecated `punycode`, therefore, it is better to upgrade the `whatwg-url` to the latest one 

https://github.com/jsdom/whatwg-url/issues/261

To reproduce it, try the below piece example

```js
const fetch = require('node-fetch') // 2.7.0

fetch('https://jsonplaceholder.typicode.com/todos/1')
      .then(response => response.json())
      .then(json => console.log(json))
```

```
➜  node-fetch git:(2.7) ✗ node x.js
(node:4461) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
{ userId: 1, id: 1, title: 'delectus aut autem', completed: false }
```

## Changes

Upgrade `whatwg-url` to the latest

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     },
     "homepage": "https://github.com/bitinn/node-fetch",
     "dependencies": {
-        "whatwg-url": "^5.0.0"
+        "whatwg-url": "^14.0.0"
     },
     "peerDependencies": { 
         "encoding": "^0.1.0"


jimmywarting:
Uhm, sry... we can't
whatwg-url@14 requires NodeJS v18

where as node-fetch@2.x still has backwards compatible all the way back to NodeJS v4

node-fetch@v2.x is currently locked to only receiving bug/security fixes. 

---

Your option is either to update to node-fetch@3 or using the new built-in fetch from NodeJS themself now that it's shipped with NodeJS v18+

rlindner81:
> Uhm, sry... we can't whatwg-url@14 requires NodeJS v18
> 
> where as [node-fetch@2.x](mailto:node-fetch@2.x) still has backwards compatible all the way back to NodeJS v4
> 
> [node-fetch@v2.x](mailto:node-fetch@v2.x) is currently locked to only receiving bug/security fixes.
> 
> Your option is either to update to node-fetch@3 or using the new built-in fetch from NodeJS themself now that it's shipped with NodeJS v18+

@jimmywarting I don't follow this logic. nodejs below v18 is _out of maintenance_. 
https://nodejs.org/en/about/previous-releases

So, it clearly falls under "security fixes" to update `node-fetch@2.x` to at least nodejs v18... Our project, and I would expect the majority of others, uses commonjs and are locked into `node-fetch@2.x`. Unfortunately, the native fetch in nodejs v18 (or v20) is not really usable enough, when it comes to streaming to replace node-fetch.

radist2s:
@rlindner81, I agree.

I guess I'm not the only one who's intimidated by the Node.js 22 warning:
```
(node:90843) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```

tbroyer:
punycode was removed from whatwg-url in v9, which is the version that dropped support for Node.js 10: https://github.com/jsdom/whatwg-url/releases/tag/v9.0.0

This means that node-fetch v2 could safely be updated to some version of whatwg-url between v9 and v12 (v13 being the one removing support for Node.js 14)

In the mean time, adding an override in my package.json (as many others have done already) does the job.

PR #1799, opened 2024-01-10
refactor: allow exclude user agent using null value

PrimeWalkerVN:
## Purpose
Some contexts, the default user agent of the request does not need to contain of header list, so use null value to exclude it from header list of the request

## Changes
Change logic of setting user-agent request's header

## Additional information

___

- [x] I updated src/request.js

___

- fix #1799


trantrongkim98:
- LGTM

PR #1800, opened 2024-01-18
upgrade fetch-blob

sirenkovladd:
## Purpose
Upgrade version fetch-blob from 3.1.4 to 4.0.0
add to cicd version of node "18.18.2", "20.11.0", "21.5.0"
this update will reduce the size of dependencies from about 9.1MB to 0.4MB

## Changes
- changed package.json
- in test "supports supplying a famliy option to the agent" add custom callback since http.Agent->lookup changed from v20
- fixed for >v20 route /chunked/multiple-ending in custom server

___

- [ ] I updated readme
- [ ] I added unit test(s)
- [X] checked tests for versions 14.21.3, 16.20.2, 18.18.2, 20.11.0, 21.5.0

___

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        node: ["12.20.0", "14.13.1", "16.0.0"]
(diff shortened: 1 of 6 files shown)


sirenkovladd:
@jimmywarting @xxczaki

Martinspire:
would be nice to see an update here

PR #1829, opened 2024-11-09
Add abort reason as the AbortError cause

ikonst:
To better capture the cause of the abort, we can attach the abort signal's [reason](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) as the cause of the `AbortError` that we throw.

___

- [ ] I updated readme
- [ ] I added unit test(s)


ikonst:
I can add tests if we think it's a good idea to make this change.

PR #1835, opened 2025-02-02
Update README.md

its2badwespend:
fetch('https://github.com/')
    .then(res => {
        console.log(res.ok);
        console.log(res.status);
        console.log(res.statusText);
        console.log(res.headers.raw());
        console.log(res.headers.get('content-type'));
    });

## Additional information

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix #000


its2badwespend:

PR #1860, opened 2026-02-13
Fix FetchError.errno type to match Node.js SystemError

### LINKED ISSUE #1831: Why is FetchError.errno a string and not a number?
Why is errno a string and not a number according to [SystemError](https://nodejs.org/docs/latest-v20.x/api/errors.html#class-systemerror)

```
export class FetchError extends Error {
	constructor(message: string, type: string, systemError?: Record<string, unknown>);

	name: 'FetchError';
	[Symbol.toStringTag]: 'FetchError';
	type: string;
	code?: string;
	errno?: string;
}
```

veeceey:
This fixes the TypeScript type definition for `FetchError.errno` to properly match Node.js `SystemError` spec, where `errno` is a number rather than a string.

## Changes

1. **TypeScript definitions** (`@types/index.d.ts`):
   - Changed `errno?: string` to `errno?: number` 
   - Added `erroredSysCall?: string` property (was missing but set in implementation)

2. **Implementation** (`src/errors/fetch-error.js`):
   - Fixed to assign `errno` from `systemError.errno` (number) instead of incorrectly using `systemError.code` (string)

3. **Tests** (`test/main.js`):
   - Updated custom FetchError test to properly set `errno` as a number
   - Removed errno assertion from network failure test (platform-specific numeric value)

## Testing

Ran full test suite - all existing tests pass (3 failures are pre-existing flaky network tests unrelated to this change).

Fixes #1831


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #1861, opened 2026-02-13
fix: resolve punycode deprecation warning by updating whatwg-url

### LINKED ISSUE #1797: Faulty NPM upload 2.7.0
I have noticed a deprecation warning for _punycode_ and searched for the problem. 

The problem was with _whatwg-url_ **v5** and regarding issue #1794 this is already known. The pull request #1795 addresses this issue and updates the dependency of _whatwg-url_ from **v5** to **v14**.

The change list of the pull request does indeed reflect the correct change:

```json
    "name": "node-fetch",
    "version": "2.6.7",

    "dependencies": {
        "whatwg-url": "^5.0.0"
    },
```

```json
    "name": "node-fetch",
    "version": "2.7.0",

    "dependencies": {
        "whatwg-url": "^14.0.0"
    },
```

But when I actually install _node-fetch_ **v2.7.0** via npm or [look at the code on npmjs.com](https://www.npmjs.com/package/node-fetch/v/2.7.0?activeTab=code) there is still the previous dependency to _whatwg-url_ **v5**:

```json
    "name": "node-fetch",
    "version": "2.7.0",

    "dependencies": {
        "whatwg-url": "^5.0.0"
    },
```

This means that the issue #1794 is not resolved with the currently available version of _node-fetch_ **v2**!

### LINKED ISSUE #1826: Deprecation issue with the punycode module in your project
Hello,

I am encountering a deprecation issue with the punycode module in your project. The warning message is as follows:

(node:19880) [DEP0040] DeprecationWarning: The punycode module is deprecated. Please use a userland alternative instead.

Could you consider updating your project to use a more recent version of punycode (e.g., punycode@2.3.1)?

Thank you in advance for your help.

Best regards,
valorisa

veeceey:
Bumps `whatwg-url` from `^5.0.0` to `^7.0.0` to fix the punycode deprecation warning (DEP0040) that shows up on Node.js 21+.

The root cause is that `whatwg-url@5.0.0` depends on `tr46@~0.0.3`, which uses Node's built-in `punycode` module — deprecated since Node 21. Updating to `whatwg-url@7` pulls in `tr46@^1.0.1`, which depends on the userland `punycode` npm package instead, so the warning goes away.

I went with v7 rather than a higher version because it has no `engines` restriction, keeping backward compat with older Node versions that 2.x still supports. The only API node-fetch uses from whatwg-url is the `URL` constructor (as a polyfill fallback), which is fully compatible across versions.

Verified locally:
- All existing tests pass (same results as unmodified 2.x — the two pre-existing failures on Node 22 are from `url.parse()` deprecation, unrelated)
- `node --throw-deprecation` no longer throws for punycode
- IDN/punycode encoding still works correctly (e.g. `münchen.de` -> `xn--mnchen-3ya.de`)

Fixes #1826
Fixes #1797


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

TimShilov:
I'm not sure this is enough.

I have also just encountered this deprecation warning, but it is not coming from `tr64` but from `whatwg-url` itself:
```
[16:14:01]  ERROR  Node warning: The punycode module is deprecated. Please use a userland alternative instead. The punycode module is deprecated. Please use a userland alternative instead.

    at node:punycode:7:10
    at BuiltinModule.compileForInternalLoader (node:internal/bootstrap/realm:398:7)
    at BuiltinModule.compileForPublicLoader (node:internal/bootstrap/realm:337:10)
    at loadBuiltinModule (node:internal/modules/helpers:125:7)
    at loadBuiltinWithHooks (node:internal/modules/cjs/loader:1195:15)
    at Module._load (node:internal/modules/cjs/loader:1285:48)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
    at Module.require (node:internal/modules/cjs/loader:1548:12)
    at require (node:internal/modules/helpers:152:16)
    at Object.<anonymous> (/Users/tim.shilov/p/node_modules/.pnpm/whatwg-url@7.0.0/node_modules/whatwg-url/lib/url-state-machine.js:2:18)
```

I believe that at least version [whatwg-url@9.0.0](https://github.com/jsdom/whatwg-url/releases/tag/v9.0.0) is required, as that's the one where `punycode` usage was actually removed.

However, even `9.0.0` is probably not enough as it still depends on `tr64@2.1.0`, but the `punycode` deprecation was fixed only in [tr46@4.0.0](https://github.com/jsdom/tr46/releases/tag/v4.0.0).

veeceey:
You're right, @TimShilov. Thanks for the detailed report.

The original change bumping to `^7.0.0` was insufficient. I've investigated the whatwg-url release history and found the root cause:

- `whatwg-url` v5.x through v8.x all contain `const punycode = require("punycode")` directly in `lib/url-state-machine.js` / `dist/url-state-machine.js`, calling into Node's deprecated built-in module. None of those versions list `punycode` as an npm dependency, so they rely on Node's built-in.
- **`whatwg-url@9.0.0`** is the first release that fully removes this direct `require("punycode")` call. From v9 onward, all punycode work is delegated to `tr46`, which depends on the userland `punycode` npm package (v2.x) instead.

I've updated the range to `^9.0.0` in the latest commit. Note that `whatwg-url@9.x` requires Node >=12, but that is consistent with where node-fetch's active user base is today.

PR #1866, opened 2026-03-03
docs: add note about Node.js 18+ native fetch

### LINKED ISSUE #1857: Update documentation with a disclaimer about node supporting already fetch
Hello,

First thank you so much for this library.

With node 21 supporting already fetch API (stable) and 18 ( experimental ) should we create a disclaimer in the repository about not needing to install node-fetch for those use cases? 

This would save people from increasing their bundle size unnecessary if they are in a most recent node versions. 

Maybe even a warning if installed in those runtimes could be an interesting feature.

What do you think about this ?

theluckystrike:
Closes #1857


theluckystrike:
Closing in favor of #1867

PR #1867, opened 2026-03-03
docs: add Node 18+ native fetch disclaimer

### LINKED ISSUE #1857: Update documentation with a disclaimer about node supporting already fetch

theluckystrike:
Add a note in the Installation section about Node.js 18.11.0+ having native fetch available. This helps users decide if they need this library for their use case.

Closes #1857
---
*Contributed by [theluckystrike](https://github.com/theluckystrike) | [Zovo](https://zovo.one) — Chrome Extension Studio*


theluckystrike:
With Node 18+ shipping a built-in global `fetch`, new users often install `node-fetch` without realizing they don't need it. This adds a brief note helping users make an informed decision.

theluckystrike:
Closing - repo has not had a commit in 18+ months (last: Aug 2024), blacklisted per Zovo PR policy.

PR #1870, opened 2026-03-04
docs: clarify node-fetch error classes in error handling guide

### LINKED ISSUE #549: Document types of fetch errors more thoroughly

nthbotast:
## Summary
- clarify that operational failures are not always `FetchError`
- document common error classes users may encounter (`AbortError`, `FetchError`, `TypeError`, `RangeError`, `SyntaxError`, `Error`)
- keep guidance on `error.type`, `error.code`, and test coverage reference

Closes #549

## Testing
- not run (documentation-only change)


rsbasic (commented):
**Review: APPROVE (docs only, LOW risk)**

This improves the error handling guide by making the error class hierarchy explicit. The changes are accurate:

- Correctly distinguishes `AbortError`, `FetchError`, `TypeError`, `RangeError`, `SyntaxError` as separate error paths
- Clarifies that `FetchError` with `type = 'system'` wraps Node.js core I/O errors
- Removes ambiguous language ("all operational errors other than aborted requests") and replaces with specific class names
- Updates the test file reference (though the link still points to `master` branch — should be `main`)

The rewrite is clearer and more useful than the original. A developer reading this can now write a proper `catch` block that handles each error type correctly.

**Minor:** The link to `test/main.js` still uses the `master` branch URL. The repo's default branch is `main`. Not a blocker.

**Verdict:** Safe to merge. Strictly improves documentation accuracy.

PR #1875, opened 2026-03-06
fix(types): align FetchError.errno with NodeJS ErrnoException

### LINKED ISSUE #1831: Why is FetchError.errno a string and not a number?

nthbotast:
## Summary
- update `FetchError.errno` type from `string` to `NodeJS.ErrnoException['errno']`
- keep type behavior aligned with Node's `ErrnoException` contract
- adjust the existing type test snippet to accept both number and string errno values

Closes #1831

## Validation
- `git diff --cached --check`
- ⚠️ `npm run test-types` currently fails in this ephemeral environment due to `@types/node/http2.d.ts` parser/type errors (toolchain mismatch), unrelated to this patch


rsbasic (commented):
**Review: APPROVE (types only, LOW risk)**

Changes `FetchError.errno` type from `string` to `NodeJS.ErrnoException['errno']`. This is a type-level correction — no runtime code changes.

The current type (`errno?: string`) is inaccurate. Node.js `ErrnoException.errno` is `number | undefined`, not `string`. The existing type lets TypeScript users write code like `if (err.errno === 'ECONNREFUSED')` which compiles but is incorrect at runtime — `errno` is a numeric system error code, not a string error name. That is what `err.code` is for.

The test file change (`throw new TypeError(error.errno as string | undefined)` → `throw new TypeError(String(error.errno))`) is consistent — it no longer assumes errno is a string.

**One consideration:** This is technically a breaking change for TypeScript consumers who currently compare `errno` to string values. Those comparisons are already wrong at runtime (they would never match), so fixing the type surfaces existing bugs rather than introducing new ones.

**Verdict:** Safe to merge. Corrects a type inaccuracy that masks real bugs.

PR #1878, opened 2026-03-08
fix(request): ignore forbidden request headers

### LINKED ISSUE #1781: Node-fetch allows setting forbidden headers
https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

Results in bugs like #1780

nthbotast:
## Summary
- ignore forbidden request header names from `RequestInit.headers` before dispatch
- keep internal header behavior intact (e.g. computed `content-length`)
- add targeted regression tests for forbidden names and prefixed `proxy-` / `sec-` headers

Fixes #1781

## Validation
- npm test -- --grep "forbidden request headers|internal content-length|allow setting Accept header"


rsbasic (commented):
**Review: REQUEST CHANGES (source code, HIGH risk)**

This PR adds a forbidden headers list per the Fetch spec and silently strips them from outgoing requests. While spec-compliant for browser environments, this is a breaking behavioral change for node-fetch as a server-side HTTP client.

**What it does:**
Adds a hardcoded list of headers that are silently removed before the request is sent: `accept-charset`, `accept-encoding`, `access-control-request-headers`, `access-control-request-method`, `connection`, `content-length`, `cookie`, `date`, `dnt`, `expect`, `host`, `keep-alive`, `origin`, `referer`, `te`, `trailer`, `transfer-encoding`, `upgrade`, `via`, and all `proxy-*` and `sec-*` prefixed headers.

**Why this is problematic for node-fetch specifically:**

1. **`cookie`** — Server-side code intentionally sets cookies for authenticated API calls. Silently stripping them breaks authentication flows with no error message.
2. **`proxy-authorization`** — Required for authenticated proxy environments (common in enterprise networks). Stripping this silently breaks corporate proxy access.
3. **`host`** — Used for virtual host routing. Stripping silently changes which server handles the request.

In a browser, these headers are forbidden because JavaScript should not manipulate them. On the server, developers set them intentionally. node-fetch has always allowed this — changing it silently (no warning, no opt-in) would break production code.

**Additional context from issue #1882:** This PR is from nthbotast, an account created February 27, 2026, that has submitted 160 PRs across multiple HTTP client libraries in 31 days. A similar PR to nodejs/undici (#4860) that would have changed proxy connection behavior was caught and closed by their maintainers.

**Verdict:** Should not be merged as-is. If the goal is spec compliance, this should be opt-in (a `strictMode` option) rather than a silent default change. At minimum, it should emit a warning when a forbidden header is stripped so developers can diagnose the behavioral change.

PR #1902, opened 2026-07-03
docs: fix unbalanced parens in Response example and dead CHANGELOG links

lenoxfernando:
- The `body.blob()` example in the README closes the `new Response(...)` call in the wrong place (`}.blob()`), so the snippet throws a SyntaxError as written.
- Both upgrade guides link to `CHANGELOG.md`, which no longer exists in the repo — pointed them at GitHub Releases instead.

Found while running [docrot](https://github.com/getdocrot/docrot), a small checker that verifies docs examples and links against the code. Happy to adjust anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -811,7 +811,7 @@ http.createServer(async function (req, res) {
   const arrayBuffer = await new Response(req).arrayBuffer()
   const blob = await new Response(req, {
     headers: req.headers // So that `type` inherits `Content-Type`
-  }.blob()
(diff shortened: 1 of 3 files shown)


lbesecker195 (approved):
Verified: `CHANGELOG.md` no longer exists in the repo, so both upgrade-guide links were dead, and the README snippet did close `new Response(...)` in the wrong place. LGTM.

PR #1908, opened 2026-07-28
fix: validate Response status range and null-body status per fetch spec

marceli1404:
# Description

Per the [Fetch spec](https://fetch.spec.whatwg.org/#response-class), the Response constructor should validate:

1. **Status range**: Throw RangeError if status is not in 200-599
2. **Null body status**: Throw TypeError if body is provided with a null-body status (204, 304)
3. **Status text**: Validate for invalid characters

### `src/response.js`
- Added `NULL_BODY_STATUSES` set ([101, 204, 304])
- Status range validation throws `RangeError` for status outside [200, 599]
- Status text validation rejects CR, LF, and NUL characters
- Null body status check throws `TypeError` when body is non-null with status 204/304/101
- `Response.error()` now uses `Reflect.construct` to bypass constructor validation since it uses status 0

### `src/index.js`
- Internal HTTP response handling for 204/304 now passes `null` body instead of the response stream, since these statuses must not have a body per spec

### `test/response.js`
- 5 new test cases for status range and null body validation

## Verification
- All 392 tests pass (5 new), only 3 pre-existing unrelated failures remain
- All existing 204/304 response handling continues to work correctly
- Git commit signed and verified


marceli1404:
Extending this PR: added **205 Reset Content** (and 103) to the null-body statuses, per the fetch spec (a null body status is 101, 103, 204, 205, or 304).

- \NULL_BODY_STATUSES\ now \[101, 103, 204, 205, 304]\ in src/response.js
- fetch-time body nulling now also handles 205 in src/index.js
- new regression tests: \should handle reset content response with no body\ (main.js) + 205 cases in the constructor tests (response.js)

All suites pass (only the 3 pre-existing Node-24 environmental failures remain: 2 chunked-timing flakes + the deprecated agent 'family' option).

PR #1852, opened 2025-07-25
fix crash on abort when body is not a stream

sbingner:
## Purpose
This causes node-fetch to stop crashing node if you abort a request

## Additional information

___

- [ ] I updated readme
- [ ] I added unit test(s)

___

- fix #1801

Issue #1801, opened 2024-01-23
Unhandled 'error' event when using AbortSignal to cancel requests


AleCaste:
**Reproduction**

```javascript
const controller = new AbortController();

const fetchTimeout = setTimeout(() => {
    controller.abort();
}, 500);   // Try different values from 0 - 3000

const requestOptions = {
    method: 'POST',
    headers: {
      'content-type': 'application/json'
    },
    body: JSON.stringify({test: 'test'}),
    signal: controller.signal
};

try {
  const response = await fetch(url, requestOptions)
} catch (err) {
  // handle error
}
```
The problem is somehow intermittent. Try different timeout values until you can reproduce it.
I get the following error:
```
node:events:368
      throw er; // Unhandled 'error' event
      ^

AbortError: The operation was aborted.
    at abort (file:///D:/xampp/htdocs/fetch_retry_test/node_modules/node-fetch/src/index.js:70:18)
    at file:///D:/xampp/htdocs/fetch_retry_test/node_modules/node-fetch/src/index.js:87:4
    at new Promise (<anonymous>)
    at fetch (file:///D:/xampp/htdocs/fetch_retry_test/node_modules/node-fetch/src/index.js:49:9)
    at wrappedFetch (D:\xampp\htdocs\fetch_retry_test\node_modules\fetch-retry\index.js:71:9)
    at Timeout._onTimeout (D:\xampp\htdocs\fetch_retry_test\node_modules\fetch-retry\index.js:127:11)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
Emitted 'error' event on Readable instance at:
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  type: 'aborted'
}
```

**Expected behavior**
The error happens in index.js in the following code block:
```javascript
const abort = () => {
  const error = new AbortError('The operation was aborted.');
  reject(error);
  if (request.body && request.body instanceof Stream.Readable) {
    request.body.destroy(error);   // <==== HERE IS WHERE THE UNHANDLED ERROR OCCURS!!!!
  }

  if (!response || !response.body) {
    return;
  }

  response.body.emit('error', error);
};
```
To fix it, you need a listener for the "error" event of the Readable stream, like this below:
```javascript
const abort = () => {
[...]

PR #1911, opened 2026-09-01
fix: prevent crash when aborting a request with a non-stream body

marceli1404:
### Bug
Calling `fetch()` with a **non-stream body** (string, Buffer, Blob, URLSearchParams, etc.) and an **already-aborted** `AbortController` **crashes the process**:

```js
const ac = new AbortController();
ac.abort();
fetch(url, { method: 'POST', body: 'hello world', signal: ac.signal });
// Unhandled 'error' event -> process terminates with exit code 1
```

The fetch promise rejects cleanly with `AbortError`, but the process still dies because the internally-created `Stream.Readable.from(body)` has no `'error'` listener.

### Root cause
In the `Body` constructor, the `'error'` handler is attached only when the **original `body`** is a `Stream`:

```js
if (body instanceof Stream) {
    body.on('error', ...);
}
```

For non-stream bodies, the actual stream used at runtime is `Stream.Readable.from(...)`, which never gets a listener. When `abort()` calls `request.body.destroy(error)` on that un-listened stream, the error event is unhandled.

### Fix
Attach the handler to **`stream`** (the value actually used) instead of `body`:

```js
if (stream instanceof Stream) {
    stream.on('error', ...);
}
```

For stream bodies `stream === body`, so existing behavior is unchanged; for non-stream bodies the internally-created stream now gets the same handler.

### Verification
- New regression test fails on unpatched code (crash), passes with the fix.
- Full `AbortController` describe block: 32 passing.
- Targeted test passes across all three AbortController variants (native, mysticatea, polyfill).


marceli1404:
Note on overlap: this addresses the same root cause as the earlier open PR #1852 (\ix crash on abort when body is not a stream\) and issue #1801.

- #1852 adds a parallel error listener only when stream !== body; this PR instead changes the existing condition from ody instanceof Stream to stream instanceof Stream (single listener, same behavioral outcome) **and** includes a regression test that fails on unpatched code.
- Whatever the maintainers prefer is fine with me: I'm happy to close this PR and port the regression test onto #1852, or keep this one as the canonical fix if the cleaner single-listener version is preferred.

This bug otherwise needs no further coordination on my side — both approaches fix the same crash.

PR #1914, opened 2026-09-20
fix: include AggregateError causes in request failure reason

### LINKED ISSUE #1841: Failing without clear reason if https module throwed AggregateError
**Reproduction**

Sorry there is no exact reproduction steps since this is network related but if both ipv4 and ipv6 are supported for the domain and both method failed, https module will throw AggregateError.

https://github.com/nodejs/node/blob/v22.11.0/lib/net.js#L1121

node-fetch catches error with 'error' event and extracts 'message' property of error, but the `AggregateError` have empty error message so no usable error information is provided.

https://github.com/node-fetch/node-fetch/blob/8b3320d2a7c07bce4afc6b2bf6c3bbddda85b01f/src/index.js#L108

In my case, `fetch('https://floss.socual')` is what we do and we see error mesage like `FetchError: request to https://floss.social/ failed, reason: `

**Expected behavior**

We should able to see all or some of error messages from `AggregateError` like `FetchError: request to https://floss.social/failed, reason: ETIMEOUT and ENETUNREACH`

**Screenshots**

Image

**Your Environment**

| software         | version
| ---------------- | -------
| node-fetch       | 3.3.2
| node             | 22.11.0
| npm              | not used
| pnpm             | 9.6.0
| Operating System | Linux 6.8.0 (according to uname -a) debiaon 12.8 (according to /etc/debian_version)

Actually it's `node:22.11.0-slim` docker container

**Additional context**

N/A

Possibly related: #1786,

dyk1454683243-sudo:
## Purpose

When Node's https/net stack fails both IPv4 and IPv6 connect attempts, it emits an `AggregateError` whose `message` is empty. node-fetch only interpolated `error.message`, so users saw:

```
FetchError: request to https://example.com/ failed, reason:
```

Fixes #1841

## Changes

- When formatting the request-failure reason, if `error.message` is empty, join nested `error.errors` codes/messages with ` and `.
- Single `Error` instances with a non-empty message are unchanged.
- Added a unit test that injects an `AggregateError` (or empty-message multi-error on Node < 15) on the request error path via a custom `http.Agent`.

## Additional information

Dual-stack connect failures are the usual source of this `AggregateError` (see Node `net.js`).

- [ ] I updated readme
- [x] I added unit test(s)

___

- Fixes #1841


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

